PR #4, opened 2022-04-07
Patch push to support export to webdav remotes

### LINKED ISSUE #3: TODO list WebDAV support
This issue should contain an overview of all issues that should be addressed/fixed before end of April 2022, in order to seamlessly integrate `clone` and `push` to WebDAV siblings, which might also require a `create-sibling-webav`.

### Outline of the current implementation strategy:

1. For each webdav-sibling with name `n`, create two remotes: `n-tree`, and `n-git`. The first is a git-annex special remote of type webdav with exporttree set to yes. The second is a `datalad-annex::`-special remote that points to the same location as `n-tree`.
2. Patch (via datalad-next) the push command to redirect pushes to a special remote with export-tree to `git annex export`, as described here: https://github.com/datalad/datalad/issues/3127. The export implementation could for example be done via `git annex export` or via something similar to `x-export-to-webdav`.
3. Define a publish dependency between both of them. Then pushing, for example, to `n-git` could autmatically trigger pushing to `n-tree`, which leads to an automated push of `n-tree`.
4. Integrate them with the new credential system.

### Work packages:

- [x] datalad contains a refactoring of `push` that isolates the actual data transport into `_transfer_data` (PR https://github.com/datalad/datalad/pull/6618).
- [ ] datalad_next contains a patch in `push_to_export_remote.py` that patches `_transfer_data`. There are currently two alternative patches in two branches and two PRs:
    -- Alternative1: PR #4: use x-export-to-webdav to implement the equivalent of `_push_data` for exporttree-remotes. This is most likely *TOO* specific and contains probably *TOO* much logic (subdatasets). But the existing code can probably be re-used in parts.
   -- Alternative2: PR #5: use plain `git annex export` to push `HEAD` to the remote. The current code does currently not perform all checks that `_push_data` implements. There might be no need for the checks, or there might be need for more checks and logic, for example w.r.t. subdatatasets. That depends on the position of `_push_data` in the `datalad push` logic, which I am not familiar with.

- [x] Ensure that the new credential system is used.
- [x] Add a `siblings-create-webdav` that will take a name-base and create two siblings: &lt;name-base&gt;`-tree`, which is a webdav special remote with `export-tree` set to `yes`, and &lt;name-base&gt;`-vcs`, which is a `datalad-annex::`-remote which points to the same location as the &lt;name-base&gt;`-tree` remote. This is done in PR #6 

### Authentication and credentials:

- https://github.com/mih/datalad-mihextras/issues/30: round up of issues that need to be fixed for VCS+Storage backends

- https://github.com/mih/datalad-mihextras/issues/28: check this out for credential handling in `x-export-to-webdav`

- https://github.com/datalad/datalad/issues/6451: describes an authentication scenario for webdav

- https://github.com/datalad/datalad/issues/5483: describes support for B2DROP which is webdav-based

christian-monch:
This PR adds a patch that will patch `push` in `datalad.core.distributed.push`, so that it supports pushing to a git annex special remote that has exporttree == yes.

Fixes #3 

- [x] Add tests for `push_to_export_remote.py`
- [x] Handle credentials
- [x] Verify export target validity
- [x] Add tests for target validation code


datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,106 @@
...
+                   force: Optional[str],
+                   jobs: Optional[Union[str, int]],
+                   res_kwargs: Dict,
+                   got_path_arg: bool
+                   ) -> Generator:
+
+    if _is_export_remote(repo, target):
+        # TODO:
+        #  - check for configuration entries, e.g. what to export
+        #  - check for all kind of things that are checked in push._push_data
```
bpoldrack, line 71: In my current estimation that's not needed - I'll push a suggestion to an additional RF in `push` to have the relevant checks done outside of `_push_data`.

mslw:
Relaying @mih's comment from group chat so it doesn't get lost - probably we should make datalad-next depend on already released DataLad 0.16 even if it means more complex patching (ie. relying on push as-is, not on its refactored version) and if the approach here works well, only then introduce changes to core. Which would mean patching push or push_data. Originally proposed way: https://github.com/datalad/datalad/issues/3127#issuecomment-962871443

mih:
The dependency on 0.16 is also done in https://github.com/datalad/datalad-next/commit/071c8c61485b31c6577bcdc842ffb97a06cd981e

mih (commented):
Not sure how long ago I wrote these comments. Posting them now. Were stuck pending. Sorry.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,106 @@
...
+        if ds.config.getbool('remote.{}'.format(target), 'annex-ignore', False):
+            lgr.debug(
+                "Target '%s' is set to annex-ignore, exclude from data-export.",
+                target,
+            )
+            return
+
+        res_kwargs['target'] = target
+
+        repo.call_git(["annex", "export", "HEAD", "--to", target])
```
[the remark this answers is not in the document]
christian-monch: I added credential handling in https://github.com/datalad/datalad-next/pull/4/commits/29038a8c74e30c72442874d44610a3181b3253da

datalad_next/patches/__init__.py
```diff
@@ -1,5 +1,6 @@
 from . import (
     annexrepo,
     create_sibling_ghlike,
+    push_to_export_remote,
```
mih, line 4: FTR: Import time increment seems to be 10ms. Not bad.

In comparison, the `create_sibling_ghlike` patch is 40ms, `annexrepo` is 95ms, `siblings` is 4ms.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,145 @@
...
+    get_specialremote_credential_envpatch,
+    get_specialremote_credential_properties,
+    needs_specialremote_credential_envpatch,
+)
+
+
+__docformat__ = "restructuredtext"
+
+
+lgr = logging.getLogger('datalad_next.push_to_export_remote')
```
mih, line 27: I think, it is better to use the same logger as push (i.e. same name).
```suggestion
lgr = logging.getLogger('datalad.core.distributed.push')
```

at least that is what is done elsewhere.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _is_export_remote(remote_info: Optional[Dict]) -> bool:
+    """Check if remote_info is valid and has exporttree set to "yes"
+
+    :param remote_info: None or a dictionary the contains git annex special
+    remote info
+    :return: True if exporttree is yes, else False
```
mih, line 35: Here and elsewhere, we should stick to the numpydoc conventions

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,145 @@
...
+    :param target: name of the remote that should be checked
+    :return: True if git annex reports exporttree is yes, else False
+    """
+    remote_info = [
+        record
+        for record in repo.get_special_remotes().values()
+        if record.get("name") == target]
+    if len(remote_info) == 1:
+        return remote_info[0].get("exporttree") == "yes"
+    return False
```
mih, line 57: Function is not used, it seems.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,145 @@
...
+                "Target '%s' is set to annex-ignore, exclude from data-export.",
+                target)
+            return
+
+        # Check for credentials
+        sr_params = {
+            "type": remote_info.get("type"),
+            "url": remote_info.get("url")
+        }
+        credential_name, credentials = None, None
```
mih, line 95: Technically, we have no use for `credential_name`

mih: Unless we also do the update-on-success pattern

mih (commented):
Went over the code now. Will do some testing next.

datalad_next/patches/push_to_export_remote.py
christian-monch: ok, will change it

datalad_next/patches/push_to_export_remote.py
christian-monch: Thx, removed

datalad_next/patches/push_to_export_remote.py
christian-monch: will remove it. I think updates should be left to create-sibling-webdav

datalad_next/patches/tests/test_push_to_export_remote.py
```diff
@@ -0,0 +1,81 @@
...
+def test_patch_pass_through():
+    # Ensure that the original _transfer_data is called if the target remote
+    # has exporttree # not set to "yes"
+    from datalad.core.distributed.push import _transfer_data as _original_transfer_data
+    from datalad_next.patches.push_to_export_remote import _transfer_data
+
+    print(_original_transfer_data, _transfer_data)
+    _transfer_data(*([None]*9))
+    with patch("datalad_next.patches.push_to_export_remote.push._push_data") as pd_mock:
+        _original_transfer_data(
```
bpoldrack, line 53: @christian-monch
Took me a while to figure, but this test does fail b/c this call only creates the generator. It doesn't execute anything.

christian-monch: Thx. (face-palm)

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+        if credentials and needs_specialremote_credential_envpatch(remote_type):
+            env_patch = get_specialremote_credential_envpatch(
+                remote_type,
+                credentials)
+
+        res_kwargs['target'] = target
+
+        with patch.dict('os.environ', env_patch):
+            try:
+                repo.call_git(["annex", "export", "HEAD", "--to", target])
```
mih, line 199: It would be ideal to call it with `--json --json-error-messages`. This will produce reports like

```
{"command":"export demo","success":true,"input":[],"error-messages":[],"file":".datalad/.gitattributes"}
{"command":"export demo","success":true,"input":[],"error-messages":[],"file":".datalad/config"}
{"command":"export demo","success":true,"input":[],"error-messages":[],"file":".gitattributes"}
```

they could be feed into `annexjson2result()` and give us copyto-like result reports.

christian-monch: Thx, will do that

mih: It could even to --json-progress and we could use `_call_annex_records_items_()`

mih (commented):
I tested this out by hand (non-recursive), and it works. Also a subsequent push that removes content works.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+        #  create_sibling_webdav.py
+        credential_manager = CredentialManager(ds.config)
+        credentials = (credential_manager.query(
+            _sortby='last-used',
+            **credential_properties) or [(None, None)])[0][1]
+    return credentials
+
+
+# TODO: create a function similar to AnnexRepo.get_special_remote() and put it
+#  into core.
```
mih, line 70: Given that we already carry an `annexrepo` patch here, we can just actually patch it in, and register it in https://github.com/datalad/datalad/issues/6642

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+        with patch.dict('os.environ', env_patch):
+            try:
+                repo.call_git(["annex", "export", "HEAD", "--to", target])
+            except CommandError as cmd_error:
+                ce = CapturedException(cmd_error)
+                yield {
+                    **res_kwargs,
+                    "action": "copy",
+                    "status": "error",
+                    "message": str(ce),
```
mih, line 206: I forgot the details, but do we need `str()` here @bpoldrack ?

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+                   jobs: Optional[Union[str, int]],
+                   res_kwargs: Dict,
+                   got_path_arg: bool
+                   ) -> Generator:
+
+    target_uuid, remote_info = ([
+        (uuid, info) for uuid, info in repo.get_special_remotes().items()
+        if info.get("name") == target] or [(None, None)])[0]
+
+    if _is_export_remote(remote_info):
```
mih, line 151: Personal style note: In such cases (almost the entire function body is a conditional), i find it to be cognitively less complex to reverse the conditional, and to the minimal `else` first, followed by an explicit return. Like so:

```suggestion
    if not _is_export_remote(remote_info):
            yield from push._push_data(
            ds,
            target,
            content,
            data,
            force,
            jobs,
            res_kwargs.copy(),
            got_path_arg=got_path_arg,
        )
        return
```

and then take the main conditional and deindent it. Essentially saying "we don't handle this one thing here", rather than saying "here is a list of all things we handle here". Otherwise there are 60 lines to scroll to remember what was the other thing.

Or put the conditional clause in a function, whatever makes the main decision immediately obvious when going through the function from start to end.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+push._transfer_data = _transfer_data
+push.Push._params_["force"] = Parameter(
+    args=("-f", "--force",),
+    doc="""force particular operations, possibly overruling safety
+    protections or optimizations: use --force with git-push ('gitpush');
+    do not use --fast with git-annex copy ('checkdatapresent'); force an
+    annex export (to git annex remotes with "exporttree" set to "yes");
+    combine all force modes ('all').""",
+    constraints=EnsureChoice(
+        'all', 'gitpush', 'checkdatapresent', 'export', None))
```
mih, line 246: I'd say that we also need to amend the main `Push.__doc__` and announce the additional capabilities. Maybe just append

```
Features added by the datalad-next extension

Transparently transfer data to git-annex special remotes in export-mode (`exporttree=yes`), whenever ...
```

datalad_next/patches/push_to_export_remote.py
christian-monch: Adressed as suggested in commit 14b1d978f6eaa32d90cd2fe0131f04d1d101d5f9.

The function `_transfer_data` now uses `_call_annex_record_items_()` and yields all results immediately.

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,240 @@
...
+            data,
+            force,
+            jobs,
+            res_kwargs.copy(),
+            got_path_arg=got_path_arg,
+        )
+
+
+lgr.debug("Patching datalad.core.distributed.push._transfer_data")
+push._transfer_data = _transfer_data
```
mih, line 237: I'd say that we also need to amend the main `Push.__doc__` and announce the additional capabilities. Maybe just append

```suggestion
push._transfer_data = _transfer_data
push.Push.__doc__ += """\


Features added by the datalad-next extension

Transparently transfer data to git-annex special remotes in export-mode (`exporttree=yes`), whenever ...
"""
```

datalad_next/patches/push_to_export_remote.py
mih: FTR: git-annex does not seem to distinguish a file that is already up-to-date, and one that actually needs copying. The only way to discover that seems to be to look for `unexport` messages:

Here only `myfile.txt` has changes since the last export

```json
{"command":"unexport fz-juelich.sciebo.de-storage","success":true,"input":[],"error-messages":[],"file":"myfile.txt"}
{"byte-progress":125,"action":{"command":"export fz-juelich.sciebo.de-storage","input":[],"file":".datalad/.gitattributes"},"total-size":125,"percent-progress":"100%"}
{"command":"export fz-juelich.sciebo.de-storage","success":true,"input":[],"error-messages":[],"file":".datalad/.gitattributes"}
{"byte-progress":63,"action":{"command":"export fz-juelich.sciebo.de-storage","input":[],"file":".datalad/config"},"total-size":63,"percent-progress":"100%"}
{"command":"export fz-juelich.sciebo.de-storage","success":true,"input":[],"error-messages":[],"file":".datalad/config"}
{"command":"export fz-juelich.sciebo.de-storage","success":true,"input":[],"error-messages":[],"file":".gitattributes"}
{"byte-progress":5,"action":{"command":"export fz-juelich.sciebo.de-storage","input":[],"file":"myfile.txt"},"total-size":5,"percent-progress":"100%"}
{"command":"export fz-juelich.sciebo.de-storage","success":true,"input":[],"error-messages":[],"file":"myfile.txt"}

```

datalad_next/patches/push_to_export_remote.py
christian-monch: Addressed as suggested

datalad_next/patches/push_to_export_remote.py
christian-monch: done

datalad_next/patches/push_to_export_remote.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+lgr.debug("Patching datalad.core.distributed.push.Push docstring and parameters")
+push.Push.__doc__ += """\
+
+
+Features added by the datalad-next extension
+
+Transparently transfer data to git-annex special remotes in export-mode (`exporttree=yes`), whenever ...
```
mih, line 250: This is TODO (for later) https://github.com/datalad/datalad-next/issues/42

mih:
And merging! Thx @christian-monch

PR #5, opened 2022-04-08
WIP: POC: ALT-2: Support exporttree git annex remotes

christian-monch:
This PR is an alternative to PR #4 
This PR partially solves issue #3, it does not touch authentication

This PR adds a patch that will patch push in datalad.core.distributed.push, so that it supports "pushing" to a git annex special remote that has exporttree == yes. Instead of using `git push` this code will use `git annex export HEAD --to <remote>`.

- [ ] ensure that all necessary checks are done before and after calling `git annex export ...`


mih:
I am closing this PR. We have settled on https://github.com/datalad/datalad-next/pull/4.

PR #6, opened 2022-04-08
Add a first version of create-sibling-webdav

christian-monch:
This PR introduces the command `create-sibling-webdav`, which creates a set of two remotes, one for git annex export and one for datalad push.

- One is called <name>-vcs and is meant to contain the git-repository info.
- The other one is called <name>-tree and is meant to contain the annex-exported tree.

Credential manager is still "datalad.downloaders.credentials.UserPassword".

The siblings are meant to be used with a publish dependency

### DIFF
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -204,6 +204,7 @@ class CreateSiblingWebDAV(Interface):
             if failed:
                 return
 
+        return
         # TODO the rest should be wrapped into a helper function and be executed with
         # foreach-dataset
 
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -1,0 +1,157 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 et:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""High-level interface for creating a combi-target on a WebDAV capable server
+ """
+import logging
+import os
+from typing import (
+    Optional,
+    Union,
+)
+from urllib.parse import urlparse
+
+from datalad.distribution.dataset import (
+    Dataset,
+    EnsureDataset,
+    datasetmethod,
+    require_dataset,
+)
+from datalad.downloaders.credentials import UserPassword
+from datalad.interface.base import Interface
+from datalad.interface.utils import eval_results
+from datalad.support.param import Parameter
+from datalad.support.annexrepo import AnnexRepo
+from datalad.support.constraints import (
+    EnsureNone,
+    EnsureStr,
+)
+
+
+__docformat__ = "restructuredtext"
+
+lgr = logging.getLogger('datalad_next.create_sibling_webdav')
+
+
+class CreateSiblingWebDAV(Interface):
+    """
+
+    """
+    _examples_ = [
+    ]
+
+    _params_ = dict(
+        url=Parameter(
+            args=("url",),
+            metavar="https://<host>[:<port>]/<local part>",
+            doc="URL identifying the target WebDAV server",
+            constraints=EnsureStr()),
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to process.  If
+            no dataset is given, an attempt is made to identify the dataset
+            based on the current working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        name=Parameter(
+            args=('-s', '--name',),
+            metavar='NAME',
+            doc="Name of the sibling.",
+            constraints=EnsureStr() | EnsureNone(),
+            required=True),
+        credential=Parameter(
+            args=("--credential",),
+            doc="""Name of the credentials that should be used to access
+            the WebDAV server.""",
+        )
+    )
+
+    @staticmethod
+    @datasetmethod(name='create_sibling_webdav')
+    @eval_results
+    def __call__(
+            url: str,
+            *,
+            dataset: Optional[Union[str, Dataset]] = None,
+            name: Optional[str] = None,
+            credential: Optional[str] = None):
+        """
+
+        :param url:
+        :param dataset:
+        :param name:
+        :param credential:
+        :return: a generator yielding result records
+        """
+
+        ds = require_dataset(
+            dataset or ".",
+            check_installed=True,
+            purpose=f'create WebDAV sibling(s)')
+
+        res_kwargs = dict(
+            action=f'create_sibling_webdav',
+            logger=lgr,
+            refds=ds.path,
+        )
+
+        if not isinstance(ds.repo, AnnexRepo):
+            yield {
+                **res_kwargs,
+                "status": "error",
+                "msg": "require annex repo to create WebDAV sibling"
+            }
+            return
+
+        parser_url = urlparse(url)
+        if parser_url.query:
+            yield {
+                **res_kwargs,
+                "status": "error",
+                "msg": "URLs with query are not yet supported"
+            }
+            return
+
+        name_base = name or parser_url.hostname
+        git_name = name_base + "-wd-vcs"
+        annex_name = name_base + "-wd-tree"
+
+        datalad_annex_url = (
+            "datalad_annex::"
+            + url
+            + "?type=webdav&url={noquery}&encryption=none&exporttree=yes"
+            + "" if credential is None else f"&dlacredential={credential}"
+        )
+
+        # Add the datalad_annex:: sibling to datalad
+        from datalad.api import siblings
+
+        siblings(
+            action="add",
+            dataset=ds,
+            name=git_name,
+            url=datalad_annex_url)
+
+        # Add a git-annex webdav special remote. This might requires to set
+        # the webdav environment variables accordingly.
+
+        credential_holder = UserPassword(name=credential)()
+        os.environ["WEBDAV_USERNAME"] = credential_holder["user"]
+        os.environ["WEBDAV_PASSWORD"] = credential_holder["password"]
+        ds.repo.call_annex([
+            "initremote",
+            annex_name,
+            "type=webdav",
+            f"url={url}",
+            "exporttree=yes",
+            "encryption=none"
+        ])
+
+        yield {
+            **res_kwargs,
+            "status": "ok"
+        }
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -24,10 +24,15 @@ from datalad.distribution.dataset import (
 )
 from datalad.downloaders.credentials import UserPassword
 from datalad.interface.base import Interface
+from datalad.interface.common_opts import (
+    recursion_flag,
+    recursion_limit
+)
 from datalad.interface.utils import eval_results
 from datalad.support.param import Parameter
 from datalad.support.annexrepo import AnnexRepo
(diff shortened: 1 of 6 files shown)


mih:
Cool! Thx!

Re:

> * One is called -vcs and is meant to contain the git-repository info.
> * The other one is called -tree and is meant to contain the annex-exported tree.

I would prefer to continue the pattern established by the RIA combo: no suffix for the git-repo part, and `-storage` for the annex. Will take a closer look

christian-monch:
> Cool! Thx!
> 
> Re:
> 
> > * One is called -vcs and is meant to contain the git-repository info.
> > * The other one is called -tree and is meant to contain the annex-exported tree.
> 
> I would prefer to continue the pattern established by the RIA combo: no suffix for the git-repo part, and `-storage` for the annex. Will take a closer look

Good idea, wasn’t aware of that convention

mslw:
> > Cool! Thx!
> > Re:
> > > * One is called -vcs and is meant to contain the git-repository info.
> > > * The other one is called -tree and is meant to contain the annex-exported tree.
> > 
> > 
> > I would prefer to continue the pattern established by the RIA combo: no suffix for the git-repo part, and `-storage` for the annex. Will take a closer look
> 
> Good idea, wasn’t aware of that convention. Will do that when I am back, until then it’s free for grabs

I also wasn't aware of the convention, but the same was pointed out by @bpoldrack 

I'll also take a look and hopefully manage to update the arguments to match `create-sibling-ria`.

I think, following our initial ideas, ideally there would be flags to choose between creating:
- git-repo & storage (default)
- git-repo only (`--no-storage-sibling` in ria) 
- storage only (not available for ria, but in this context still useful as a wrapper for `git annex initremote` to be used with a git-repo sibling elsewhere)

mih:
Need to think this through a bit more. Few aspects that need to be considered:
- there is an an export-mode possibility, in contrast to `create-sibling-ria`
- `datalad-annex::` URLs are only possible with https://github.com/mih/datalad-mihextras/releases/tag/0.6.0, which would need to become a dependency
- webdav siblings in plain export mode with the VCS part are already pretty useful, so `mihextras` might be considered optional at this point
- we would need an `--existing` mode switch
- we need `--recursive` -- and with that we need to consider how to support non-export sibling nesting
- `--credential` should be implemented analog to the patch for `create-sibling-ghlike` in this extension

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+            refds=ds.path,
+        )
+
+        if not isinstance(ds.repo, AnnexRepo):
+            yield {
+                **res_kwargs,
+                "status": "error",
+                "msg": "require annex repo to create WebDAV sibling"
+            }
+            return
```
mih, line 108: I believe that a `datalad-annex::`-type sibling would be meaningful even for a non-annex dataset.

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+
+        parser_url = urlparse(url)
+        if parser_url.query:
```
mih, line 111: This would also need to be conditional on the desire to have a git-sibling.

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+
+        name_base = name or parser_url.hostname
+        git_name = name_base + "-wd-vcs"
+        annex_name = name_base + "-wd-tree"
+
+        datalad_annex_url = (
+            "datalad_annex::"
+            + url
+            + "?type=webdav&url={noquery}&encryption=none&exporttree=yes"
+            + "" if credential is None else f"&dlacredential={credential}"
```
mih, line 127: This raises an interesting case. It is technically not necessary to do `datalad-annex:: + URL + ?type=webdav&url={noquery}&....` it could just as well be `datalad-annex::?type=webdav&url={url}...`

But I do not think that the `datalad-annex` implementation actually tests URLs with a query component. I strongly suspect that it is possible with the right kind of quoting, but a test plus the resulting documentation is needed.

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+        # Add the datalad_annex:: sibling to datalad
+        from datalad.api import siblings
+
+        siblings(
```
mih, line 133: should `yield from siblings(...)`

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+            dataset=ds,
+            name=git_name,
+            url=datalad_annex_url)
+
+        # Add a git-annex webdav special remote. This might requires to set
+        # the webdav environment variables accordingly.
+
+        credential_holder = UserPassword(name=credential)()
+        os.environ["WEBDAV_USERNAME"] = credential_holder["user"]
+        os.environ["WEBDAV_PASSWORD"] = credential_holder["password"]
```
mih, line 144: we better use `with patch.dict('os.environ', ...):` to not leave the credentials around in long running python sessions.

mih (commented):
Few notes, mostly to self.

datalad_next/create_sibling_webdav.py
mih: I also think it must be `datalad-annex::` not `datalad_annex::`

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,157 @@
...
+
+        :param url:
+        :param dataset:
+        :param name:
+        :param credential:
+        :return: a generator yielding result records
+        """
+
+        ds = require_dataset(
+            dataset or ".",
```
mih, line 92: FTR: I believe this should be just `dataset`. The only deviation from this pattern that I could find is

```
datalad/metadata/extract_metadata.py:        dataset = require_dataset(dataset or curdir,
```

Passing `curdir` forces a `Dataset(curdir)`, whereas passing `None` will result in a `Dataset(get_dataset_root(getpwd()))`, which is the more standard, and more desirable behavior, I think, because it enables working within subdirectories of datasets seamlessly.

mih:
FTR: I looked into possibilities for implementing unit tests without having to rely on an external service. So far, https://pypi.org/project/WsgiDAV/ seems to be the most promising utility. I failed to get meaningful behavior with https://github.com/andrewleech/PyWebDAV3

bpoldrack (commented):
I just noticed, that my comments where "pending". Submitting only now. Sorry.

However - they are largely covered by @mih comments as well.

datalad_next/create_sibling_webdav.py
bpoldrack: Copying over here from my review. Same concern, different suggestion (although mine would be for later, since it's a change in core)

This is:
1. "Leaking" the credentials into the user's environment.
2. Possibly overwriting existing user settings for use with another webdav special remote.

I think, we would want to allow `AnnexRepo.call_annex` to take an `env` parameter and pass it down instead, patching `env` for the resulting subprocess only.

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: I think this is the wrong place for this comment. Adding support for this might be good, but it should be done in core, and can only later be adopted here. It is also not totally straightforward. The current core runner already supports ENV manipulation, but in doing so already performs several copies of entire environment. "Patching" the environment is something we need to do rather often, but passing entire environment (what we support) is not very useful, in general. What you are asking for will likely result in user code "patching" the environment, in order to pass a full environment to `call_annex`, which passes it to the runner. If that is the case, there is no gain.

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,437 @@
...
+from datalad.support.exceptions import CapturedException
+from datalad_next.credman import CredentialManager
+
+
+__docformat__ = "restructuredtext"
+
+lgr = logging.getLogger('datalad_next.create_sibling_webdav')
+
+
+class CreateSiblingWebDAV(Interface):
```
mih, line 48: Must use `@build_doc`

mih: done

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: done

mih:
There are still many TODOs and no tests, nor docs, but I believe this current state is a workable draft that provides meaningful functionality.

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,437 @@
...
+    @staticmethod
+    @datasetmethod(name='create_sibling_webdav')
+    @eval_results
+    def __call__(
+            url: str,
+            *,
+            dataset: Optional[Union[str, Dataset]] = None,
+            name: Optional[str] = None,
+            storage_name: Optional[str] = None,
+            storage_sibling: Optional[str] = 'yes',
```
christian-monch, line 145: `Optional` should be removed. It is not correctly reflecting the type of `storage_sibling`, since `storage_sibling` is always a `str`.

```suggestion
            storage_sibling: str = 'yes',
```

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,437 @@
...
+            credential: Optional[str] = None,
+            existing: Optional[str] = None,
+            recursive: Optional[bool] = False,
```
christian-monch, line 148: `Optional` should be removed. It is not correctly reflecting the type of `recursive`, since `recursive` is always a `bool`.

```suggestion
            recursive: bool = False,
```

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,547 @@
...
+    export: bool
+    """
+    remote_url = "datalad-annex::?type=webdav&encryption=none&" \
+                 "exporttree={export}&dlacredential={cred}&url={url}".format(
+        export='yes' if export else 'no',
+        cred=credential_name,
+        # TODO urlquote, because it goes into the query part of another URL
+        url=url,
+    )
+    print("MAKEREPO", remote_url)
```
christian-monch, line 420: Probably a left over from testing.
```suggestion
```

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,547 @@
...
+    Interface,
+    build_doc,
+)
+from datalad.interface.common_opts import (
+    recursion_flag,
+    recursion_limit
+)
+from datalad.interface.results import get_status_dict
+from datalad.interface.utils import eval_results
+from datalad.support.param import Parameter
```
christian-monch, line 37: `log_progress()` is used below, but never imported.

```suggestion
from datalad.log import log_progress
from datalad.support.param import Parameter
```

datalad_next/create_sibling_webdav.py
```diff
@@ -0,0 +1,547 @@
...
+        "encryption=none"
+        # TODO consider
+        #chunk/chunksize
+        # TODO embedding credentials would simplify a non-datalad
+        # push/copy, but would also scatter duplicates of credentials
+        # around that need maintenance when expiring/changing
+        #embedcreds
+        # TODO autoenable?
+    ]
+    print("MAKESTORE", cmd_args)
```
christian-monch, line 473: Probably left from testing:
```suggestion
```

christian-monch (commented):
There is one missing import and a few minor things, other than that it LGTM.

mih:
Thx @christian-monch for the suggestions -- merged!

The print statements, plus the various TODOs still need to be addressed -- quite a long way to go still.

PR #10, opened 2022-04-19
New serve_path_via_webdav test helper

mih:
It deploys a local webdav server using wsgidav and cheroot, which become
dev-dependencies. Moreover, the tests of the test helper need
webdavclient3.

The included test documents the intended usage. The test pattern
provides full access to the webdav-exported path via the file-system,
avoiding slow webdav API calls.

TODO

- [x] Figure out threading issue that causes the webdav tests to stall on `WebDAVPath.__exit__` and specially on `server_thread.join()`. It smells like a concurrency issue. With the following patch, I can reliably prevent stalling, replacing it with  `No connection with http://127.0.0.1:43612`

```diff
--- a/datalad_next/tests/utils.py
+++ b/datalad_next/tests/utils.py
@@ -74,6 +74,9 @@ class WebDAVPath(object):
         lgr.debug('Stopping WEBDAV server')
         # graceful exit
         self.server.stop()
+        from time import sleep
+        sleep(1)
+        self.server.stop()
         # wait for shutdown
         self.server_thread.join()
         lgr.debug('WEBDAV server stopped')
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,4 +18,4 @@ update-buildsupport:
 		--squash \
 		--prefix _datalad_buildsupport \
 		https://github.com/datalad/datalad-buildsupport.git \
-		master
+		main
--- a/README.md
+++ b/README.md
@@ -21,8 +21,6 @@ software packages that intend to build on this improved set of functionality.
 # create and enter a new virtual environment (optional)
 $ virtualenv --python=python3 ~/env/dl-next
 $ . ~/env/dl-next/bin/activate
-# needs a particular datalad version (for now)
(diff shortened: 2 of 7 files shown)


mih:
This is now feature-complete. Thx @christian-monch for sorting out the threading!

Personally, I think it might be needed to replace the fixed-port setting of the helper with another approach that pick an available port. However, given the time pressure, I would not do this here. Implementing such a change later would have no impact on the test implementations that can be build with the present code.

mih:
All green!

PR #11, opened 2022-04-20
Fix test stalling

christian-monch:
This PR fixes issue #10

The reason for the stalling tests was the main thread could not successfully connect to the WebDAV-server when the server was still in the process of binding the socket address and starting to listen on the socket. The situation could occur when the server-thread was not finished with the socket setup while the main thread already tried to connect to the server.

This is fixed now by performing the server socket setup before starting the server-thread and before returning to the main thread


mih (approved):
This makes a lot of sense, and it also works reliably for me. No stalls in several dozen attempts. Thx!

PR #16, opened 2022-04-22
Use common helper get_specialremote_credential_properties()

mih:
This homogenized the code between create_sibling_webdav() and the more
generic `datalad-annex::` git remote helper.


mih:
No objections recorded.

PR #18, opened 2022-04-22
Fix TODO-items in `create_sibling_webdav`

christian-monch:
This commit fixes a number of TODO-items. Some are left to be addressed by the implementation of the `--existing`-
and the `--dry-run`-switches.

- [x] add tests for improved URL checking
- [x] add tests for `_get_url_credential`

### DIFF
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -9,7 +9,6 @@
 """High-level interface for creating a combi-target on a WebDAV capable server
  """
 import logging
-import os
 from typing import (
     Optional,
     Union,
@@ -17,6 +16,7 @@ from typing import (
 from urllib.parse import (
     quote as urlquote,
     urlparse,
+    urlunparse,
 )
 
 from datalad.distribution.dataset import (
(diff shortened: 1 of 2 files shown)


christian-monch:
@mih Is the comment starting in `create_sibling_webdav.py:421` (`#TODO dlacredential=`) still valid?

datalad_next/create_sibling_webdav.py
```diff
@@ -481,8 +484,6 @@ def _create_storage_sibling(ds, url, name, existing, credential, export):
     )
 
 
-# TODO in create-sibling-ria something like this is spaghetti-coded, it should be
-# RF'ed eventually
```
mih, line 485: I'd prefer to keep this comment, or have an issue in -core about it.

mih (approved):
LGTM, thx!

mih:
The comment with dlacredential= is valid insofar as the situation is not solved. I believe that there should be no such argument, unless a name was actually given. But I have not made up my mind yet.

datalad_next/create_sibling_webdav.py
christian-monch: Leaving it in

mih: It seems you did not actually leave it in. I have filed https://github.com/datalad/datalad/issues/6644 now.

PR #19, opened 2022-04-22
Auto-deploy credentials on `AnnexRepo.enable_remote()` (and with it also `siblings enable`)

mih:
Does not yet allow for manual credential entry on `enable_remote()`


mih:
Will merge this now, please file issues if something pops up during testing.

PR #21, opened 2022-04-22
Implement --existing for create-sibling-webdav

bpoldrack:
- Changes default for `--existing` to `error`
- Adds discovery of not yet enabled special remotes with conflicting names
- `existing=skip` skips the creation of the "conflicting" remote not the entire dataset. Meaning: A possibly specified storage sibling would be skipped independently on whether or not the git remote is skipped. Not sure whether that is advantageous or not. WDYT on that aspect, @mih, @mslw ?
- the above also somewhat implies that there are up to two results per dataset

- [x] add tests

### DIFF
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -36,6 +36,7 @@ from datalad.interface.common_opts import (
 from datalad.interface.results import get_status_dict
 from datalad.interface.utils import eval_results
 from datalad.log import log_progress
+from datalad.support.annexrepo import AnnexRepo
 from datalad.support.param import Parameter
 from datalad.support.constraints import (
     EnsureChoice,
@@ -343,16 +344,22 @@ class CreateSiblingWebDAV(Interface):
 
         # Generate a sibling for dataset "ds", and for sub-datasets if recursive
         # is True.
-        for res in ds.foreach_dataset(_dummy,
-                                      return_type='generator',
-                                      result_renderer='disabled',
-                                      recursive=recursive,
-                                      recursion_limit=recursion_limit):
-            # unwind result generator
-            for partial_result in res.get('result', []):
+        if not recursive:
+            for partial_result in _dummy(ds, ds):
                 result = res_kwargs.copy()
                 result.update(partial_result)
                 yield result
+        else:
+            for res in ds.foreach_dataset(_dummy,
+                                          return_type='generator',
+                                          result_renderer='disabled',
+                                          recursive=recursive,
+                                          recursion_limit=recursion_limit):
+                # unwind result generator
+                for partial_result in res.get('result', []):
+                    result = res_kwargs.copy()
+                    result.update(partial_result)
+                    yield result
 
         # this went well, update the credential
         credname, credprops = cred
@@ -606,17 +613,6 @@ def _yield_ds_w_matching_siblings(
       Path to the dataset with a matching sibling, and name of the matching
       sibling in that dataset.
     """
-    # in recursive mode this check could take a substantial amount of
-    # time: employ a progress bar (or rather a counter, because we don't
-    # know the total in advance
-    pbar_id = 'check-siblings-{}'.format(id(ds))
-    if recursive:
-        log_progress(
-            lgr.info, pbar_id,
-            'Start checking pre-existing sibling configuration %s', ds,
-            label='Query siblings',
-            unit=' Siblings',
-        )
 
     def _discover_all_remotes(ds, refds, **kwargs):
         """Helper to be run on all relevant datasets via foreach
@@ -636,27 +632,43 @@ def _yield_ds_w_matching_siblings(
             remotes = repo.get_remotes()
         return remotes
 
-    for res in ds.foreach_dataset(
-            _discover_all_remotes,
-            recursive=recursive,
-            recursion_limit=recursion_limit,
-            return_type='generator',
-            result_renderer='disabled',
-    ):
-        # unwind result generator
-        if 'result' in res:
-            for name in res['result']:
-                if recursive:
+    lgr.error(f"DEBUG: {recursive}, {ds}")
+    if not recursive:
+        for name in _discover_all_remotes(ds, ds):
+            if name in names:
+                yield ds.path, name
+
+    else:
+        # in recursive mode this check could take a substantial amount of
+        # time: employ a progress bar (or rather a counter, because we don't
+        # know the total in advance
+        pbar_id = 'check-siblings-{}'.format(id(ds))
+        log_progress(
+            lgr.info, pbar_id,
+            'Start checking pre-existing sibling configuration %s', ds,
+            label='Query siblings',
+            unit=' Siblings',
+        )
+
+        for res in ds.foreach_dataset(
+                _discover_all_remotes,
+                recursive=recursive,
+                recursion_limit=recursion_limit,
+                return_type='generator',
+                result_renderer='disabled',
+        ):
+            # unwind result generator
+            if 'result' in res:
+                for name in res['result']:
                     log_progress(
                         lgr.info, pbar_id,
                         'Discovered sibling %s in dataset at %s',
                         name, res['path'],
                         update=1,
                         increment=True)
-                if name in names:
-                    yield res['path'], name
+                    if name in names:
+                        yield res['path'], name
 
-    if recursive:
         log_progress(
             lgr.info, pbar_id,
             'Finished checking pre-existing sibling configuration %s', ds,
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -330,7 +330,7 @@ class CreateSiblingWebDAV(Interface):
             else:
                 dsurl = url
 
-            yield from _create_sibling_webdav(
+            return _create_sibling_webdav(
                 ds,
                 dsurl,
                 # we pass the given, not the discovered, credential name!
@@ -442,7 +442,7 @@ def _create_sibling_webdav(
                                           storage_name, storage_sibling,
                                           url, 'storage')
         else:
-            yield _create_storage_sibling(
+            yield from _create_storage_sibling(
                     ds,
                     url,
                     storage_name,
@@ -458,7 +458,7 @@ def _create_sibling_webdav(
                                           storage_name, storage_sibling,
                                           url, 'git')
         else:
-            yield _create_git_sibling(
+            yield from _create_git_sibling(
                     ds,
                     url,
                     name,
@@ -489,7 +489,7 @@ def maybe_skip_sibling(credential, credential_name, ds, existing,
                     f"{name if sibling_type == 'git' else storage_name}")
     elif existing == 'reconfigure':
         if sibling_type == 'git':
-            yield _create_git_sibling(
+            yield from _create_git_sibling(
                 ds,
                 url,
                 name,
@@ -501,7 +501,7 @@ def maybe_skip_sibling(credential, credential_name, ds, existing,
                                 if storage_sibling != 'no' else None
             )
         elif sibling_type == 'storage':
-            yield _create_storage_sibling(
+            yield from _create_storage_sibling(
                 ds,
                 url,
                 storage_name,
@@ -557,7 +557,7 @@ def _create_git_sibling(
         name=name,
         url=remote_url,
         publish_depends=publish_depends,
-        return_type='item-or-list',
+        return_type='generator',
         result_renderer='disabled')
 
 
@@ -591,7 +591,7 @@ def _create_storage_sibling(ds, url, name, credential, export, reconfigure):
             'WEBDAV_PASSWORD': credential[1],
     }):
         ds.repo.call_annex(cmd_args)
-    return get_status_dict(
+    yield get_status_dict(
         ds=ds,
         status='ok',
         action='create_sibling_webdav.storage',
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -616,39 +616,39 @@ def _yield_ds_w_matching_siblings(
         for name in _discover_all_remotes(ds, ds):
             if name in names:
                 yield ds.path, name
+        return
 
-    else:
-        # in recursive mode this check could take a substantial amount of
-        # time: employ a progress bar (or rather a counter, because we don't
-        # know the total in advance
-        pbar_id = 'check-siblings-{}'.format(id(ds))
-        log_progress(
-            lgr.info, pbar_id,
-            'Start checking pre-existing sibling configuration %s', ds,
-            label='Query siblings',
-            unit=' Siblings',
-        )
+    # in recursive mode this check could take a substantial amount of
+    # time: employ a progress bar (or rather a counter, because we don't
+    # know the total in advance
+    pbar_id = 'check-siblings-{}'.format(id(ds))
+    log_progress(
+        lgr.info, pbar_id,
+        'Start checking pre-existing sibling configuration %s', ds,
+        label='Query siblings',
+        unit=' Siblings',
+    )
 
-        for res in ds.foreach_dataset(
-                _discover_all_remotes,
-                recursive=recursive,
-                recursion_limit=recursion_limit,
-                return_type='generator',
-                result_renderer='disabled',
-        ):
-            # unwind result generator
-            if 'result' in res:
-                for name in res['result']:
-                    log_progress(
-                        lgr.info, pbar_id,
-                        'Discovered sibling %s in dataset at %s',
-                        name, res['path'],
-                        update=1,
-                        increment=True)
-                    if name in names:
-                        yield res['path'], name
+    for res in ds.foreach_dataset(
+            _discover_all_remotes,
+            recursive=recursive,
+            recursion_limit=recursion_limit,
+            return_type='generator',
+            result_renderer='disabled',
+    ):
+        # unwind result generator
+        if 'result' in res:
+            for name in res['result']:
+                log_progress(
+                    lgr.info, pbar_id,
+                    'Discovered sibling %s in dataset at %s',
+                    name, res['path'],
+                    update=1,
+                    increment=True)
+                if name in names:
+                    yield res['path'], name
 
-        log_progress(
-            lgr.info, pbar_id,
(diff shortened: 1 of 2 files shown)


datalad_next/create_sibling_webdav.py
```diff
@@ -293,12 +292,15 @@ def _dummy(ds, refds, **kwargs):
...
             _dummy,
+            recursive=recursive,
+            recursion_limit=recursion_limit,
             return_type='generator',
             result_renderer='disabled',
         ):
             # unwind result generator
             if 'result' in res:
-                yield from res['result']
+                for partial_result in res['result']:
```
mih, line 302: maybe `res.get('result',[])` then? There is no guarantee that such a field exists.

datalad_next/create_sibling_webdav.py
```diff
@@ -369,41 +371,100 @@ def _create_sibling_webdav(
...
+                        ds,
+                        url,
+                        storage_name,
+                        credential,
+                        export=export_storage,
+                        reconfigure=True,
+                    )
+            else:
+                # Shouldn't happen, since 'error' was treated upfront
+                raise ValueError(f"Unexpected value of 'existing': {existing}")
```
mih, line 400: Would be great to put this conditional into a helper fx:

```py
if storage_name in existing_siblings:
  r = dowhatisneededifthingsexist(...,existing)
```

datalad_next/create_sibling_webdav.py
```diff
@@ -369,41 +371,100 @@ def _create_sibling_webdav(
...
+                    export=export_storage,
+                    reconfigure=False,
+                    publish_depends=storage_name
+                        if storage_sibling != 'no' else None
+            )
+        results.append(r)
+
+    print ("MADE", ds, url, storage_sibling, name, storage_name, existing, credential)
+
+    return results
```
mih, line 449: What is the collection of results achieving? It seems to make the command less responsive, and prevents flow control from the outside.

mih:
> existing=skip skips the creation of the "conflicting" remote not the entire dataset. Meaning: A possibly specified storage sibling would be skipped independently on whether or not the git remote is skipped. Not sure whether that is advantageous or not. WDYT on that aspect, @mih, @mslw ?

I think it makes sense to honor the value of the `storage_sibling` parameter. If it says create both, and only one exists, the other should be created.

datalad_next/create_sibling_webdav.py
bpoldrack: The reason is that I had some trouble with the callable passed to `foreach_dataset` being a generator. But I'll have a second look.

datalad_next/create_sibling_webdav.py
```diff
@@ -296,8 +295,10 @@ def _dummy(ds, refds, **kwargs):
                                       recursive=recursive,
                                       recursion_limit=recursion_limit):
             # unwind result generator
-            if 'result' in res:
-                yield from res['result']
+            for partial_result in res.get('result', []):
+                result = res_kwargs.copy()
+                result.update(partial_result)
+                yield result
```
mih, line 301:
```suggestion
                yield dict(res_kwargs, **partial_result)
```

datalad_next/create_sibling_webdav.py
```diff
@@ -108,13 +108,12 @@ class CreateSiblingWebDAV(Interface):
...
             doc="""action to perform, if a (storage) sibling is already
-            configured under the given name and/or a target already exists.
-            In this case, a dataset can be skipped ('skip'), an existing target
-            repository be forcefully re-initialized, and the sibling
-            (re-)configured ('reconfigure'), or the command be instructed to
-            fail ('error').""", ),
+            configured under the given name.
+            In this case, a sibling creation can be skipped ('skip') or the
+            sibling (re-)configured ('reconfigure'), or the command be
+            instructed to fail ('error').""", ),
```
mih, line 116: What if the target already exists? The condition is removed from the docs, does it mean it won't be handled?

bpoldrack: Yes. I don't think it can be handled. How would we discover it and what would we need to to differently in such case?

mih: It could offer to verify that the remote state actually matches the local expectations. Here is a function that implemented that (I mentioned it in the chat earlier).

https://github.com/mih/datalad-mihextras/blob/6245ed3192040624c7db7d25c1fcf10e4b49976d/datalad_mihextras/export_to_webdav.py#L243-L271

The key point is: What is a user supposed to do, if there is conflicting content on the remote? The documentation is now silent about this -- which is suboptimal even if it turns out that really nothing can be done (which I am not convinced of).

bpoldrack: Conflicting what with? This is not push, but create.

bpoldrack: I think the only option is to say: We talk directly (not via special remote) to the webdav location to find whether there's anything at all and consider this `existing`.

datalad_next/create_sibling_webdav.py
bpoldrack: Fixed.

datalad_next/create_sibling_webdav.py
bpoldrack: Done.

datalad_next/create_sibling_webdav.py
bpoldrack: For now just added respective entry to the command's doc.

datalad_next/create_sibling_webdav.py
```diff
@@ -69,6 +70,9 @@ class CreateSiblingWebDAV(Interface):
     server needs to be specified for creating a sibling. However, the sibling
     setup can be flexibly customized (no storage sibling, or only a storage
     sibling, multi-version storage, or human-browsable single-version storage).
+    The target location is currently expected by this command to have no
+    potentially conflicting content. Please make sure of that beforehand!
+    Such content may be overwritten.
```
mih, line 75: I think we need to be explicit here: what is "potentially conflicting content" exactly? When would it be overwritten?

datalad_next/create_sibling_webdav.py
mih: Left a comment there

datalad_next/create_sibling_webdav.py
```diff
@@ -343,16 +344,22 @@ def _dummy(ds, refds, **kwargs):
...
         # Generate a sibling for dataset "ds", and for sub-datasets if recursive
         # is True.
-        for res in ds.foreach_dataset(_dummy,
-                                      return_type='generator',
-                                      result_renderer='disabled',
-                                      recursive=recursive,
-                                      recursion_limit=recursion_limit):
-            # unwind result generator
-            for partial_result in res.get('result', []):
+        if not recursive:
```
mih, line 350: This special case could be implemented in a cheaper fashion using `recursion_limit=0` if not recursive.

```sh
(datalad-dev) mih@meiner /tmp/rec (git)-[master] % dl foreach-dataset -R 0 ':'
foreach-dataset(ok): /tmp/rec (dataset)
(datalad-dev) mih@meiner /tmp/rec (git)-[master] % dl foreach-dataset ':'
foreach-dataset(ok): /tmp/rec (dataset)
foreach-dataset(ok): /tmp/rec/s1 (dataset)
foreach-dataset(ok): /tmp/rec/s2 (dataset)
```

bpoldrack: Agree. Will do.

datalad_next/tests/test_create_sibling_webdav.py
```diff
@@ -368,3 +371,136 @@ def query(self, *args, **kwargs):
...
+@with_credential(
+    'dltest-mywebdav', user=webdav_cred[1], secret=webdav_cred[2],
+    type='user_password')
+@with_tree(tree={'sub': {'f0': '0'},
+                 'sub2': {'subsub': {'f1': '1'},
+                          'f2': '2'},
+                 'f3': '3'})
+@with_tempfile
+@serve_path_via_webdav(auth=webdav_cred[1:])
+def test_existing_switch(localpath, remotepath, url):
```
mih, line 385: Here is an interesting twist to this test that covers a conflict scenario that will occur often:

When a webdav sibling is in annex-mode, the annex object tree will be in the root. This means that for recursively deposited subdatasets, they is a built-in conflict (potential) for any  subdataset with a three-character hex-compliant mountpoint. When a git-sibling is also included, there will be a hard conflict with `3f7` and `eb3`. These should be handled gracefully somehow -- worth including here.

datalad_next/tests/test_create_sibling_webdav.py
```diff
@@ -368,3 +371,136 @@ def query(self, *args, **kwargs):
...
+        res,
+        action='create_sibling_webdav',
+        status='notneeded',
+        type='dataset',
+        path=subsub.path,
+    )
+
+    # should fail upfront with first discovered remote that already exist
+    assert_raises(IncompleteResultsError, ds.create_sibling_webdav, f'{url}',
+                  storage_sibling='yes', existing='error', recursive=True, **ca)
```
mih, line 480: I think it would be more useful to give `on_failure='ignore'` and test for the particular result (composition) that would trigger `IncompleteResultsError`. The former is relevant here, there latter is the present behavior of datalad core.

datalad_next/create_sibling_webdav.py
mih: I believe the only thing that is touched on the server by this command is the root dir of each dataset (i.e. `mkdir`).

datalad_next/create_sibling_webdav.py
```diff
@@ -327,7 +330,7 @@ def _dummy(ds, refds, **kwargs):
             else:
                 dsurl = url
 
-            return _create_sibling_webdav(
+            yield from _create_sibling_webdav(
```
mih, line 333: I believe that this is not actually necessary, because it is OK to "return" a generator, and still have it be a generator. I used `return`, because `foreach-dataset` takes whatever is returned and places it into the result property. It neither expects a generator, nor does it have any handling for such a case. I consider this semantically more appropriate by a small margin.

datalad_next/create_sibling_webdav.py
```diff
@@ -422,41 +429,107 @@ def _create_sibling_webdav(
     # simplify downstream logic, export yes or no
     export_storage = 'export' in storage_sibling
 
+    existing_siblings = [r[1] for r in
+                         _yield_ds_w_matching_siblings(ds,
+                                                       (name, storage_name),
+                                                       recursive=False)
+                         ]
```
mih, line 436: A plea for

```suggestion
    existing_siblings = [
        r[1] for r in _yield_ds_w_matching_siblings(
            ds,
            (name, storage_name),
            recursive=False)
    ]
```

datalad_next/create_sibling_webdav.py
```diff
@@ -422,41 +429,107 @@ def _create_sibling_webdav(
...
-            existing,
-            credential,
-            export=export_storage,
-        )
+        if storage_name in existing_siblings:
+            yield from maybe_skip_sibling(credential, credential_name, ds,
+                                          existing, export_storage, name,
+                                          storage_name, storage_sibling,
+                                          url, 'storage')
+        else:
```
mih, line 444: TBH: I do not like this pattern.

Every call is now duplicated, because one decision re `existing` (namely is there a conflict) is tested outside, and the rest inside `maybe_skip_sibling`.

`maybe_skip_sibling` is also maybe-skip-or-reconfigure`. And carries a huge signature only to be able to set `reconfigure=True`.

Why not 

- RF into `_skip_sibling(name, existing_siblings, existing) -> bool`
- replace `if (storage_)name in existing_siblings` with `if _skip_sibling((storage_)name, existing_sibling, existing)`
- perform all result yielding here
- perform all calls to `_create(_storage)_sibling` here, and once
- set their `reconfigure` to `storage_name in existing_siblings` (cached)

should half the code

datalad_next/create_sibling_webdav.py
```diff
@@ -422,41 +429,107 @@ def _create_sibling_webdav(
...
+    else:
+        # Shouldn't happen, since 'error' was treated upfront
+        raise ValueError(f"Unexpected value of 'existing': {existing}")
 
 
 def _create_git_sibling(
-        ds, url, name, existing, credential_name, credential, export,
-        dependency=None):
+        ds, url, name, credential_name, credential, export, reconfigure,
+        publish_depends=None):
```
mih, line 519: I passed `existing`, because I was thinking about future patches, and what information would be valuable in this function. Now it is limited to handle the reconfigure case, and no, e.g., `replace` that may come in the not too distant future.

datalad_next/create_sibling_webdav.py
```diff
@@ -470,32 +543,29 @@ def _create_git_sibling(
...
+    return ds.siblings(
+        action="add" if not reconfigure else "configure",
         name=name,
         url=remote_url,
-        # TODO probably needed when reconfiguring, but needs credential patch
-        fetch=False,
-        publish_depends=dependency,
-        return_type='generator',
+        publish_depends=publish_depends,
+        return_type='item-or-list',
```
mih, line 551: Unclear why this change to `return_type` is needed.

I would also like to keep `fetch=False` to indicate that this is the desired behavior.

datalad_next/create_sibling_webdav.py
```diff
@@ -470,32 +543,29 @@ def _create_git_sibling(
...
         result_renderer='disabled')
 
 
-def _create_storage_sibling(ds, url, name, existing, credential, export):
+def _create_storage_sibling(ds, url, name, credential, export, reconfigure):
```
mih, line 555: Same comment as above, passing `existing` is more future proof and hardly more complex.

datalad_next/create_sibling_webdav.py
```diff
@@ -422,41 +429,107 @@ def _create_sibling_webdav(
...
+                    ds,
+                    url,
+                    storage_name,
+                    credential,
+                    export=export_storage,
+                    reconfigure=False,
+                )
+
     if 'only' not in storage_sibling:
-        yield from _create_git_sibling(
```
mih, line 435: The `yield from` pattern was desired, because the minimalistic implementation now has only one result per call, but this is hardly going to stay.

datalad_next/create_sibling_webdav.py
```diff
@@ -470,32 +543,29 @@ def _create_git_sibling(
         # e.g., it is not uncommon for credentials to be named after URLs
         remote_url += f'&dlacredential={urlquote(credential_name)}'
 
-    yield from ds.siblings(
-        # TODO set vs add, consider `existing`
-        action="add",
+    return ds.siblings(
```
mih, line 546: The kinda assumed that `siblings` will never give more than one result. I don't think this is a robust setup.

datalad_next/create_sibling_webdav.py
```diff
@@ -540,37 +610,61 @@ def _yield_ds_w_matching_siblings(
...
+                            for k, v in repo.get_special_remotes().items()]
+                           )
+        else:
+            remotes = repo.get_remotes()
+        return remotes
+
+    if not recursive:
+        for name in _discover_all_remotes(ds, ds):
+            if name in names:
+                yield ds.path, name
```
mih, line 635: Please `return` here and dedent the rest of the function.

datalad_next/create_sibling_webdav.py
bpoldrack: Think so, too. Except: The git remote comes with it's own if enabled, right?

bpoldrack: But still: Any other content can still be in conflict with a subsequent push. And we can't detect it, if it wasn't created by another export, I think.

datalad_next/create_sibling_webdav.py
bpoldrack: Can see that. On the other hand: That distinction implies that the use of return vs yield needs to be reevaluated with every change in implementation. So - not sure. No too strong opinion.

datalad_next/create_sibling_webdav.py
mih: So nothing will be overwritten by *this* command.

datalad_next/create_sibling_webdav.py
mih: Turns out to be false. Would not work with `**` unless yielded.

mih: Done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/create_sibling_webdav.py
mih: done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/create_sibling_webdav.py
mih: Done

datalad_next/tests/test_create_sibling_webdav.py
mih: Done

datalad_next/tests/test_create_sibling_webdav.py
mih: Done.

At least superficially. The real test would be a subsequent `push`

datalad_next/create_sibling_webdav.py
mih: I went for a minimal statement

> This command does not check for conflicting content on the WebDAV server!

mih:
Off it goes. Thx @bpoldrack

PR #22, opened 2022-04-25
Strip existing=None possibility from Parameter constraint

mih:
This is no meaningfully defined behavior (discussed elsewhere).

Extracted from https://github.com/datalad/datalad-next/pull/21


mih:
MacOS brew setup disfunctional right now. Nevertheless merging.

PR #23, opened 2022-04-25
Complete desired credential handling for `create_sibling_webdav()`

mih:
- when a credential is not given, auto-discovery is wanted, and not just for the initial create, but also for any subsequent git-operation
- specifically test scenarios with declared vs. discovered credentials
- define desired behavior in docs
- test with credential name that needs to be URL-encoded


mih:
Tests are happy, let's go!

Issue #29, opened 2022-04-25
Add `datalad.clone.url-substitute.webdav(s)`


mih:
This is a mechanism provided by the core package. We could use it, to make the specification of webdav-based `datalad-annex::` URLs more convenient.

We could automatically turn

`webdav://example.com/path`

into

`datalad-annex::?type=webdav&encryption=none&url=http://example.com/path`

which is much less cryptic enter (analog for `https` could be `webdavs://`)

PR #35, opened 2022-04-26
Support cloning dataset from WebDAV with simplified URLs

mih:
For common cases (no encryption, no particular credentials) we can
bypass the need for the monster `datalad-annex::` URLs that make all
git-annex special remote internals explicit.

For WebDAV this introduces a few convenience "protocols" that are
automatically converted into the matching `datalad-annex::`
specification.

- `webdav://` -> http://, exporttree=yes
- `webdavs://` -> https://, exporttree=yes
- `webdav-obj://` -> http://, exporttree=no
- `webdavs-obj://` -> https://, exporttree=no

Mapping is implemented using the standard clone URL substitution
mechanism, and only if the given URL contains no query string.

This is a mere convenience, full `datalad-annex::` URL specification
remains supported.

The choice for `webdav://` to imply `exporttree=yes` reflects my
oppinion that this should turn out to be the more frequent use case
for the target audience that exchanges data via WebDAV.

Fixes datalad/datalad-next#29


mih:
Thx! Responses below:

> * webdav urls typically include "dav" or "webdav" somewhere, making it something like `webdavs://service-name.org/webdav/path` (when reading I see webdav twice, which looks slightly weird, but overall still looks nice)

I saw no better and equally reliable way than to repurpose the protocal. It is also what we do for `osf://`

> * wasn't immediately intuitive for me that `webdav://` would be `exporttree=yes`, and `webdav-obj://` would be `exporttree=no`, but in the end I agree with your reasoning that export might be the more common case

Yes, do note that the particular choice of `-obj` is conditional on a pending decision re the `create-sibling-webdav` API.

> * is webdav over http (no s) common enough to support it?

It is a trivial mapping, and extra work to not support it.

> * for me, no particular credentials wouldn't probably be the common case (I like the ability to be explicit), but I can see it being useful

I am not sure how this would work. These URLs are what one would provided for people to copy/paste from a website. Even with the `datalad-annex::` URLs there is no way for a service provider to know the ID of the credential you would want to use personally.

mih:
I decided to strip `webdav(s)-obj://` again. It is too ugly, there is time pressure and I have no better idea.

The remaining config shows the pattern, though, and any user could follow it to create their own.

PR #43, opened 2022-04-27
Replace `storage_sibling` parameter with `--mode` switch

### LINKED ISSUE #36: Replace `create-sibling --storage-sibling`
Status quo
```
  --storage-sibling {yes|export|only|only-export|no}
                        Both Git history and file content can be hosted on WebDAV. With 'yes', a storage sibling and a Git repository sibling are
                        created ('yes'). Alternatively, creation of the storage sibling can be disabled ('no'), or a storage sibling can be created
                        only and no Git sibling ('only'). The storage sibling can be set up as a standard git-annex special remote that is capable
                        of storage any number of file versions, using a content hash based file tree ('yes'|'only'), or as an export-type special
                        remote, that can only store a single file version corresponding to one unique state of the dataset, but using a human-
                        readable data data organization on the WebDAV remote that matches the file tree of the dataset ('export'|'only-export').
                        When a storage sibling and a regular sibling are created, a publication dependency on the storage sibling is configured for
                        the regular sibling in the local dataset clone. Constraints: value must be one of ('yes', 'export', 'only', 'only-export',
                        'no') [Default: 'yes']
```

Problems:

- docs and labels rely on the behavior implied by `yes` to be the default that needs no explanation, or an explanation that requires technical knowledge
- the anticipated standard usage is `--storage-sibling export`

Proposal

```
--mode {annex|filetree|annex-only|filetree-only|git-only}
```

Advantages

- Similar to OSF (albeit not the same, because that one uses `exportonly`)
- All alternatives are equally completely specified with their labels (default switching is easier)

mih:
Fixes #36

### DIFF
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -170,29 +170,37 @@ class CreateSiblingWebDAV(Interface):
             command be instructed to fail ('error').""", ),
         recursive=recursion_flag,
         recursion_limit=recursion_limit,
-        storage_sibling=Parameter(
-            args=("--storage-sibling",),
-            dest='storage_sibling',
+        mode=Parameter(
+            args=("--mode",),
             constraints=EnsureChoice(
-                'yes', 'export', 'only', 'only-export', 'no'),
-            doc="""Both Git history and file content can be hosted on WebDAV.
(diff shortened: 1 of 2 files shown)


mih:
Will merge now to simplify last mile to release.

PR #58, opened 2022-05-16
Fix metadata (URL) in setup.cfg

mslw:
The URL field in setup.cfg metadata is rendered on PIP as "Homepage". 
This will now point to next's github project instead of extension template.


mih (approved):
Thanks for catching that!

PR #66, opened 2022-05-24
Add --private to datalad-create

bpoldrack:
This provides a patch to add a --private switch to datalad-create, that
would set `annex.private=true` (needs to be done before annex-init) for
the new dataset. With that neither availability nor the existence of
this location will be recorded in the git-annex branch.


mih:
It seems there is no need anymore.

Issue #17, opened 2022-04-22
Consolidate matching siblings-finding between create_sibling_ria and create_sibling_webdav


christian-monch:
`create_sibling_ria` has code to determine matching siblings. Currently `create_sibling_webdav` has a function that does similar things, i.e. `_yield_ds_w_matching_siblings`.

Once the functions is perfected, the implementations should be consolidated. That means the function should probably move to core and be used by both sibling implementations.

(This issue replaces a TODO in `create_sibling_webdav.py`

mih:
This is the companion in -core (where it would need to be fixed): https://github.com/datalad/datalad/issues/6644

Closing this one. Thx!

PR #67, opened 2022-05-25
MNT: Use helper function from core

adswa:
Companion PR for https://github.com/datalad/datalad/pull/6706, in response to
#17.
Can only be merged if https://github.com/datalad/datalad/pull/6706 gets into core.


adswa:
sure, that makes sense!

adswa:
Filed  #70

PR #71, opened 2022-06-01
Must reset the push docstring before regenerating it

mih:
Otherwise the builddoc decorator will amend the existing pre-generated
docstring, effectively duplicating it.


mih:
The test failure is non-specific to this PR -- also the mainline is failing right now.

PR #73, opened 2022-06-08
Port test suite to pytest

mih:


mih:
This works -- minus the problem that is already present in the mainline.

Issue #72, opened 2022-06-08
git-annex > 10.20220504 breaks `git-clone` from `datalad-annex::...exporttree=yes` (any special remote type)


mih:
TL;DR:

git-annex now forcefully and unconditionally distrusts `exporttree=yes` special remotes. This makes it seemingly impossible for the `datalad-annex::` remote helper implementation to use `annex fsck` to probe a specific key as a remote.

----

Mainline fails https://ci.appveyor.com/project/mih/datalad-next/build/job/98jegq2gaw7i04rg

```
eq_(ds.repo.get_hexsha(ds.repo.get_corresponding_branch()),
1386AssertionError: 'b1c9aef80331c775dc06372d5cb2536543a15455' != None
1387
```

Possibly a git version issue. Does not replicate for me locally.

Update: Nope, git version in both cases is 2.36.1

Looking into the failure of `datalad_next.tests.test_create_sibling_webdav.test_common_workflow_export` specifically: this is near-identical to `datalad_next.tests.test_create_sibling_webdav.test_common_workflow_explicit_cred` (which passes), but uses the `--filetree` mode for the webdav sibling. The dataset clone the respective check is failing on, indeed contains no commits.

Rerunning the `clone` call does not produce any messages.

Manually looking at the repository deposited via `push` on webdav (and the underlying filesystem location) reveals no problem -- looks good with all expected content/branches/commits present.

So the issue is clone from a `filetree`-mode deposit.

Appveyor uses a git-annex snapshot `git-annex version: 10.20220505-g2df856370`, locally I have `git-annex version: 10.20220504`

Deploying this snapshot locally also triggers the failure. This points fair and square to a behavior change in git-annex past 10.20220504.

Git-annex did indeed do two export-related changes since that release:

```
  * Special remotes with importtree=yes or exporttree=yes are once again
    treated as untrusted, since files stored in them can be deleted or
    modified at any time.
    (Fixes a reversion in 8.20201129)
  * Special remotes using exporttree=yes and/or importtree=yes now
    checksum content while it is being retrieved, instead of in a separate
    pass at the end.
 ```

The breakage is also detected by the more direct test of the `datalad-annex` remote helper: `datalad_next.gitremote.tests.test_datalad_annex.test_export_remote` also failing to locate any content in a clone, immediately after cloning. Here is a demonstrator:

```
% datalad create --no-annex myds
% cd myds
[...]

PR #74, opened 2022-06-09
(Temporarily?) discontinue probing remotes for XDLRA--refs

mih:
This is no longer possible with git-annex 10.20220505 and later.

Ideally, we would get this feature back. See
https://git-annex.branchable.com/bugs/10.20220525_cannot_probe_for_untrusted_remote_key/

But for now a forced-download will have to do.

Closes datalad/datalad-next#72


bpoldrack (approved):
Agree with the approach and see no issue with the code.

Issue #75, opened 2022-06-09
Replace use of `fsck` in `datalad-annex::` with `checkpresentkey`


mih:
See https://github.com/datalad/datalad-next/issues/72#issuecomment-1151398665

PR #76, opened 2022-06-10
Switch `datalad-annex::` to use git-annex plumbing 'transferkey'

mih:
This avoids the complications with changing behavior of `git-annex-fsck`
with 10.20220525 (see
https://github.com/datalad/datalad-next/issues/72).  Moreover, it allows
for moving to an opportunistic download that just tries, instead of
having to check with the remote first. This saves a request, and should
translate to measurable speed-up for slow remotes (e.g. WebDAV).

Closes datalad/datalad-next#75

### DIFF
--- a/datalad_next/gitremote/datalad_annex.py
+++ b/datalad_next/gitremote/datalad_annex.py
@@ -891,39 +891,26 @@ class RepoAnnexGitRemote(object):
 
         self.log("Get refs from remote")
         ra = self.repoannex
+
         # in case of the 'web' special remote, we have no actual special
         # remote, but URLs for the two individual keys
         sremotes = ra.get_special_remotes()
+        # if we do not have a special remote reported, fall back on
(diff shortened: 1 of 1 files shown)


mslw (approved):
If I understand correctly:
- in #72, the `git-annex-checkpresentkey` command ([docs](https://git-annex.branchable.com/git-annex-checkpresentkey/)) was suggested as a more suitable replacement for `git-annex-fsck`
- but since this is always followed by `git-annex-get`, the two can be replaced with a single `git-annex-transferkey` ([docs](https://git-annex.branchable.com/git-annex-transferkey/)); errors of which are handled in the case of missing keys
- the PR does just that

I didn't know any of these things before, but looking at the docs and the diff, this seems clear-cut to me.

mslw:
The CI checks pass, and there's two reviews, so I'm merging.

Thanks a lot for getting to the bottom of the issue @mih!

Issue #26, opened 2022-04-25
`datalad-annex::` partially broken on windows


mih:
Here is the tracking issue of the responsible functionality in git-annex:

https://git-annex.branchable.com/bugs/Fails_to_drop_key_on_windows___40__Access_denied__41__/?updated

PR #77, opened 2022-06-10
Enable full feature-set for `datalad-annex::` on windows

mih:
#26 is still valid with today's git-annex. This PR implements a rather draconian workaround: Instead of dropping the repoannex keys, it wipes out the entire object tree of the local `repoannex` utility repo.

This is possible, because this utility repo is constructed on the fly from the URL specification, and only contains the two keys that need to be dropped. However, given the bluntness of the approach, it is limited to the platform on which the original issue is happening: windows.


mih:
The coverage report is confusing. When starting this PR I specifically confirmed that the reenabled test actually fails on appveyor windows, and it nevertheless reports that same code it as "not covered". I'd say this is false and can be ignored here.

PR #84, opened 2022-07-10
Aid changelog composition via `scriv`

mih:
This change provides everything necessary to allow developers to use
https://github.com/nedbat/scriv to provide changelog items.

In contrast to the practice in datalad-core, this neither means
recycling commit messages, nor C&P from PR descriptions.

The basic goals for an appropriate changelog are:

1. Changelog items target users, not developers, hence should be
   written for this audience specifically
2. A changelog should give any user a reasonably self-contained
   description.

Workflow:

1. A changeset is composed in an arbitrary way, typically posted
   as a PR.
2. At any time before a merge, a developer can run
   `scriv create` to generate a pre-formatted changelog snippet.
3. This snippet can be processed/edited in any way a developer
   sees fit (with scriv --add/edit or without), as long as it is
   included in the changeset as a dedicated file in changelog.d/
4. When preparing a release, the release manager can run
   `scriv collect --version X.X.X` to assemble a changelog for
   a given version. `scriv` will include a present snippet in
   the changelog and deletes the snippets afterwards.

With this approach, a targeted changelog can be generated from any
branch, without having to identify relevant commits or merges. The
mere presence of a changelog snippet documents the need for its
inclusion. Merging a released branch into another branch,
automatically removes changelog items for released changes.

`scriv` is primarily a convenience layer, pretty much all steps can
be performed without it.

Moving changelog composition outside of commit messages and PR
descriptions helps maintain clarity re target audience expectations.

While `scriv` does not support all the little things we tried in the past with changelogs, it also does not get in the way of additional custom processing, either at the level of individual snippet or the entire changelog.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,10 @@
+### PR checklist
+
+- [ ] If this PR is not complete, select "Create Draft Pull Request" in the pull request button's menu.
+  Consider using a task list (e.g., `- [ ] add tests ...`) to indicate remaining to-do items.
+- [ ] Provide an overview of the changes you're making and explain why you're proposing them. Ideally, include them as a new file in `changelog.d/`
+- [ ] Include `Fixes #NNN` somewhere in the description if this PR addresses an existing issue.
+- [ ] If you would like to list yourself as a DataLad contributor and your name is not mentioned please modify .zenodo.json file.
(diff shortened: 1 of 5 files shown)


bpoldrack:
Seems nice, let's try. Maybe put scriv in requirements-devel?

mih:
I also thought about that, but it is not needed anywhere where this file's content is routinely interpreted. So I concluded that it is just making things slower for little benefit.

PR #91, opened 2022-07-23
Equip `GitRepo.diffstatus()` with typechange detection

mih:
This can be a basis for implementing patch generation for `diff`.

This PR is the companion to https://github.com/datalad/datalad/pull/6797 (which has been abandoned).

TODO
- [ ] https://github.com/datalad/datalad/pull/7088 is added a slowing workaround to `GitRepo.__save_add_submodules()` due to the lack of type-annotation ability in core. In order to preserve the speed when handling large numbers of subdatasets, we better replace `GitRepo.__save_add_submodules()` in here too.

### DIFF
--- a/datalad_next/patches/__init__.py
+++ b/datalad_next/patches/__init__.py
@@ -1,4 +1,5 @@
 from . import (
+    gitrepo,
     annexrepo,
     configuration,
     create_sibling_ghlike,
--- a/datalad_next/patches/gitrepo.py
+++ b/datalad_next/patches/gitrepo.py
@@ -1,8 +1,16 @@
+from itertools import chain
 import logging
 
+from datalad.config import (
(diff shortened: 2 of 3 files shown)


mih:
I am closing this PR after a year of nothingness. It can now join the original https://github.com/datalad/datalad/pull/6797 in the afterlife.

The underlying issues remain, of course. Working on  https://github.com/datalad/datalad-next/pull/361 I found different approaches to do this. Rather than an expensive patch (the approach here), I think a "change iterator" around `git diff-index` (and friends) would be a simpler approach.

The implementation that I am now abandoning here is already a half-step in that direction (`git diff-files`), and became the workhorse in `gooey-status-light`. The substantial speed differences compared to `status` in -core make it evident that going further in this direction is useful.

PR #92, opened 2022-07-27
NF: datalad tree command

### LINKED ISSUE #78: What about a `datalad tree` command?
Idea and implementation in `sh` by @catetrai at https://github.com/catetrai/tree-datalad

catetrai:
New command `datalad tree` for displaying directory/dataset hierarchies.

Closes #78

Note:

- Tests are of the 'end-to-end' kind for now. Can add more proper unit tests once we consolidate the implementation.
- As discussed in #78, I have included only long-form parameters for now. I think it would be worth considering short-form variants as well, to make it feel similar to `tree`. My suggestion would be `-L` for `--depth` (consistent with `tree` syntax) and `-R` for `--dataset-depth` (consistent with `--recursion-limit` in other datalad commands).
- Performance may be an issue in case of large and deep dataset hierarchies. I played around with running the command on the 'mother of all superdatasets' at datasets.datalad.org, installing some datasets recursively. To me it looks like a classic Not Great, Not Terrible™. There may very well be a more efficient algorithm for the search, so let me know if you think it's worth looking into it.

As a first-time contributor, this was really fun! The development environment (including tests and building docs) was straightforward to set up following the instructions. I was also able to reuse helpers in the utils modules (e.g. for creating test directory trees) which saved me lots of time.

Looking forward to your review! Please be brutal :-)


adswa:
Hey @catetrai, this is really cool, thanks much for your PR! This is mostly just a heads-up that some people who you'd want to review this are currently on vacation and might not see this immediately. I just checked out your branch an ran things (no time yet to look at code), and this is *super* cool! I'll see if I can take a closer look tomorrow, and also find out what the cause of the CI failures is; things pass locally, but there are wide-spread failures in the CI such as

```
[ERROR] Internal error, cannot import interface 'datalad_next.tree': TypeError(Expected maxsize to be an integer or None) 
making output directory... done
```

catetrai:
Thank you @adswa for the heads up! :) I also saw that some tests are failing on Windows (something to do with terminal colors), so I will try to fix those in the mean time. And then see what tests fail next :D

mih:
Wonderful @catetrai ! I only had a brief first look (I will do the review on the weekend), and tried running it. There are some aspects I need to dig into, and possible come up with recommendations for changes, but overall looks great and works, as far as I can see after a few min. I now approved this PR for running the full test battery.

mih (commented):
I gave it another/closer look. I think it may be best to approach two things first:

1. A separation of tree generation and tree rendering (see inline comments)
2. Possible performance implications of the design

Re (2): If I got it right, you are essentially doing a recursive walk on the directory tree. For each directory a helper determines whether a) there is a dataset root, and b) how deep that dataset is in the tree. This could be expensive (I believe it is done for each dataset independently), but anything could be expensive with a large tree.

So my Q is, have you considered collecting some information upfront and maybe go by datasets with the main recursion, rather than by directory? This could be faster with deep hierarchies.

Here is a sketch:

- take incoming `path` and give to `get_dataset_root()` -- this will tell you, whether you are in a dataset and where it is.
- if needed for depth, iterate over `Dataset(..).get_superdataset(topmost=False)` until you found the root of the hierarchy, and count the layers -- only once
- now walking the tree from the given path
  - with the `Dataset.subdatasets(path=<given path>)` get the next layer of subdatasets underneath the directory you are processing, until you hit one of them, each directory is just that, and the dataset depth is unchanged

Such an implementation is likely more complex. Maybe you have considered such an approach already. I don't think it is much different from what you went for, but it would be using some additional information on the side to avoid unconditional exploration for each tree node.

So far for now. I will dig into the PR more, a bit later.

catetrai:
Hi @mih,

thank for you for the comments!

Regarding performance:

You are correct, there is a lot of repeated traversing of the same nodes. In fact, within the main node generator that starts from the tree root, a new subtree with unlimited depth is generated for each directory (though it's heavily pruned by the exclusion conditions). The worst-case search scenario is datasets with huge subtrees _without_ any datasets underneath.

In an earlier implementation (see link below), I ran the node generator just once, with unlimited depth to find all datasets located anywhere in the tree. Whenever I hit a dataset, I would generate all its parents first. I stored the visited parents to prevent yielding the same node twice, if I revisited them later. This way, there was minimal repetition of traversed nodes (at the cost of storing a partial history, but it's a cost I'd be willing to pay if it would significantly speed up execution).

https://github.com/datalad/datalad-next/blob/12b61601ef524a19360cfa3d356530b0986f2359/datalad_next/tree.py#L513

However I abandoned this approach because it was relatively slow (it still had to do a full, unlimited-depth tree traversal). Also, since we could not predict the final depth of each subtree in advance, it was not possible to identify the "last child" of each parent, needed to display the correct branch-tip symbol. So the output would look broken depending on the input directory structure (see failing tests at [12b616](https://github.com/datalad/datalad-next/tree/12b61601ef524a19360cfa3d356530b0986f2359)).

Some added complexity also comes from wanting to display (1) datasets that are located below another dataset, but are not registered as subdatasets of the parent (they would not be recognized by `Dataset.subdatasets(path)`), and (2) not-installed datasets (though only if they are registered subdatasets -- otherwise, no idea how to identify them). I personally find both options useful for detecting potentially confusing installations (e.g. `create subds` without `-d`, or non-recursive `clone`), so I would still try to work them into the search strategy.

I will mull over your suggestions again and see if I can come up with an improvement.

catetrai:
Btw, just changed the formatting of the print output so the dataset marker is located before the path instead of after. I think it looks tidier and more readable.

datalad_next/tree.py
```diff
@@ -0,0 +1,793 @@
...
+    EnsureNone,
+    EnsureStr, EnsureInt, EnsureRange,
+)
+from datalad.support import ansi_colors
+
+lgr = logging.getLogger('datalad.local.tree')
+
+
+@build_doc
+class TreeCommand(Interface):
```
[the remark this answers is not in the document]
catetrai: True. When trying to follow the convention, I got naming exhaustion syndrome from trying to come up with a synonym for `Tree` to name the actual class for tree generation :-). I thought it was fine since we can explicitly set the class name in the command suite definition. Or would you rather keep the naming convention for the sake of consistency?

datalad_next/tree.py
```diff
@@ -0,0 +1,793 @@
...
+             code_py="tree('/tmp', dataset_depth=2, include_hidden=True)",
+             code_cmd="datalad tree /tmp --dataset-depth 2 --include-hidden"),
+        dict(text="Display first- and second-level subdatasets and their "
+                  "contents, up to 1 directory deep within each dataset",
+             code_py="tree(dataset_depth=2, depth=1)",
+             code_cmd="datalad tree --dataset-depth 2 --depth 1"),
+    ]
+
+    @staticmethod
+    @datasetmethod(name='tree')
```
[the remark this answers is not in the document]
catetrai: From my POV, the command is most effective when used to describe a context that can include multiple datasets, like a directory with lots of superdatasets.

Do you think there is a use case for calling `tree()` specifically on a `Dataset`? After all, a `Dataset` already knows about its own subdatasets (and can limit dataset depth using `-r -R`), whether they are installed, etc. So it seems like `tree` would not add much value.

If I understand correctly, the only difference would be in the way to invoke the command (as a method of `Dataset` vs. a standalone function from `datalad.api`)?

datalad_next/tree.py
```diff
@@ -0,0 +1,793 @@
...
+                max_dataset_depth=dataset_depth,
+                exclude_node_func=build_excluded_node_func(
+                    include_hidden=include_hidden, include_files=include_files)
+            )
+        else:
+            tree = Tree(
+                Path(path),
+                max_depth=depth,
+                exclude_node_func=build_excluded_node_func(
+                    include_hidden=include_hidden, include_files=include_files)
```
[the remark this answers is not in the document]
catetrai: Good point. Done.

datalad_next/tree.py
```diff
@@ -0,0 +1,793 @@
...
+            # print one line at a time to improve UX / perceived speed
+            print(line)
+        print("\n" + tree.stats() + "\n")
+
+        # return a generic OK status
+        yield get_status_dict(
+            action='tree',
+            status='ok',
+            path=path,
+        )
```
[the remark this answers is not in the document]
catetrai: Yes, good point! I had a feeling that the result set could be used more effectively (as opposed to an impoverished dict with just an "ok" status and the `path` attribute).

Since we want to print the tree nodes as soon as they are computed, it makes sense to use the node as "reporting unit". The result dict can yield from the `generate_nodes()` function, using the node attributes as custom dict keys. I will try to create a custom results renderer for it.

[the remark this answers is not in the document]
catetrai: I added a custom renderer (one result for each tree node) and a custom summary renderer, which includes both the tree 'report line' and the generic renderer (just to show the 'ok' status -- there is no error status that needs reporting, right?). It works with `-f` 'tailored' (the default tree display), 'generic', 'json[_pp]'. I only added the most relevant custom attributes in the results dict -- let me know if I should include fewer/more.

catetrai:
> I will mull over your suggestions again and see if I can come up with an improvement.

Hi @mih,

I made 2 big changes to the tree traversal:

#### 1. Caching git operations

Profiling revealed that a major bottleneck was the repeated calls to `ds.subdatasets()` in the `get_superdatasets()` function.

We query direct superdatasets in two places:

- in `is_dataset()`, when trying to detect non-installed subdatasets (only for empty directories)
- in the constructor of `DatasetNode` when calculating subdataset depth.

When a dataset has a lot of subdatasets (no matter if installed or not), this adds up to a lot of redundant calls to `git submodules` (once for each subdataset, all trying to get the same parent ds)! So I extracted `ds.subdatasets()` in a separate function that caches results, so that we read a given parent's submodules only once. This means deep hierarchies will still be relatively slow, but at least we will see an improvement for large flatter hierarchies.

Unfortunately, to do that I had to reconstruct the `get_subdatasets()` function to extract out the call to `ds.subdatasets()`. Maybe there is a better way / some other utility that would do this?

https://github.com/catetrai/datalad-next/blob/nf-tree/datalad_next/tree.py#L337-L349


#### 2. A single helper tree instead of exponentially many

Before, I was creating a new secondary `Tree` for basically each directory, in order to decide whether the directory was a parent of a dataset and thus needed to be pruned or not in the main tree. This was extremely wasteful as I would be re-traversing the same nodes over again. Now I just instantiate a single helper/secondary `Tree`:

https://github.com/catetrai/datalad-next/blob/27450657e10e8b0fd8cc8e1beabf7f1ad10a810d/datalad_next/tree.py#L686-L701

There is definitely a speed improvement that in my POV goes from "I will never see the end of this, better Ctrl+C" to "at least it's running" (compare 7a74655d14f4f8ee89685682d35e33493b5d39a7 with 27450657e10e8b0fd8cc8e1beabf7f1ad10a810d on the `///` dataset). The implementation is also slightly more complex, but I think it's worth it. What do you say?

datalad_next/tree.py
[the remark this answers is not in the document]
mih: Nah, all good. I just wanted to mention it.

datalad_next/tree.py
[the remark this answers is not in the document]
mih: Yes, that is the only difference.

Re a use case of calling `ds.tree()` -- I think we are moving in that direction. I will comment on it, in the context of the result renderer.

datalad_next/tree.py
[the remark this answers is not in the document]
mih: Thanks, look great!

datalad_next/tree.py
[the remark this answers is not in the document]
mih: Thanks for making this change. It moves the implementation closer to the rest of the family.

Looking at the result records for a simple example (one superdataset with one immediate subdataset, and one immediate subdirectory) I see this:

```sh
% datalad -f json_pp tree
{
  "action": "tree",
  "depth": 0,
  "node_str": "[\u001b[1;37mDS~0\u001b[0m] \u001b[1;35m/tmp/mike\u001b[0m",
  "path": "/tmp/mike",
  "status": "ok",
  "tree_stats": "0 datasets, 0 directories, 0 files",
  "type": "dataset"
}
{
  "action": "tree",
  "depth": 1,
  "node_str": "\u251c\u2500\u2500 [\u001b[1;37mDS~1\u001b[0m] \u001b[1;35msub\u001b[0m/",
  "path": "/tmp/mike/sub",
  "status": "ok",
  "tree_stats": "1 dataset, 0 directories, 0 files",
  "type": "dataset"
}
{
  "action": "tree",
  "depth": 1,
  "node_str": "\u2514\u2500\u2500 \u001b[1;34msubdir\u001b[0m/",
  "path": "/tmp/mike/subdir",
  "status": "ok",
  "tree_stats": "1 dataset, 1 directory, 0 files",
  "type": "directory"
}
```

which is rendered to 

```sh
% datalad tree
[DS~0] /tmp/mike
├── [DS~1] sub/
└── subdir/

1 dataset, 1 directory, 0 files

tree(ok): /tmp/mike (dataset)
```

I think it would make sense to exercise the split of rendering and structured reporting all the way to the end:

The `node_str` properties already seems to be redundant with the result of the information in the record. The formating (adding the ASCII coloring, specific layout decision) could move entirely into the renderer. This would also improve the speed for consumers of the structured information, because the custom renders are not even called for them -- not that it is a lot of processing that is taking place here, but it would be cleaner.

The `tree_stats` property is pre-formated for console rendering, but it would be good to also have it be three properties with the actual integer values -- again to facilitate consuming this info in external tools.

These changes would move the implementation closer to addressing the usecase of walking and annotating a directory tree. I can imagine a a number of consumers for such a tool -- including, but not limited to a future DataLad-aware GUI!


Lastly, I noticed that the stats include `0 files` unless I run the command with `--include-files`. I think it would be better (re principle-of-least-surprise) to not report on number of files when it is unknown, rather than suggesting that there are no files.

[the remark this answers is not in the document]
mih: Oh, I forgot a minor point: I think it is unnecessary to have the generic renderer produce the line

```
tree(ok): <path> (<type>)
```

If people want that, they can run `datalad -f generic tree`

mih:
The speed improvements are noticeable, indeed! Awesome!

I also have a small patch suggestion to make things faster even when there is little to do:

```diff
diff --git a/datalad_next/tree.py b/datalad_next/tree.py
index 357131a..bf098c0 100644
--- a/datalad_next/tree.py
+++ b/datalad_next/tree.py
@@ -29,6 +29,7 @@ from datalad.interface.results import (
 from datalad.interface.utils import (
     eval_results, generic_result_renderer,
 )
+from datalad.local.subdatasets import Subdatasets
 from datalad.support.constraints import (
     EnsureNone,
     EnsureStr, EnsureInt, EnsureRange,
@@ -294,7 +295,8 @@ def get_subds_paths(ds_path: Path):
     def res_filter(res):
         return res.get('status') == 'ok' and res.get('type') == 'dataset'
 
-    return Dataset(ds_path).subdatasets(
+    return Subdatasets.__call__(
+        dataset=ds_path,
         recursive=False,
         result_filter=res_filter,
         on_failure='ignore',
```

This avoids an expensive import of the full datalad API that happens when calling dataset methods. You can do some poor-man's profiling with

```
DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=debug datalad tree
```

To see the message

```
Importing datalad.api to possibly discover possibly not yet bound method 'subdatasets'
```

which on my system with several extension packages installed leads to a >100ms detour. Here is the performance on the minimal (one super, one sub, one subdir) example from elsewhere in this PR:

Before

```
1: datalad tree
real        0.754       0.023       0.730       0.741       0.797       
user        0.643       0.020       0.606       0.646       0.680       
sys         0.122       0.014       0.094       0.121       0.145       
```

With patch

```
1: datalad tree
real        0.645       0.060       0.570       0.647       0.755       
user        0.545       0.036       0.509       0.536       0.615       
sys         0.112       0.026       0.067       0.114       0.146       
```

mih:
> Unfortunately, to do that I had to reconstruct the get_subdatasets() function to extract out the call to ds.subdatasets(). Maybe there is a better way / some other utility that would do this?

I need to think about this, and I will. I am heading out for some offline-time, though. It might take me till end of next week to actually get to it.

datalad_next/tree.py
```diff
@@ -0,0 +1,972 @@
...
+                path=node.path,
+                type=node.TYPE,
+                depth=node.depth,
+                node_str=line,
+                tree_stats=tree.stats()
+            )
+
+    @staticmethod
+    def custom_result_renderer(res, **kwargs):
+        print(res["node_str"])
```
mih, line 169: Here and elsewhere, it would be better to use

```py
from datalad.ui import ui
ui.message(...)
```

This takes care of not interfering with other datalad-messaging (e.g. progress bars), and it is also to some degree platform-aware -- it might already help with the windows CI issues that are still present right now.

mih:
I did some further testing, running it on large directory hierarchies, pretty much as a replacement of the regular `tree`.

I found that the node traversal needs some additional protection against file system issues. Here are a few examples:

```py
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 474, in _generate_tree_nodes
    yield DirectoryOrDatasetNode(
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 969, in __new__
    if is_dataset(path):
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 268, in is_dataset
    if Path(path / ".datalad" / "config").is_file():
  File "/usr/lib/python3.9/pathlib.py", line 1456, in is_file
    return S_ISREG(self.stat().st_mode)
  File "/usr/lib/python3.9/pathlib.py", line 1232, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/usr/lib/ssl/private/.datalad/config'


  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 474, in _generate_tree_nodes
    yield DirectoryOrDatasetNode(
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 969, in __new__
    if is_dataset(path):
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 278, in is_dataset
    if is_empty_dir():
  File "/home/mih/hacking/datalad/next/datalad_next/tree.py", line 276, in is_empty_dir
    return not any(path.iterdir())
  File "/usr/lib/python3.9/pathlib.py", line 1160, in iterdir
    for name in self._accessor.listdir(self):
OSError: [Errno 40] Too many levels of symbolic links: "/home/mih/PlayOnLinux's virtual drives/....
```

It is probably best to catch any and all of these issues at a central place and report them by yielding custom result records with `status='impossible'` or `status='error'`, and let the result renderer complain appropriately.

Right now the first occurrence of such an issue crashes the command.

mih:
Here are some performance comparisons. I do them in ipython after importing datalad, to avoid the penality of the datalad startup cost (which are unrelated to this command implementation):

```py
>>> from datalad.api import tree
>>> pwd
'/opt'

>>> timeit tree()
...
0 datasets, 114 directories, 0 files

14.4 ms ± 371 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

>>> timeit !tree -d
114 directories

6.53 ms ± 302 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

50% of the speed of `tree` is pretty good IMHO!

Now with files

```
>>> timeit tree(include_files=True)
...
0 datasets, 114 directories, 1347 files

46.8 ms ± 2.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

>>> timeit !tree
...
114 directories, 1347 files

12.7 ms ± 751 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

With files the performance drops to 25%.

mih:
Another random thought: Maybe we can consider embracing UTF icons as an alternative type maker. Maybe https://www.utf8icons.com/character/9921/white-draughts-king, instead of `DS`.

datalad_next/tree.py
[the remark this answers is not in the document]
catetrai: > These changes would move the implementation closer to addressing the usecase of walking and annotating a directory tree. I can imagine a a number of consumers for such a tool -- including, but not limited to a future DataLad-aware GUI!

That sounds like a great idea and something lots of currently-on-the-fence potential users would welcome, I'm sure -- just like git-GUIs (both desktop clients and Git[Hub,Lab] and the like) have lowered the barrier of entry for git.

The initial design was guided by the assumption that the tree display is all there is to the command. But you are right that this need not be the case. And if we offload that to the custom renderer, there is a much cleaner separation between node generation and output formatting, which improves the code in any case.

The results dict then becomes the crucial interface, so it needs to contain all the information that may be needed by consumers. A few questions on this:

https://github.com/datalad/datalad-next/blob/9d074fd0dcda12096e767d1f8f25e3d80bcdf142/datalad_next/tree.py#L157-L175

- Are there restrictions/conventions on the data types of the dict values? I tried to include only built-in types that are convertible to JSON (for the JSON renderers), such as `path` as string instead of a `Path` object.
- What about nested dicts (like the `count` attribute)? Based on a brief check through the other datalad commands, the general rule seems to be to keep the dict flat and minimal. In this case though, we need to pass all the information needed for the custom renderer (basically 'serializing' the `Tree` and `_TreeNode` (sub)classes), which adds up a number of attributes. Do you think there is a more appropriate pattern here that we could use?
- Is it okay to include an attribute conditionally, like the dataset-related attributes only for dataset nodes? Or should there be a fixed set of attributes, with value `None` where not applicable? (Again, this relates to the expectations on the side of the consumers).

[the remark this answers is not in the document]
catetrai: > Lastly, I noticed that the stats include 0 files unless I run the command with --include-files. I think it would be better (re principle-of-least-surprise) to not report on number of files when it is unknown, rather than suggesting that there are no files.

Done. Indeed, that is in line with the behaviour of `tree` when using the `-d` switch for directories-only.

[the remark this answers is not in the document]
catetrai: > Oh, I forgot a minor point: I think it is unnecessary to have the generic renderer produce the line
> 
> ```
> tree(ok): <path> (<type>)
> ```
> 
> If people want that, they can run `datalad -f generic tree`

Agree. Done.

catetrai:
> This avoids an expensive import of the full datalad API that happens when calling dataset methods. You can do some poor-man's profiling with
>
> [...]
> 
> which on my system with several extension packages installed leads to a >100ms detour. Here is the performance on the minimal (one super, one sub, one subdir) example from elsewhere in this PR:

Thanks for the debugging tips!

Interesting, I had not considered the costs of import for different modules.

I must have a mini-installation, it took a negligible amount of time on my system:

```shell
➜ DATALAD_LOG_TIMESTAMP=1 DATALAD_LOG_LEVEL=debug datalad tree --dataset-depth 1 .. 2>&1 | grep -A 1 Importing

2022-08-01 19:48:41,893 [DEBUG] Importing datalad.api to possibly discover possibly not yet bound method 'subdatasets' 
2022-08-01 19:48:41,896 [DEBUG] Building doc for <class 'datalad.core.local.create.Create'> 
```

Still, it's a quick and easy win, so I applied the patch.

catetrai:
> It is probably best to catch any and all of these issues at a central place and report them by yielding custom result records with `status='impossible'` or `status='error'`, and let the result renderer complain appropriately.

Yes, you are absolutely right. I have neglected the error handling aspect until now. I wonder how the exceptions should be reflected in the custom renderer, depending on whether a single (file) node is affected or a whole subtree, or some calculation of dataset properties. There may be recoverable and non-recoverable errors. I will try to add some tests for the permission denied and self-referencing symlink situations and come up with a first draft.

catetrai:
> Another random thought: Maybe we can consider embracing UTF icons as an alternative type maker. Maybe https://www.utf8icons.com/character/9921/white-draughts-king, instead of DS.

That would be neat, I had considered it too for the [starship prompt config](https://github.com/catetrai/starship-prompt-datalad) but did not find an icon that was distinctive enough. There's 📂 and even 🗄️, but no database icon -- though draughts would make for a decent imitation!

On my browser/terminal the symbol you linked is not rendered (it shows as '?'). Maybe stick with emojis as they are more likely cross-platform?

datalad_next/tree.py
catetrai: Done. And indeed, it fixed the encoding error on the Windows tests. Awesome!

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+                message = "error opening dir" \
+                    if node.exception.name == "PermissionError" \
+                    else node.exception.message
+
+                yield get_status_dict(
+                    status="error",
+                    message=message,
+                    exception=node.exception,
+                    **res_dict
+                )
```
catetrai, line 207: RE: error handling: I added an attribute `exception` on the `_TreeNode` to store any exception (wrapped in a `CapturedException`) that may have occurred when creating the object, e.g. `PermissionError` when trying to access a directory. The goal is to display the error and continue yielding the rest of the tree without crashing.

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+
+        # set suffix for directories
+        dir_suffix = ""
+        if depth > 0 and node_type in ("directory", "dataset"):
+            dir_suffix = "/"
+
+        # append symlink target if symlink
+        symlink_target = ""
+        if "symlink_target" in res:
+            symlink_target = " -> " + res["symlink_target"]
```
catetrai, line 286: Added rendering of symlinks with typical UNIX formatting `<link> -> <target>`, as in `tree` and `ls`.

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+        i.e. are also located under the tree root or any parent of
+        the tree root (within a distance of ``max_depth``).
+
+        Otherwise, the same subtree could be generated multiple times in
+        different places, potentially in a recursive loop (e.g. if the
+        symlink points to its parent).
+
+        This is similar to the logic of the UNIX 'tree' command, but goes a
+        step further to prune all duplicate subtrees.
+        """
```
catetrai, line 652: When trying out failure scenarios (such as your `[Errno 40] Too many levels of symbolic links` example), I came across the case of a symlink pointing to a directory within the tree. This caused the recursive generation of the same subtree for 33 (!) times (I think that's the value of `MAXSYMLINKS` [on MacOS](https://github.com/apple-oss-distributions/xnu/blob/xnu-4903.270.47/bsd/sys/param.h#L207)). While it may be an edge case, `tree` can handle it, so we should too. The goal is to not generate any duplicate nodes.

See tests: https://github.com/datalad/datalad-next/blob/de16e02068ce565cfbf697ca86d1e7ce95233eef/datalad_next/tests/test_tree.py#L753-L763

mih: Strictly speaking, a symlink is a file (even when it points to a directory). In terms of datalad result records, I'd give such a node a `type="symlink"` (neither file nor directory). In terms of counting for the sake of summary rendering I have no preference, as long as the structured records are precise. For that, it may make sense to record the fact that a symlink is pointing to a directory. I have currently no splendid idea how that should be done. I will keep thinking and return here.

datalad_next/tree.py
mih: Preserving the exception via `CapturedException` is exactly how it should be done. We have some convention how to label them in a result record. I will cross-reference that, when going through the code again.

mih: When testing 26e2f178704634a1ae1c2f58d02712940879e76e I found a condition that still causes `tree` to crash.

```sh
$ tree cups/ssl
cups/ssl
├── server.crt -> /etc/ssl/certs/ssl-cert-snakeoil.pem
└── server.key -> /etc/ssl/private/ssl-cert-snakeoil.key

0 directories, 2 files
```

vs.

```sh
$ datalad tree cups/ssl
/<abspath cut>/cups/ssl
[ERROR  ] [Errno 13] Permission denied: '/<abspath cut>/cups/ssl/server.key' (PermissionError) 
```

with

```sh
% ls -la cups/ssl/server.key
lrwxrwxrwx 1 mih mih 38 Dec 25  2012 cups/ssl/server.key -> /etc/ssl/private/ssl-cert-snakeoil.key
```

where the target exists, but `/etc/ssl/private` is not readable:

```sh
% ls -l /etc/ssl
total 36
drwxr-xr-x 3 root root     16384 Jun 27 11:16 certs/
-rw-r--r-- 1 root root     12441 Jun  4 15:25 openssl.cnf
drwx--x--- 2 root ssl-cert  4096 Jun  5 10:27 private/
```

Sidenote: When the `path` argument given to `tree` is relative, the command renderer would start with an absolute path as the tree root, while the POSIX `tree` would output a relative path.

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+        if is_dataset(potential_ds_root):
+            return potential_ds_root  # it's a match
+
+        # we go one directory higher and try again
+        ds_root = Path.resolve(potential_ds_root / '..')
+    return ds_root
+
+
+@lru_cache
+def get_superdataset(path: Path):
```
mih, line 512: I think it would be worthwhile to develop this function to be a full replacement of `Dataset.get_superdataset()` and patch the latter away with the former. It would need to understand `topmost=`, `datalad|registered_only=` as additional parameters to make this happen.

This does not have to be done in this PR, but if not, should be acknowledged in a issue, such that we do not accumulate alternative definitions of what is a (super)dataset and what is not.

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+        Example line:
+        ``│   │   ├── path_dir_level3``
+        """
+        from datalad.support import ansi_colors
+
+        # get values from result record
+        node_type = res["type"]
+        node_path = res["path"]
+        depth = res["depth"]
+        exhausted_levels = res["exhausted_levels"]
```
mih, line 229: Here is an example result record from a deep tree, but for toplevel-ish directory:

```json
{
  "action": "tree",
  "count": {
    "datasets": 5,
    "directories": 9469
  },
  "dataset_abs_depth": 0,
  "dataset_depth": 0,
  "dataset_is_installed": true,
  "depth": 1,
  "exhausted_levels": [
    2,
    7,
    9,
    10,
    11,
    12,
    13,
    8,
    5,
    3,
    6,
    4
  ],
  "path": "/home/mih/dcm",
  "status": "ok",
  "type": "dataset"
}
```

If I get it right, the `exhausted_levels` are reported to understand for which branches of the tree to no longer draw `| ` prefix markers.

But in the case of a `depth=1` record, it should not be necessary to report any `exhausted_levels`, because they should all be exausted. In general, any report that is larger than the present `depth` is unnecessary.

As a "cosmetic" fix, we could do something like

```suggestion
        exhausted_levels = [l for l in res["exhausted_levels"] if l < depth]
```

but I am wondering whether this set should not be pruned in the tree node generator itself?

datalad_next/tree.py
```diff
@@ -0,0 +1,1130 @@
...
+        )
+
+        for node in tree.generate_nodes():
+            # yield one node at a time to improve UX / perceived speed
+            res_dict = {
+                "action": "tree",
+                "path": str(node.path),
+                "type": node.TYPE,
+                "depth": node.depth,
+                "exhausted_levels": list(tree.exhausted_levels),
```
mih, line 170: Here is an example result record from a deep tree, but for toplevel-ish directory:

```json
{
  "action": "tree",
  "count": {
    "datasets": 5,
    "directories": 9469
  },
  "dataset_abs_depth": 0,
  "dataset_depth": 0,
  "dataset_is_installed": true,
  "depth": 1,
  "exhausted_levels": [
    2,
    7,
    9,
    10,
    11,
    12,
    13,
    8,
    5,
    3,
    6,
    4
  ],
  "path": "/home/mih/dcm",
  "status": "ok",
  "type": "dataset"
}
```

If I get it right, the `exhausted_levels` are reported to understand for which branches of the tree to no longer draw `| ` prefix markers.

But in the case of a `depth=1` record, it should not be necessary to report any `exhausted_levels`, because they should all be exausted. In general, any report that is larger than the present `depth` is unnecessary.

As a "cosmetic" fix, we could do something like

```suggestion
                "exhausted_levels": [l for l in tree.exhausted_levels if l < node.depth]
```

but I am wondering whether this set should not be pruned in the tree node generator itself?

mih:
Here are some profiling observations. I ran `datalad tree /usr`. It took 11.243s.

7.6s or 72% of the time is taken to generate the tree nodes. 24% is for visualization with the result renderer. Of these 7.6s,

- 10% in `path_depth()`
- 20% in `is_dataset()`
- 27% in `is_excluded()`

These may be pieces to optimize.

For result rendering, 62% of the time is spend in `ui.message()` (no surprise here). However, 85% of that time goes into `log_progress()` -- this is a lot. It is actually 12% of the total runtime of the entire command. This should become an issue in datalad core. I suspect it is caused by the messages used to turn the progress bars off and on again, when a message needs to be put out. This is useful in general, but needs to get optimized. In this case, there is no progress bar active at all, and the penalty is rather high. Maybe `ui.message()` could inspect the progress machinery directly, rather than going through the logging system unconditionally. Note to self: Need to investigate further.

datalad_next/tree.py
catetrai: > When testing https://github.com/datalad/datalad-next/commit/26e2f178704634a1ae1c2f58d02712940879e76e I found a condition that still causes tree to crash.

Good catch! I added this case to the test for broken symlinks. (I assume that a symlink to a file in an inaccessible directory is considered broken, since we cannot verify the existence of its target. `tree` also displays it as broken.)

The test passes now, but checks for permission errors are still not robust enough. Every `is_dir()` call is vulnerable. I will try to centralize error handling in the node constructor so I can set the `exception` property right away. This means instantiating nodes immediately, even if they will not be yielded in the end (hoping it does not impose too much of a performance cost). This may make other components cleaner as well, which currently operate only on paths and need to duplicate some of the node-creation logic.

catetrai: > Sidenote: When the path argument given to tree is relative, the command renderer would start with an absolute path as the tree root, while the POSIX tree would output a relative path.

This is a deliberate deviation from `tree` (see tests `test_root_path_is_normalized` and `test_no_difference_if_root_path_absolute_or_relative`). I always want to see the root printed as full path, without having to prefix it with `$PWD`. Other than for consistency with `tree` (and `ls`, `find`, ...), I don't see a reason to use relative paths. Maybe they help with readability if the root path is very long. This is my personal preference though. What's your take?

datalad_next/tree.py
catetrai: For the purpose of tree display, I think the link's target should dictate the node type and summary count. While the symlink itself is always technically a file, we follow symlinks when traversing the tree, so it makes sense to count links pointing to directories as directories (`tree` does this too, even without the `-l` option to follow links).

However for the results record, you are right that we may want the node type (unless it's a dataset) to match the actual file type -- what you'd get from `ls` or `stat -f "%T"` in Linux. Then for a symlink pointing to a directory we'd have `type=symlink, is_directory=True`. This is not too different from the current `type=directory, is_symlink=True`... but has the advantage of aligning with the file type model, so could make it more intuitive to use for consumers.

datalad_next/tree.py
catetrai: Yes, that bothered me too! `exhausted_levels` carried around the entire history of the Tree. It should be scoped to the current node instead.

Fixed it in the node generator:
https://github.com/datalad/datalad-next/blob/6db0e12fdef117c53f4dd7c84fa455e7e740596a/datalad_next/tree.py#L728-L733

Added test:
https://github.com/datalad/datalad-next/blob/6db0e12fdef117c53f4dd7c84fa455e7e740596a/datalad_next/tests/test_tree.py#L528-L531

datalad_next/tree.py
mih: That works for me. Thanks for the explanation!

mih:
As of https://github.com/datalad/datalad-next/pull/92/commits/6db0e12fdef117c53f4dd7c84fa455e7e740596a the implementation is now capable of handling my extraordinarily twisted and obscure home directory. Awesome!

```
259 datasets, 173209 directories
datalad tree /home/mih  133.17s user 29.30s system 85% cpu 3:10.20 total
```

I spotted a bunch of pieces where it took a long time to get to the next output line (sometimes 10+ seconds), I will investigate what is causing the slowness in these spots.

datalad_next/tree.py
mih: https://github.com/datalad/datalad/issues/6938

datalad_next/tree.py
mih: Nice! Thx!

catetrai:
Interesting story that of the subdatasets!

> I spotted a bunch of pieces where it took a long time to get to the next output line (sometimes 10+ seconds), I will investigate what is causing the slowness in these spots.

Just to be sure, did this happen with `--dataset-depth` or `--depth` or no depth parameter at all? A noticeable slowdown on single nodes I would expect from `DatasetTree` (when using `--dataset-depth`) because of the full-space search for datasets. But for regular tree generation it seems unusual. I am using `///` for benchmarking, will try to find a similar example.

catetrai:
Btw, Appveyor build on Ubuntu and MacOS is failing. Seems to be a problem with loading the command interface. I can run my tests locally just fine though. Maybe Python version issue?

mih:
With https://github.com/datalad/datalad/pull/6942 the runtime of `datalad tree` on my homedir is more than 3x faster

```
259 datasets, 173211 directories
datalad tree  35.40s user 8.60s system 100% cpu 43.875 total
```

down from 

```
259 datasets, 173211 directories
datalad tree  125.53s user 19.71s system 100% cpu 2:24.58 total
```

mih:
Looking at the tests, it would be nice to make them as silent as possible for all pieces that are just utilities. The rest of datalad is not yet very good at that, but I am slowly making my way through the codebase. Here is a sketch of what I mean:

```diff
diff --git a/datalad_next/tests/test_tree.py b/datalad_next/tests/test_tree.py
index 617a72a..1cdee4d 100644
--- a/datalad_next/tests/test_tree.py
+++ b/datalad_next/tests/test_tree.py
@@ -34,6 +34,9 @@ from ..tree import (
 
 # ============================ Helper functions ===============================
 
+ckwa = dict(result_renderer='disabled')
+
+
 @contextmanager
 def ensure_no_permissions(path: Path):
     """Remove all permissions for given file/directory and restore the
@@ -177,17 +180,17 @@ def path_ds():
     # create datasets / repos
     root = temp_dir_root / "root"
     BasicGitTestRepo(path=root / "repo0", puke_if_exists=False)
-    superds0 = Dataset(root / "superds0").create(force=True)
-    sd0_subds0 = superds0.create("sd0_subds0", force=True)
-    sd0_subds0.create("sd0_sub0_subds0", force=True)
-    superds1 = Dataset(root / "superds1").create(force=True)
-    superds1.create(Path("sd1_dir0") / "sd1_d0_subds0", force=True)
-    Dataset(root / "superds1" / "sd1_ds0").create(force=True)
+    superds0 = Dataset(root / "superds0").create(force=True, **ckwa)
+    sd0_subds0 = superds0.create("sd0_subds0", force=True, **ckwa)
+    sd0_subds0.create("sd0_sub0_subds0", force=True, **ckwa)
+    superds1 = Dataset(root / "superds1").create(force=True, **ckwa)
+    superds1.create(Path("sd1_dir0") / "sd1_d0_subds0", force=True, **ckwa)
+    Dataset(root / "superds1" / "sd1_ds0").create(force=True, **ckwa)
     BasicGitTestRepo(
         path=root / "superds1" / "sd1_dir0" / "sd1_d0_repo0",
         puke_if_exists=False)
-    sd1_subds0 = superds1.create("sd1_subds0", force=True)
-    sd1_subds0.drop(what='all', reckless='kill', recursive=True)
+    sd1_subds0 = superds1.create("sd1_subds0", force=True, **ckwa)
+    sd1_subds0.drop(what='all', reckless='kill', recursive=True, **ckwa)
 
     yield temp_dir_root
 
```

datalad_next/tree.py
```diff
@@ -0,0 +1,1172 @@
...
+        integer) or *above* the tree root (negative integer)
+    """
+    if is_path_relative_to(path, root):
+        return len(path.relative_to(root).parts)
+    elif is_path_relative_to(root, path):
+        return - len(root.relative_to(path).parts)
+    else:
+        raise ValueError("Could not calculate directory depth: "
+                         f"'{path}' is not relative to the tree root "
+                         f"'{root}' (or vice-versa)")
```
mih, line 405: Given the frequency with which this function is called, I made an attempt to make it faster. The use of `is_path_relative_to()` prior to `relative_to()` essentially implies that `relative_to()` is ran at least twice on the same path combo.

Here is my alternative implementation:

```suggestion
    sign = 1
    try:
        rpath = path.relative_to(root)
    except ValueError:
        try:
            rpath = root.relative_to(path)
            sign = -1
        except ValueError:
            raise ValueError(
                "Could not calculate directory depth: "
                f"'{path}' is not relative to the tree root "
                f"'{root}' (or vice-versa)")
    return sign * len(rpath.parts)
```

Impact estimate (via `datalad -f disabled --on-failure ignore tree <path>`):

```
# on my 23k-dir /usr with no datasets (with patch)
real        5.187       0.457       4.253       5.184       5.932       
# on my 23k-dir /usr with no datasets (without patch)
real        5.851       0.851       4.388       5.763       7.191       

# on dataset with 80k directories across two levels (with patch)
real        8.709       0.995       7.370       9.005       9.752       
# on dataset with 80k directories across two levels (without patch)
real        9.380       0.559       8.608       9.621       9.911       
```

Not breathtaking, but 5-10% faster.

datalad_next/tree.py
```diff
@@ -0,0 +1,1172 @@
...
+    try:
+        if path.is_symlink():
+            # ignore symlinks even if pointing to datasets, otherwise we may
+            # get duplicate counts of datasets
+            lgr.debug("Path is a symlink, will not check if it points to a "
+                      f"dataset: '{path}'")
+            return False
+
+        if (path / ".datalad" / "config").is_file() or \
+                (path / ".datalad" / "metadata").is_dir():
```
mih, line 449: We can simplify this to:

```suggestion
        if (path / ".datalad" / "config").is_file():
```

The metadata dir is/will be no longer used in future datasets. The original implementation still needs to be adjusted too. This change has no significant performance impact.

datalad_next/tree.py
```diff
@@ -0,0 +1,1172 @@
...
+
+
+def is_subds_of_parent(subds_path: Path, parent_path: Path):
+    return any(
+        is_path_relative_to(Path(p), subds_path)
+        for p in get_subds_paths(parent_path)
+    )
+
+
+def get_dataset_root_datalad_only(path: Path):
```
mih, line 504:
```suggestion
@lru_cache
def get_dataset_root_datalad_only(path: Path):
```

On my system and the 80k directory dataset, this gives 10-15% speedup and has no performance impact in for directories not contained in a dataset.

mih:
Thank you for bearing with me all this time. I left a bunch of additional comments and suggestions re performance. I think this is good and pretty much done.

I still need to take a look at the failing CI.

I believe the last aspect to talk about is the finalization of the API, including, but not limited to, short options.

I want to start with thinking about a conceptual separation of rendering related vs discovery related parameters. For example, `--dataset-depth` says `maximum level of nested subdatasets to display`. For any programmatic caller it would be good to communicate whether this (also) affects discovery. Most other datalad command have `--recursive` and `--recursion-limit` to tackle similar use cases. Do you think we can map that concept to `tree` too? Maybe for `tree` it does not make sense to have `--recursive` default to `False`, but otherwise it should match quite closely.

Same issue (discovery vs display) applies to `--include-files|hidden`, and also `--depth`.

datalad_next/tree.py
```diff
@@ -0,0 +1,1172 @@
...
+
+    @property
+    def tree_root(self) -> Path:
+        """Calculate tree root path from node path and depth"""
+        parents = self.parents
+        return parents[0] if parents \
+            else self.path  # we are the root
+
+    @property
+    def parents(self) -> list[Path]:
```
mih, line 954: > Btw, Appveyor build on Ubuntu and MacOS is failing. Seems to be a problem with loading the command interface. I can run my tests locally just fine though. Maybe Python version issue?

It is caused by the type annotation in Python 3.8:

```
$ python
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datalad_next.tree import TreeCommand
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/appveyor/projects/datalad-next/datalad_next/tree.py", line 919, in <module>
    class _TreeNode:
  File "/home/appveyor/projects/datalad-next/datalad_next/tree.py", line 954, in _TreeNode
    def parents(self) -> list[Path]:
TypeError: 'type' object is not subscriptable
>>> 

```

```suggestion
    # More accurate annotation only from PY3.9 onwards
    # def parents(self) -> list[Path]:
    def parents(self) -> list:
```

mih: ATTN: I will commit this change to see how the test behave afterwards.

mih:
I push small changes that hopefully sort out the test failures.

One works around PY3.7-incompatible type annotations. We need to decide whether to remove/simplify them prior merge, or bump the minimun supported version to be above the oldest maintained Python release.

catetrai:
> Looking at the tests, it would be nice to make them as silent as possible for all pieces that are just utilities. The rest of datalad is not yet very good at that, but I am slowly making my way through the codebase. Here is a sketch of what I mean:

Makes sense! Done in 7a66ba6.

datalad_next/tree.py
catetrai: Thank you for the proposed change! Looks much cleaner as well.

Indeed, `relative_to()` and its seemingly cheap `PurePath` string manipulation was eating up a huge chunk of runtime!

Got a significant speedup after cutting it out from several places (a551e25):

```shell
# dataset with 689 non-installed subdatasets

# before:
➜ python -m timeit -s 'from datalad_next.tree import TreeCommand; path="datasets.datalad.org/openneuro"' 'TreeCommand.__call__(path, dataset_depth=0, result_renderer="disabled")'
1 loop, best of 5: 7.37 sec per loop

# after:
➜ python -m timeit -s 'from datalad_next.tree import TreeCommand; path="datasets.datalad.org/openneuro"' 'TreeCommand.__call__(path, dataset_depth=0, result_renderer="disabled")'
1 loop, best of 5: 1.04 sec per loop
```

datalad_next/tree.py
catetrai: Good to know. Fixed in 070953dba.

datalad_next/tree.py
catetrai: Makes sense! Fixed in af014207.

datalad_next/tree.py
catetrai: Thank you for troubleshooting and pushing the changes!

datalad_next/tree.py
mih: Thank you for optimizing it further!

> Not sure what the use case could be for having datasets as subdirectories that are not submodules. Do you know if this is a common setup for any workflow?

`datalad-next` is providing similari-ish use case itself. The `datalad-annex::` git remote helper uses utility repositories underneath `.git` to function. Given that, I am not confident that there is no valid use case.

I will look into the `--dataset-depth` use case too.

catetrai: > datalad-next is providing similari-ish use case itself. The datalad-annex:: git remote helper uses utility repositories underneath .git to function. Given that, I am not confident that there is no valid use case.

That would already violate the current assumption that we need not look for datasets under the `.git` folder:

https://github.com/datalad/datalad-next/blob/070953dba2f7478c17c922a2d530e38512a6af48/datalad_next/tree.py#L853-L856

If these utility repos are purely used for internal machinery and are not generally of interest to the user, I would not mind skipping them. They could still be discovered in the 'full traversal' mode.

From a brief look at `datalad_annex.py` it seems like the utility repos are plain git(-annex) repos though, is that correct? In that case they would not be counted as datasets, since `tree` uses a narrow definition of dataset (only repos with a `.dataset/config` file -- even more accurate, but more expensive, would be to check for dataset ID). This is admittedly inconsistent with (most?) other datalad commands, which work on pure git repos too, and in general with the philosophy that a dataset is a repo. Thinking about it, we could provide an option `--repo-type=(datalad-only|all)` (default `datalad-only`) to detect all repos as datasets.

catetrai:
> I believe the last aspect to talk about is the finalization of the API, including, but not limited to, short options.

I updated the command docstring, parameters and examples to reflect the current implementation, so we can use this as a starting point for discussing changes:

https://github.com/datalad/datalad-next/blob/e32cf5843990769b49d0b1715cbe9c0fce82c4f5/datalad_next/tree.py#L51-L219

> Most other datalad command have --recursive and --recursion-limit to tackle similar use cases. Do you think we can map that concept to tree too? Maybe for tree it does not make sense to have --recursive default to False, but otherwise it should match quite closely.

Yes, for me `--recursion-limit` has similar semantics as `--dataset-depth`. With the difference that since `tree` does not operate on a dataset, the limit is applied to whatever (sub)datasets are found below the tree root. But I think that's close enough to still be intuitive to grasp if someone is familiar with `-r`/`-R` usage in other datalad commands. I propose we rename the `--dataset-depth` option to `--recursion-limit` (with short option `-R`). `--recursive` could be used for unconstrained search (which we actually don't have in the current implementation -- need to specify some arbitrary value like `--dataset-depth=100`). What do you say?

> I want to start with thinking about a conceptual separation of rendering related vs discovery related parameters.

I had previously considered distinguishing between two kinds of exclusion filters: 'display+traversal' exclusion filters (=directories will not be yielded nor traversed when searching for datasets) vs. 'display only' exclusion filters (=directories will not be yielded standalone, but may be traversed in the dataset search).

For example, with the `--include-hidden` option, hidden directories are not reported nor traversed in the regular tree mode. But in dataset search mode (if `--dataset-depth` is specified), then they will be traversed (and therefore reported) if they are the parent of valid dataset. So we could say that the option affects reporting but not discovery.

Is this the point you wanted to address? Or am I missing the mark?

mih:
Just briefly: the newly added CI runs verify compatibility with our oldest still supported Python version (3.7) and they document that there are still type annotations in this PR that break 3.7.

catetrai:
> Just briefly: the newly added CI runs verify compatibility with our oldest still supported Python version (3.7) and they document that there are still type annotations in this PR that break 3.7. Its estimated EOL is 2023-06 -- too long to be ignored IMHO.

Thanks for reminding. I got this mixed up with the type hints fix you pushed in e4d041d. This is a different issue, caused by the [no-parenthesis syntax](https://bugs.python.org/issue36772) of the `lru_cache` decorator. It is fixed in 8510dd6, waiting for build results now to confirm it. Long(ish) live 3.7!

(Sorry for the noise this has generated. Since this has happened a few times already, I have set up tox locally to run on all supported major versions, so these compatibility issues are (hopefully) caught by the tests before I push them and rightfully annoy the CI and maintainers ;) )

datalad_next/tree.py
mih: OK, first look into `--dataset-depth` using your recommended testcase `///` with `-r -R2`. Indeed, no speed-up from limiting the traversal:

```
% datalad -f disabled tree --dataset-depth 1
datalad -f disabled tree --dataset-depth 1  20.16s user 8.46s system 103% cpu 27.538 total
% datalad -f disabled tree
datalad -f disabled tree  18.86s user 6.91s system 102% cpu 25.152 total
```

(First call only lists 33 datasets, second 2k+).

mih: I have not yet fully grokked the thing with the utility tree used to discover the "next" dataset. At least for the main/outer tree, I am fairly confident that I saw it behaving in an optimal fashion. Not touching paths that should not be touched.

For the inner tree (in `_generate_datasets()`)  I am still uncertain what it should be doing. I have a hunch that it needs to become aware of the `dataset_depth` parameter in some way to become efficient. But I guess I now understand your statement above:

> And why do we need to traverse the whole tree? Because (super)datasets could technically be anywhere. I wonder if we could make the assumption, though, that no top-level datasets are located underneath other datasets (as subdatasets that are not registered in gitmodules). Not sure what the use case could be for having datasets as subdirectories that are not submodules.

Which brings me to ask the question "what is the use case?" I am OK with "look for any top-level dataset anywhere" to not be faster than "show me everything", because I think this is not a use case where people would not be OK to wait.

You wrote:

> For example, displaying just the 33 first-level subdatasets of /// (installed with -r -R 2, without data) takes 1m20s! And that should be the hello world of command usage, not an edge case ;)

True, but in this case there is a single top-level dataset, and `datalad  subdatasets -d mydl/ -r -R1` provides the desired result in 200ms. We have that command, and people can use it. Your implementation makes no assumption of a single root dataset. It need not compete with the simple case.

I would propose two things to progress here:

1. It would be good to add some hints on performance expectations to `*Dataset discovery*`. Pretty much what you wrote above (it needs to traverse everything, because it wants to do more than a `subdatasets` call). So people can use the faster one, if that is enough for them.
2. Given the observation that a call with `--dataset-depth` is not faster than one without, there may be a chance to reduce the complexity of the implementation by simply always producing the full tree and only filter the Node outcome at the very end, rather than nesting the `Tree` generators. But that is just a hunch.

Apart from that I made a few small changes to make logging a bit faster (by prevent string formating when logging is not actually active (minuscule impact, but still the right thing to do).

```diff
diff --git a/datalad_next/tree.py b/datalad_next/tree.py
index ad16251..92a95aa 100644
--- a/datalad_next/tree.py
+++ b/datalad_next/tree.py
@@ -540,7 +540,7 @@ def is_dataset(path: Path, installed_only=False):
             # ignore symlinks even if pointing to datasets, otherwise we may
             # get duplicate counts of datasets
             lgr.debug("Path is a symlink, will not check if it points to a "
-                      f"dataset: '{path}'")
+                      "dataset: %r", path)
             return False
 
         if (path / ".datalad" / "config").is_file():
@@ -748,15 +748,16 @@ class Tree:
                     current_node.is_recursive_symlink(self.max_depth):
                 # if symlink points to directory that we may visit or may
                 # have visited already, do not recurse into it
-                lgr.debug(f"Symlink is potentially recursive, "
-                          f"will not traverse target directory: '{dir_path}'")
+                lgr.debug("Symlink is potentially recursive, "
+                          "will not traverse target directory: %r", dir_path)
                 return
 
             if current_node.exception is not None:
                 # if some exception occurred when instantiating the node
                 # (missing permissions etc), do not recurse into directory
                 lgr.debug("Node has exception, will not traverse directory: "
-                          f"path={current_node.path}, exc={current_node.exception}")
+                          "path=%r, exc=%r",
+                          current_node.path, current_node.exception)
                 return
 
             # sort child nodes alphabetically
@@ -893,9 +894,9 @@ class DatasetTree(Tree):
 
             except Exception as ex:
                 CapturedException(ex, level=10)  # DEBUG level
-                lgr.debug(f"Excluding node from tree because "
+                lgr.debug("Excluding node from tree because "
                           "an exception occurred while applying the "
-                          f"exclusion filter: '{node.path}'")
+                          "exclusion filter: %r", node.path)
                 return True  # exclude by default
 
             return False  # do not exclude
@@ -914,10 +915,7 @@ class DatasetTree(Tree):
         """Go to the next dataset or parent of dataset"""
         self._next_ds = next(self._ds_generator, None)
         if self._next_ds is not None:
-            lgr.debug(
-                f"Next dataset" +
-                (" parent" if isinstance(self._next_ds, DirectoryNode) else "")
-                + f": {self._next_ds.path}")
+            lgr.debug("Next dataset (parent): %s", self._next_ds.path)
 
     def _generate_datasets(self):
         """Generator of dataset nodes and their parent directories starting
```

When I get the next chance, I plan to look into computing some of the more expensive Node properties lazily. I have no good concept of a potential impact yet, but even little pieces shaved off of the Node instantiation time will have a solid cumulative impact.

catetrai: Thinking it over, I agree with your points.

> 1. It would be good to add some hints on performance expectations to *Dataset discovery*. Pretty much what you wrote above (it needs to traverse everything, because it wants to do more than a subdatasets call). So people can use the faster one, if that is enough for them.

Makes sense, since performance will hit a ceiling no matter the implementation. Done in b3d1ac3.

> 2. Given the observation that a call with --dataset-depth is not faster than one without, there may be a chance to reduce the complexity of the implementation by simply always producing the full tree and only filter the Node outcome at the very end, rather than nesting the Tree generators. But that is just a hunch.

I think it was a good hunch! For perceived command speed, it does not make that much of a difference IMO if the dataset tree is computed piecewise 'during' the execution of the outer main tree, or in its entirety before starting the main tree generator. Finding all datasets upfront also allows us to set a max depth for the main tree, based on the deepest found dataset (plus ``max_depth`` dataset subnodes).

The change is done in 79732a7. Since this now implies a potentially long wait right after running the command, we may want to consider hooking in a progress bar that monitors the dataset search. Do you think that would help or is it overkill?

catetrai: > Apart from that I made a few small changes to make logging a bit faster (by prevent string formating when logging is not actually active (minuscule impact, but still the right thing to do).

Happy to be proven wrong, but I was under the impression that f-strings outperformed all other string formatting constructs:

```shell
➜ python -m timeit -n 1000000 -s 'wi="wi"' '"sand%sch" % wi'
1000000 loops, best of 5: 116 nsec per loop

➜ python -m timeit -n 1000000 -s 'wi="wi"' '"sand{}ch".format(wi)'
1000000 loops, best of 5: 179 nsec per loop

➜ python -m timeit -n 1000000 -s 'wi="wi"' 'f"sand{wi}ch"'
1000000 loops, best of 5: 68 nsec per loop
```

mih: f-strings are the fastest interpolation, indeed! But what my patch does is to do *no interpolation at all* -- which is even faster ;-)

Essentially, unless a certain debug level is actually active, no CPU cycles will be wasted on string formatting/interpolation. `logging` will apply `%` formatting only at the point when an emitted log actually hits a formatters and is not excluded by any filter (where the level could just be one of many filters).

catetrai: Oh, my bad! You are passing the to-be-injected values as arguments to the logger, not to the `%` operator. So it's not a formatted string at all, until (if) the logger decides to evaluate it. That is so cool, definitely the right thing! (Off to rescuing CPU cycles I have been wasting en masse now...)

mih (approved):
Alrighty! I think we had a stellar run and this came out beautiful. Thanks for having this kind of stamina!

From my POV we can merge this now without any further changes. We can add the short options now, or in a subsequent PR.

The performance is cool from my POV too -- I think we have more substantial margins for improvement in the rest of datalad, compared to what is done in this PR.

I need to set up the contributor acknowledgement framework we use in other extension too, to acknowledge you properly for this contribution -- I will get to that soonish!

Thanks much! I love it!

datalad_next/tree.py
mih: ;-) Thx!

catetrai:
It was my pleasure @mih! Thank you for the feedback and guidance throughout. Glad to have the chance to learn a bit about datalad internals, too (well, more like took a peek underneath the API). Debugging tips also invaluable for daily usage / life.

I will rename the options and add short variants as parts of this PR.

If you or the team have any improvement suggestions (also on documentation, code style / naming conventions, refactoring the tests a bit, etc) I'm happy to follow up on separate PRs.

mih:
Huh, one of the datalad-core tests started failing. It seems to be an issue with git-annex. I do not see an immediate connection. Will investigate....

mih:
Yeah, unrelated and should be gone with the 0.17.3 release that just came out.

mih:
I just saw that a changelog snippet was still missing, so I added one with `scriv create`.

mih (approved):
Good to go!

PR #93, opened 2022-08-12
Fix results when pushing to export remote

### LINKED ISSUE #88: push to export remote patch swallows errors
This `try ... except`: https://github.com/datalad/datalad-next/blob/864896e26af7361b739b05df4fcd302b0b407db3/datalad_next/patches/push_to_export_remote.py#L216

is wrong. If `annex export` fails we get an error result from `_call_annex_records_items_` rather than an exception. And this error result is unconditionally turned into an `status='ok'` record here.

bpoldrack:
The push-to-export-remote patch of core's push command failed to handle
the results from `AnnexRepo._call_annex_records_items_` correctly.
It was not accounting for the fact, that this are JSON records as
reported by git-annex (thus ignoring a bunch of fields), which should be
done by `datalad.interface.results.annexjson2result`.
Furthermore it was unconditionally setting the status to `ok` expecting
a `CommandError` otherwise. This is not how the generator works in the
general case, though. It does yield `error` records and proceeds if it
can. `CommandError` still can happen but it's not the only thing leading
to an error result.

This led to `push` always "succeeding" with an `ok` result, rendering
all tests of export remotes useless, that rely on `push` failing if
there was something wrong.

The respective tests made were based on the same, wrong assumptions
about `_call_annex_records_items_`.

This commit (hopefully) fixes this.

Fixes #88

Edit:
Additional note: This also fixes a bug where `push` to export fails due to `annex export` not reporting a `file` in its JSON response when renaming a file. W/O `annexjson2result` the patch directly accessed `result['file']` and crashed due it being `None`.

### DIFF
--- a/changelog.d/20220812_190404_benjaminpoldrack_fix_push_patch.md
+++ b/changelog.d/20220812_190404_benjaminpoldrack_fix_push_patch.md
@@ -1,0 +1,7 @@
+### 🐛 Bug Fixes
+
+- Fixed datalad-push always reporting success when pushing to
+  an export remote.
+  Fixes https://github.com/datalad/datalad-next/issues/88 via
+  https://github.com/datalad/datalad-next/pull/93 (by @bpoldrack)
+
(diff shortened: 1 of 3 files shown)


mih (approved):
Great catch! Thx!

PR #103, opened 2022-08-24
DOC: Add an example of adding a user_password credential

mslw:
This adds an example of how to set a new credential of type user_password, with a given user property, to the examples in the
credentials command. For some reason, I always had to think twice before adding a new user_password credentials based on the existing docs (maybe it's just me). I thought this usage could have an explicit example.

The added example shows that type and user need to be defined as properties, and entry of the secret is interactive. While it was probably clear from the other examples and command description, now it's explicit. Also, it is the first example where more than one property is being given.


mih (approved):
Thx!

PR #114, opened 2022-10-14
Report incomplete cedentials

mih:
Rational inside.


bpoldrack (approved):
Awesome. Works for me.

mih:
It seems that some other pieces rely on the non-reporting of incomplete credentials. The likely need to be updated to specifically look for a `secret` property in the reported value.

Issue #115, opened 2022-10-19
Develop replacement of `Constraint` system


mih:
### Development targets

- [ ] Assemble all code in a dedicated subpackage (`datalad(_next).constraints`), comprised of a module for the base and "meta" classes, a module for the primitives (constraints for basic data types), and a module for specialized ones (e.g. `EnsureDatasetSiblingName`)
- [ ] Extend base class with API for context switching. See what `-gooey` has done with `Constraint.for_dataset()`
- [ ] Support creation of a complex compound constraint from a json schema specification that is able to validate a complete form record. The is analog to what `-gooey` has been done for building a complete constraint for a parameter from a `-core` `Parameter` declaration 
     https://github.com/datalad/datalad-gooey/blob/c6d59c24f5104693e1e8183b596b41fb32f4fc1c/datalad_gooey/param_form_utils.py#L167-L247
     Ideally, both sets of functionality become accessible via static methods of the `Constraint` class, i.e. `from_parameter()`, `from_json_schema()` https://github.com/datalad/datalad-gooey/issues/393
- [ ] Address https://github.com/datalad/datalad-gooey/issues/311 and thereby have a constraint that can validation a complete command parameterization
- [ ] Implement a "context switching" `EnsureMapping()` variant that can take another `Constraint` like `EnsureMappings`, and internally switch its context, once a value of type `Dataset` is known to a declare mapping key. Such a constraints can then be used to perform a full context-aware validation of a command parametrization. The base command handling in `-core` could then be changed to perform any and all validation *before* a command is called. This would substantially slim implementations of particular commands, homogenize behavior and reduce boilerplate code. Moreover, any command could also declare such an additional, command-level constraint that perform inter-parameter validation, specific for a command (e.g. to ensure or prevent particular parameter (value) combination, e.g. mutual exclusion etc...
- [ ] Allow for CLI generation from a `Constraint`. `-gooey` can already auto-generate a GUI for a command based on a single Constraint (per Parameter, ATM, but see above). This shows that the same should be possible for the CLI, and thereby replace the near-duplicate handling in core's CLI module.
- [ ] Add documentation support. ATM the self-documentation possible in constraint types is rather minimal (`short|long_description()`), We should add support for declaring custom documentation via a constrain constructor. This would take and replace the role of `Parameter(doc=)`
[...]

PR #116, opened 2022-10-19
Replacement for DataLad's "constraints"

### LINKED ISSUE #121: EnsureFromStdin(constraint)
We would benefit from a meta-constraint that wraps another constraint and reads items from `stdin`, passing each item through the payload constraint.

This would be a useful paradigm to factor out any input logic from the body of a command. This is pretty much providing a "batch-mode" for any command that had previously taken an iterable argument of some kind.

An `EnsureFromFile` could work in a similar fashion.

mih:
Towards https://github.com/datalad/datalad-next/issues/115

- Imports all legacy code from datalad-core
- Imports the majority of improvements and additional constraint implementations from datalad-gooey
- Brings test coverage to near-100%
- removed `expand_constraint_spec()`: Now done by 
- `Constraint.for_dataset()` API extension
- New `NoConstraint` class
- New `EnsureStr` with regex matching
- `EnsureInt|Float` as specializations of the new `EnsureIterableOf`
- New `EnsurePath` with the ability to compare a path to a reference
- New `EnsureMapping`
- New `EnsureGitRefName`
- `EnsureRange` with improved descriptions
- Bugfixes in `EnsureDType`
- New `EnsureParameterConstraint`, constraint to validate a command parameter. Comes with `EnsureParameterConstraint.from_parameter()` that supersedes a custom and internal datalad-gooey helper
- Tests no longer use nose-times helpers
- Removed `expand_constraint_spec()`: Now done by `EnsureParameterConstraint.from_parameter()` -- and properly deprecated.
- New `EnsureValue`, generalized from `EnsureNone`
- New `EnsureGeneratorFromFileLike` to read iterables from any file-like (including STDIN).
- New `EnsureJSON` to decode serialized JSON string. Together with `EnsureGeneratorFromFileLike` this automated reading JSON-lines from any file-like of STDIN.

Closes #121


datalad_next/constraints/base.py
```diff
@@ -0,0 +1,164 @@
...
+from typing import (
+    TYPE_CHECKING,
+    TypeVar,
+)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from datalad.distribution.dataset import Dataset
+
+aConstraint = TypeVar('aConstraint', bound='Constraint')
+aDataset = TypeVar('aDataset', bound='Dataset')
```
christian-monch, line 13: Just a remark: I think the convention for type var naming is capitalized. Also: in  OO-languages "aX"  usually refers to an instance of "X". Although this is not PEP-8 style ---which would be "a_x"--- the notation might be confusing to some people, I don't have an obviously superior recommendation though. "DatasetType", "DatasetLike", ...?

Also, I am not sure whether the type variables are necessary here. Type-checking should work with subclasses.

datalad_next/constraints/base.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __call__(self, value):
+        e_list = []
+        for c in self.constraints:
+            try:
+                return c(value)
+            except Exception as e:
+                e_list.append(e)
+        raise ValueError(
+            f"{value!r} voilated all possible constraints {self.constraints}")
```
christian-monch, line 115:
```suggestion
            f"{value!r} violated all possible constraints {self.constraints}")
```

datalad_next/constraints/parameter.py
```diff
@@ -0,0 +1,180 @@
...
+)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from datalad.support.param import Parameter
+
+aEnsureParameterConstraint = TypeVar(
+    'aEnsureParameterConstraint',
+    bound='aEnsureParameterConstraint',
+)
+aParameter = TypeVar('aParameter', bound='Parameter')
```
christian-monch, line 30: See comment on "aDataset".

christian-monch (approved):
Great! Thx!

Generally looks good to me. Because of the size of the PR, I did not go into it in too much detail. Left a comment on the the TypeVars though, feel free to ignore them.

datalad_next/constraints/base.py
mih: Can you clarify please? `aDataset` is meant to refer to an instance.

The purpose of type vars here is to avoid the import of the underlying class. To avoid circularity (in case of `Constraint`) or to avoid making the constraint system depending on "all of datalad" (in case of `Dataset`).

mih: FTR: here is the PEP8 snippet https://peps.python.org/pep-0008/#type-variable-names

mih: I guess `DatasetDerived` would somehow reflect the semantics of `bound=`

mih:
Thx @christian-monch !

PR #135, opened 2022-11-18
Patch datalad.runner.runnerthreads.ReadThread to use shutil buffer size default

mih:
This has a dramatic effect on the throughput, i.e. when using `ThreadedRunner()` for SSH-based downloads.


christian-monch (dismissed):
LGTM

mih:
This does not work with PY3.7 (still supported). `shutil` did not have the info exposed back then. I will import the logic.

PR #136, opened 2022-11-18
Uniform command parameter validation

mih:
- Commands can now opt-in to receive fully validated parameters. This can substantially simplify the implementation complexity of a command at the expense of a more elaborate specification of the structural and semantic properties of the parameters. This specification is achieved by declaring `Constraints`, in a `_validators_` member of a commands `Interface` class.

  For now, a corresponding `Interface.validate_args()` classmethod must also be provided that performs the validation, and returns the validated outcomes. In the future a common implementation of this method will likely be provided that will define additional common conventions re parameter validation order, and accessibility of previous validation results to subsequent validation attempts of other parameters.

  This feature is introduced as a patch to the command execution in datalad-core. With this patch, commands are now exclusively called with keyword-style parameters only.

- A new `EnsureDataset` constraint is provided that returns a `DatasetParameter` on successful validation. This return value contains the original input specification, and the `Dataset` class instance. The `resolve_path()` utility is adjust to support this parameter-type, thereby making the use of the `require_dataset()` utility obsolete.


mih:
Only some of the changes are are covered by the tests. A subsequent PR brings them, using an actual command to fully exercise the possible complexity.

mih:
Appveyor does not show up, but is here https://ci.appveyor.com/project/mih/datalad-next/builds/45424289

mih:
image

mih:
OK, merging. The real thing is coming with `download()`.

PR #149, opened 2022-11-25
`UrlOperations` distinguish between connection-error and not-found

### LINKED ISSUE #148: `UrlOperations.sniff()` exceptions should distinguish connection error from access error
When a URL works in principle, but points to something that is just not there (e.g. HTTP404) vs we cannot tell right now (HTTP5xx, or network down).

This is needed for a reliable operation of CHECKPRESENT in git-annex special remotes, and for meaningful error reporting in general.

mih:
When a URL works in principle, but points to something that is just not there (e.g. HTTP404) vs we cannot tell right now (HTTP5xx, or network down). This now raises `AccessFailedError` for a general error and `UrlTargetNotFound` for a 404-like error.

This is needed for a reliable operation of CHECKPRESENT in git-annex special remotes, and for meaningful error reporting in general.

This changeset also include dedicated unit tests for `UrlOperations`.

Closes #148

### DIFF
--- a/datalad_next/url_operations/http.py
+++ b/datalad_next/url_operations/http.py
@@ -13,7 +13,11 @@ from requests_toolbelt.downloadutils.tee import tee as requests_tee
 import www_authenticate
 
 import datalad
-from datalad_next.exceptions import DownloadError
+from datalad_next.exceptions import (
+    AccessFailedError,
+    UrlTargetNotFound,
+    DownloadError,
+)
 
 from datalad_next.requests_auth import DataladAuth
 from . import UrlOperations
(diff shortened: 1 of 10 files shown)


datalad_next/url_operations/http.py
```diff
@@ -61,8 +78,13 @@ def sniff(self, url: str, *, credential: str = None) -> Dict:
             except requests.exceptions.RequestException as e:
                 # wrap this into the datalad-standard, but keep the
                 # original exception linked
-                raise AccessFailedError(
-                    msg=str(e), status=e.response.status_code) from e
+                if e.response.status_code == 404:
+                    # special case reporting for a 404
+                    raise UrlTargetNotFound(
+                        from_url, status=e.response.status_code) from e
```
mih, line 84:
```suggestion
                        url, status=e.response.status_code) from e
```

PR #152, opened 2022-11-28
'uncurl' special remote 

mih:
TODO

- [x] complete documentation
- [ ] allow additional identifiers to come from configuration. Recorded in https://github.com/datalad/datalad-next/issues/157 (not to be aimed for here)

### DIFF
--- a/datalad_next/annexremote/tests/test_uncurl.py
+++ b/datalad_next/annexremote/tests/test_uncurl.py
@@ -1,0 +1,269 @@
+from pathlib import Path
+import pytest
+import re
+
+from datalad_next.tests.utils import (
+    with_tempfile,
+    with_tree,
+)
+from datalad_next.constraints.dataset import EnsureDataset
+from datalad_next.exceptions import IncompleteResultsError
+from datalad_next.url_operations.any import AnyUrlOperations
+
+from ..uncurl import (
+    UncurlRemote,
+    UnsupportedRequest,
+)
+
+
+# for some tests below it is important that this base config contains no
+# url= or match= declaration (or any other tailoring to a specific use case)
+std_initargs = [
+    'type=external',
+    'externaltype=uncurl',
+    'encryption=none',
+]
+
+res_kwargs = dict(
+    result_renderer='disabled',
+)
+
+class NoOpAnnex:
+    def error(*args, **kwargs):
+        pass
+
+    def info(*args, **kwargs):
+        pass
+
+    def debug(*args, **kwargs):
+        pass
+
+
+def test_uncurl_store(tmp_path):
+    # not yet
+    r = UncurlRemote(NoOpAnnex())
+    with pytest.raises(UnsupportedRequest):
+        r.transfer_store(None, None)
+
+
+def test_uncurl_remove(tmp_path):
+    # not yet
+    r = UncurlRemote(NoOpAnnex())
+    with pytest.raises(UnsupportedRequest):
+        r.remove(None)
+
+
+def test_uncurl_claimurl(tmp_path):
+    r = UncurlRemote(NoOpAnnex())
+    # if we have a match expression defined, this determines claim or noclaim
+    r.match = [re.compile('bongo.*')]
+    assert r.claimurl('bongo://joe')
+    assert not r.claimurl('http://example.com')
+    r.match = None
+    # without a match expression, the url handler decides
+    r.url_handler = AnyUrlOperations()
+    for url in ('http://example.com',
+                'https://example.com',
+                'ssh://example.com',
+                'file:///home/me'):
+        assert r.claimurl(url)
+    assert not r.claimurl('bongo://joe')
+
+
+def test_uncurl_checkurl(tmp_path):
+    exists_path = tmp_path / 'testfile'
+    exists_path.write_text('123')
+    exists_url = exists_path.as_uri()
+    no_exists_url = (tmp_path / 'notestfile').as_uri()
+    # checkurl is basically an 'exists?' test against a URL.
+    # the catch is that this test is not performed against the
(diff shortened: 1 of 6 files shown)


datalad_next/annexremote/tests/test_uncurl.py
```diff
@@ -0,0 +1,269 @@
...
+    (tmp_path / 'ria').rename(tmp_path / 'ria_moved')
+
+    # verify that no residual magic makes data access possible
+    with pytest.raises(IncompleteResultsError):
+        ds.get(target_fname, **res_kwargs)
+
+    # fix it via an access URL config,
+    # point directly via a hard-coded dataset ID
+    # NOTE: last line is no f-string!
+    url_tmpl = f'{(tmp_path / "ria_moved").as_uri()}/{srcds.id[:3]}/{srcds.id[3:]}' \
```
bpoldrack, line 238: This looks a bit odd with f-string and template string mixture. How about this?

```suggestion
    url_tmpl = (tmp_path / "ria_moved" / srcds.id[:3] / srcds.id[3:]).as_uri() \
```

datalad_next/annexremote/tests/test_uncurl.py
```diff
@@ -0,0 +1,269 @@
...
+    # the original URL. All we need is to distinguish the old base path
+    # from the structured components in the RIA store (for the
+    # latter we can simple account for 4 levels of sudirs
+    ds.configuration(
+        'set',
+        'remote.myuncurl.uncurl-match='
+        'file://(?P<basepath>.*)/(?P<dsdir>[^/]+/[^/]+)/annex/objects/.*$',
+        **res_kwargs)
+    # NOTE: last line is no f-string!
+    url_tmpl = f'{(tmp_path / "ria_moved").as_uri()}' \
```
bpoldrack, line 261: Same here. I may miss something, but I think the f-string is superfluous.
```suggestion
    url_tmpl = (tmp_path / "ria_moved").as_uri()' \
```

datalad_next/annexremote/uncurl.py
```diff
@@ -0,0 +1,349 @@
...
+        # for the committed special remote config on the URL template
+        if not self.url_tmpl:
+            # ask for the commit config, could still be empty
+            self.url_tmpl = self.annex.getconfig('url')
+            # TODO test the case of a fully absent URL template
+            # that would be fine and only verbatim recorded URLs could
+            # be sent to a downloader
+
+        # unconditionally ask git-annex for a match-url setting, any local
+        # config ammends, and does not override
```
bpoldrack, line 118: If understand correctly the rational here is, that while I may have a better idea about the to-be-formed URL locally, the same would not be true for "known" URL patterns - I may know additional ones, but that doesn't invalidate already known ones. Hence, no override but only amendment. 

However, this needs clear documentation since the difference in behavior WRT local configs can be quite surprising. (All the more, if my understanding of the rationale was not correct)

datalad_next/annexremote/uncurl.py
```diff
@@ -0,0 +1,470 @@
...
+    re-use, or 'skip' to not save the credential
+    name: httpbin-dummy
+
+    addurl http://httpbin.org/basic-auth/myuser/mypassword (from uncurl) (to ...) 
+    ok
+    (recording state in git...)
+
+By adding files via downloads from URLs in this fashion, datasets can be built
+that track information across a range of locations/services, using a possibly
+heterogeneous set of access methods.
```
yarikoptic, line 58: overall it feels like a replacement/next generation of the `datalad` annex special remote in datalad "core" since it works exactly as `datalad` one just uses the credentials/"downloader" of -next instead of core. I think it might be worth making it explicit.

mih: Thanks for the feedback. If that is true, it likely matches the behavior without url templates and match expressions configured.

Declaring feature parity would then depend on the capabilities of the `UrlOperations` in -next vs the downloaders in core. I think there is a substantial overlap, but I am not confident that it is "same". So such a declaration would need to be made with care.

That being said, feature parity is certainly a goal here (within reason).

mih:
Failing test is due to the fact that I forgot that the RIA toolchain does not work on windows at all. Will disable that test.

datalad_next/annexremote/uncurl.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def initremote(self):
+        # at present there is nothing that needs to be done on init/enable.
+        # the remote is designed to work without any specific setup too
+        pass
+
+    def prepare(self):
+        # we need the git remote name to be able to look up config about
+        # that remote
+        remotename = self.annex.getgitremotename()
```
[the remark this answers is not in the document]
mih: Correct, thx! I will add that to the documentation as a minimum requirement. It was introduced in 8.20210127. Proper functioning of next already requires a much later version, due to numerous fixed (although most stuff can work with datalad's declared minimum version of 8.20200309.

datalad_next/annexremote/uncurl.py
mih: I added a statement to the documentation.

datalad_next/annexremote/uncurl.py
mih: I have added a section on configuration to the docs.

datalad_next/annexremote/tests/test_uncurl.py
mih: Thx, I removed it. I could not commit the suggested change directly, due to the syntax error.

datalad_next/annexremote/tests/test_uncurl.py
mih: I included a change in that spirit into an update. Thx.

PR #159, opened 2022-12-02
STORE/upload support for uncurl

mih:
This works in general, but `test_uncurl_store_via_ssh` shows a strange breakage pattern for my locally. Upload works, download works, CHECKPRESENT works, but `fsck` (which is CHECKPRESENT, plus download) does *not* work, and is only broken for SSH-only.

Git annex chatter for this looks like this:

```
...
[2022-12-02 06:01:03.35514891] (Annex.ExternalAddonProcess) /home/mih/env/datalad-dev/bin/git-annex-remote-uncurl[1] --> CHECKPRESENT-SUCCESS MD5E-s9--7291db66cf824b80413cfd5b76928997.txt

[2022-12-02 06:01:03.355586187] (Annex.ExternalAddonProcess) /home/mih/env/datalad-dev/bin/git-annex-remote-uncurl[1] <-- TRANSFER RETRIEVE MD5E-s9--7291db66cf824b80413cfd5b76928997.txt .git/annex/tmp/MD5E-s9--7291db66cf824b80413cfd5b76928997.txt

  external special remote protocol error, unexpectedly received "<EOF>" (unable to parse command)
```

So it succeed "sniffing" the key location via SSH, but then the special remote process closes its stdout(?), or maybe dies entirely?

When I switch the test `test_uncurl_store_via_ssh` from a `ssh://` to a `file://` target URL, everything works.

@christian-monch can you have a look at that please?

PR #166, opened 2022-12-06
fix a bug where ssh subprocesses steal stdin

christian-monch:
This PR fixes a bug that was triggered when an annexremote process uses ssh to communicate with a remote.

The issue was the the annexremote process would not receive all data that was sent over stdin. The reason for that was that the ssh-subprocesses inherit stdin from the parant process, i.e. from annexremote, and suck in data before the annexremote can read it. This is due to a `None`-argument in the subprocess.Popen stdin keyword argument. That leads to the subprocess inheriting stdin from the parent process. Instead `subprocess.DEVNULL`, would be a proper argument.

This should fix the error discussed in PR #159 and also appears in PR #164

### DIFF
--- a/datalad_next/url_operations/ssh.py
+++ b/datalad_next/url_operations/ssh.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import logging
+import subprocess
 import sys
 from itertools import chain
 from pathlib import (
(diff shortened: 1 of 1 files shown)


mih:
Ahhhh! That was it!! Thanks much!

christian-monch:
> Excellent, thx!

NP

PR #168, opened 2022-12-07
Remove `require_dataset()` usage from codebase

### LINKED ISSUE #139: Strip `require_dataset()` use from the code base
The mainline now has `EnsureDataset` which returns `DatasetParameter` that works seamlessly with `resolve_path()`. Ths combo should make any direct usage of `require_dataset()` obsolete.

Doing removal may need to go hand-in-hand with enabling uniform input parameter validation for the respective commands (see `download()` for a pattern, which could be generalized on this occasion).

mih:
as long as `require_dataset()` is still provided upstream), and in `datalad_next.patches` (for compatibility).

It shall never be introduced in other places again. `datalad_next.constraints.dataset.EnsureDataset` can and must be used now.

Closes #139

### DIFF
--- a/datalad_next/create_sibling_webdav.py
+++ b/datalad_next/create_sibling_webdav.py
@@ -25,8 +25,10 @@ from datalad_next.dataset import (
     datasetmethod,
 )
 from datalad.distribution.dataset import (
-    EnsureDataset,
-    require_dataset,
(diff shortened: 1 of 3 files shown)


mih:
Test failure is a transient 404 from github.

PR #169, opened 2022-12-07
Reorganize the code base

### LINKED ISSUE #160: Reorganize code base
Related to #145, but more general. In prepration for a switch to semantic versioning (#138 ), I think it is beneficial to restructure the source tree. For doing that, there are things to be learned from the core package, first and foremost that the notion of an "API" is too unspecific (there is a command API, a dataset API, a repo API, a utility collections, ... all of them are an API and all of them make up the API. Also labeling things in terms of their purpose (local, distributed, utility, ...) can be difficult. it is easier to label thing according to what their are (commands, remotes, backends). Lastly, the number of files in a directory is a lesser problem than the number of lines in a file.

With that, here is what I think:

```
datalad_next/
  annexbackend/
  annexremote/
  commands/
    __init__.py (with Interface)
    create_sibling_webdav.py
    credentials.py
    download.py
    tree.py
  constraints/
  dataset/
    __init__.py (with Dataset)
  exceptions/
  gitremote/
  patches/
  url_operations/
    any.py
    file.py
    http.py
    ssh.py
  utils/
    consts.py
    log.py
    http_helpers.py
    requests_auth.py
```

So plenty of top-level categorization, but no toplevel code files (other than structural necessities). This should avoid top-level waste accumulation, and simultaneously leave room for adding (self-contained) sub-packages (similar to `cli` or `runner` in core.

mih:
This is just moving code.

Closes #160


mih:
Looking good. Appveyor already passed previously

PR #173, opened 2022-12-08
Reestablish PY3.7 compat

mih:
`Path.unlink()` only has `missing_ok` since PY3.8. But there desired `missing_ok=False` is and was the default behavior. The underlying effects are tested, hence we can simply remove the argument.


mih:
CI failure is an unrelated timeout

Issue #175, opened 2022-12-09
Importless declaration of URL handlers


mih:
Right now, the dict dict based on which AnyUrlOperations selects a handler must contain a handler class as values. This makes it necessary to import everything first, and thereby creates an avoidable startup cost.

Instead, it could just have a tuple with a module and a class name to import on demand as values.

This would avoid unconditional dependencies on special interest packages. See #176 for an example.

PR #191, opened 2022-12-13
Importless declaration of URL handler

mih:
Previously, the dict based on which `AnyUrlOperations` selects a handler must contain a handler class as values. This made it necessary to import everything first, and thereby creates an avoidable startup cost.

Now one declares a tuple with a module name and a class name to import on demand as values.

This avoids unconditional dependencies on special interest packages.

Closes datalad/datalad-next#175


mih:
Test failure is an unrelated timeout

PR #196, opened 2022-12-14
Switch AppVeyor macOS to Monterey

bpoldrack:


mih (approved):
This change is motivated by a stalling of the test environment previously used on appveyor (which was "Catalina")

Thx!

PR #212, opened 2023-01-13
Make package dependencies explicit

mih:
Previously these pieces would be pulled in via other dependencies.


mih:
Failure are all due to intermittent failure of external services and to issues fixed only in datalad 0.18.1

PR #215, opened 2023-01-13
Draft of `FsspecUrlOperations`

### LINKED ISSUE #179: Implement AWS S3 UrlOperations
with -next credential support.

### LINKED ISSUE #210: Implement `fsspec`-based `UrlOperation` class
This is used by https://github.com/datalad/datalad-fuse

The aim here would be simpler: Wrap around `fsspec.open()` to provide the standard operations `download`, `upload`, `sniff`, and possibly `delete`.

This should be relatively simple in principle. More challenging would be the provisioning of the correct credentials necessary.

The list of supported protocols is impressively long:

```py
>>> print(fsspec.available_protocols())
['file', 'memory', 'dropbox', 'http', 'https', 'zip', 'tar', 'gcs', 
 'gs', 'gdrive', 'sftp', 'ssh', 'ftp', 'hdfs', 'arrow_hdfs', 
 'webhdfs', 's3', 's3a', 'wandb', 'oci', 'asynclocal', 'adl', 'abfs',
 'az', 'cached', 'blockcache', 'filecache', 'simplecache', 'dask', 
 'dbfs', 'github', 'git', 'smb', 'jupyter', 'jlab', 'libarchive', 
 'reference', 'generic', 'oss', 'webdav', 'dvc', 'root']
```

Individual protocols do require additional software dependencies to be installed. But even the list of built-ins is long https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations

Demo:

```py
>>> import fsspec
>>> of = fsspec.open('github://datalad:datalad@/README.md')
>>> fp = of.open()
>>> fp.size
23922
>>> fp.read(125)
b'     ____            _     ....
```

This is pretty much all that is needed to implement basic support for all unauthenticated transports.

For `AnyUrlOperations` we would need to come up with a way to identify URLs that this handler should or could handle.

This IO support could be the basis for `RemoteArchive` https://github.com/datalad/datalad-next/issues/184#issuecomment-1380467844

But even without it, this tech enables partial archive content access.

Here is a demo. `big.zip` is 5G in size, and contains a 4-byte text file and a large binary blob. Here I am opening the text file from inside the archive over an SSH connection:

```py
>>> timeit fsspec.open('zip://dummyzip/tiny.txt::sftp://myhost/home/mih/big.zip').open().read()
629 ms ± 8.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> fsspec.open('zip://dummyzip/tiny.txt::sftp://juseless.inm7.de/home/mih/big.zip').open().read()
b'123\n'
```

Size reporting is still possible, hence progress reporting would be possible here too!

```py
>>> fsspec.open('zip://dummyzip/fmriprep-20.0.3.simg::sftp://juseless.inm7.de/home/mih/big.zip').open().size
4873539615
```

### LINKED ISSUE #217: Allow for URL handler definition in config
Presently `AnyUrlOperations` only consults a runtime/code registry. We could...

Specify URL handlers in configuration:

The items would be of syntax: `datalad.url-handler.<regex>.<property>`,
where `<regex>` is a match expression when to apply a particular handler.

Putting the regex into the section name would prevent the problem of undesired shadowing, e.g. when some software package installation defines `datalad.url-handler.openneuro` and I already had that, but for a different match expression. Ultimately the match expressions are the unique identifier and the git config syntax allows for putting them into the subsection names.

There can be any number of different `<property>`.

The special property `class` is something like `package.module.class` that identifies what class to import and use as a handler. Any other property would be passed on to the handlers constructor as `kwargs` (after applying `.replace('-', '_')` to the property name).

Alternatively (or maybe in addition), we could have a `kwargs` property to provide JSON-encoded arguments.

Or we could try JSON-decode any property-value, and only fall back on the
str-representation if that does not work.

Both of the latter approaches would make it easier to specify non-string
values. I am thinking of cases where it would be useful to provide
timeout durations, number of retries, boolean flags, etc.

Update: real-world arguments that need to be specified can become rather complex quickly. Example:

```
{'fs_kwargs': {'anon': True, 's3': {'anon': True}}}
```

anything other than a single `kwargs` property with a JSON encodes value seems to have little applicability.

mih:
For now this is just covering downloads and has an implementation for authentication against S3 endpoints.

The included code enables downloads of selected archived content, without having to
download the entire archive first.

See datalad/datalad#373

For ZIP/TAR archives, and Github projects this is hooked into `AnyUrlOperations` and
thereby accessible via the `datalad download` command.

Demo:

```sh
❯ datalad download 'zip://datalad-datalad-cddbe22/requirements-devel.txt::https://zenodo.org/record/7497306/files/datalad/datalad-0.18.0.zip?download=1 -'
 # Theoretically we don't want -e here but ATM pip would puke if just .[full] is provided
 # Since we use requirements.txt ATM only for development IMHO it is ok but
 # we need to figure out/complaint to pip folks
 -e .[devel]

❯ datalad download 'tar://datalad-0.18.0/requirements-devel.txt::https://files.pythonhosted.org/packages/dd/5e/9be11886ef4c3c64e78a8cdc3f9ac3f27d2dac403a6337d5685cd5686770/datalad-0.18.0.tar.gz -'
 # Theoretically we don't want -e here but ATM pip would puke if just .[full] is provided
 # Since we use requirements.txt ATM only for development IMHO it is ok but
 # we need to figure out/complaint to pip folks
 -e .[devel]
```

As demo'ed in the code, dependening on the capabilities of the
particular filesystem abstraction it needs custom handling of the
actual download process after `open()` was called.

Closes #210 
Closes #179
Closes #217

TODO

- [ ] Come up with some meaningful suits of extra requirements
- [x] Under some circumstances boto wants to authenticate even if a bucket is public. Investigate and possibly switch to starting with `anon=True`. However, that would be expensive, because it would prevent automatically using a session token provided via the environment. If that is needed, we should add additional logic.
- [x] Check S3 versioned access. `datalad download 's3://openneuro.org/ds004393/sub-26/func/sub-26_task-ImAcq_acq-task_run-05_bold.nii.gz?versionId=+WMIWSpgtnESd8J2k.BfgJ3Xo7qpQ1Kjm demo.nii.gz'` works, but it would be good to have a testcase for accessing a non-recent version of a file. Setting `version_aware=False` prevents handling about URLs with version tags. What needs testing is:
   - whether not turning it on specifically would make version tags in URLs get ignored (leading to wrong outcomes)
   - whether not turning it off specifically leads to needless requests
   For an openneuro URL a version tag is reported regardless of the `version_aware` setting and regardless of whether the requesting URL had one.
- [x] verify that anon-access can be turned off when explicitly provisioning a credential 
- [ ] Investigate speed differences: `download-url` from S3 seems to be 3x faster than `download` at times.
  In turns out to be mostly a question of how data are read from the "file" pointer. The default was/is to simply iterate over the file object. This leaves the decision making to fsspec. However, not all filesystem implementations support this iteration anyways. Switching to reading direct chunks (of a size declared explicitly from the outside) remove the slowdown entirely (and actually makes the implementation about 15% faster than the (default behavior) of the downloader in datalad-core).
- [x] Add error handling in `download()` when a remote URL is already known not to be around from stat'ing


setup.cfg
```diff
@@ -36,6 +36,9 @@ devel =
 httpsupport =
     requests
     requests_toolbelt
+remotefs =
+    fsspec
+    requests
```
adswa, line 41: I discovered in #223 that a dependency missing on my system was ``aiohttp`` in addition to those listed here

adswa: When trying S3 URLs, I was missing the s3fs dependency. However, the reporting for this was very nice and user-friendly:

```
datalad -c 'datalad.url-handler.(^|.*::)s3://openneuro.org.class=datalad_next.url_operations.fsspec.FsspecUrlOperations' -c 'datalad.url-handler.(^|.*::)s3://openneuro.org.kwargs={"fs_kwargs": {"s3": {"anon": true}}}' download 's3://openneuro.org/ds004393/sub-26/func/sub-26_task-ImAcq_acq-task_run-05_bold.nii.gz demo.nii.gz'

download(error): demo.nii.gz [download failure] [Install s3fs to access S3]
```

This error comes from fsspec directly: https://github.com/fsspec/filesystem_spec/blob/e180aa859ef081215882b2d1b67d4bc33c040330/fsspec/registry.py#L114.

Interestingly, this mapping of protocols to dependencies and errors also exists for HTTP (https://github.com/fsspec/filesystem_spec/blob/e180aa859ef081215882b2d1b67d4bc33c040330/fsspec/registry.py#L70), and I did see it in #223 - but only in the special remote debug output, not bubbled up. It would be nice to find out what makes the s3 code handle this error better, and mirror it in the generic fsspec code.

adswa:
I was looking into the open questions about versioned s3 URLs by writing a unit test based on @mslw's bucket and code snippets - thanks much!

I found that handling versioned URLs works in general. However, in case of s3, there is a problem if we configure a URL handler to be not version-aware (``ops = FsspecUrlOperations(fs_kwargs={'version_aware': False})``)  but provide it with a versioned URL (such as ``s3://mslw-datalad-test0-versioned/3versions-allversioned.txt?versionId=Tro_UjqVFJfr32v5tuPfjwtOzeqYCxi2'``):

The outcome is an error that reads like this:
```
E               datalad_next.url_operations.UrlOperationsRemoteError: UrlOperationsRemoteError for 's3://mslw-datalad-test0-versioned/3versions-allversioned.txt?versionId=Tro_UjqVFJfr32v5tuPfjwtOzeqYCxi2'
```

Internally, it is a dictionary update in ``fsspec`` that fails: 

```
path = 's3://mslw-datalad-test0-versioned/3versions-allversioned.txt?versionId=Tro_UjqVFJfr32v5tuPfjwtOzeqYCxi2'
kwargs = {'version_aware': False}

    def _un_chain(path, kwargs):
        x = re.compile(".*[^a-z]+.*")  # test for non protocol-like single word
        bits = (
            [p if "://" in p or x.match(p) else p + "://" for p in path.split("::")]
            if "::" in path
            else [path]
        )
        # [[url, protocol, kwargs], ...]
        out = []
        previous_bit = None
        kwargs = kwargs.copy()
        for bit in reversed(bits):
            protocol = kwargs.pop("protocol", None) or split_protocol(bit)[0] or "file"
            cls = get_filesystem_class(protocol)
            extra_kwargs = cls._get_kwargs_from_urls(bit)
            print(extra_kwargs)
            kws = kwargs.pop(protocol, {})
            if bit is bits[0]:
                kws.update(kwargs)
>           kw = dict(**extra_kwargs, **kws)
E           TypeError: dict() got multiple values for keyword argument 'version_aware'

../../env/next/lib/python3.11/site-packages/fsspec/core.py:331: TypeError
```

This error results from the fact that ``s3fs``'s S3 class has a method ``_get_kwargs_from_urls`` which parses S3 URLs for version strings, and if it find it, it tries to tell fsspec to become version aware:

```py
    def _get_kwargs_from_urls(urlpath):
        """
        When we have a urlpath that contains a ?versionId=

        Assume that we want to use version_aware mode for
        the filesystem.
        """
        url_storage_opts = infer_storage_options(urlpath)
        url_query = url_storage_opts.get("url_query")
        out = {}
        if url_query is not None:
            from urllib.parse import parse_qs

            parsed = parse_qs(url_query)
            if "versionId" in parsed:
                out["version_aware"] = True
        return out
```

Since we already set this key, fsspec crashes. So it seems that the case "configure version awareness to be off, but supply versioned URLs nevertheless" can't be supported. I wondering if this is something that needs documentation, or if we can do some clever handling - but parsing URLs pre-emptively for version strings seems a bit costly, IMO.

adswa:
I have for now settled on the following:

- there is a unit test that confirms that version awareness is able to retrieve the file versions specified in URLs. It uses @mslw's S3 bucket, so its probably temporary, but maybe we have means to create a similar bucket under a datalad account?
- I have found that setting the 'version_aware' key is generally a bad idea, as it interferes with s3fs's autodetection and setting of this key. I have added a docstring amendment to warn users about this - I think it isn't optimally phrased, and maybe also in the wrong place; please advise improvements

adswa:
Pretty weird failure in crippledFS tests:

```

=================================== FAILURES ===================================
_____________________________ test_fsspec_download _____________________________

tmp_path = PosixPath('/crippledfs/pytest-of-runner/pytest-0/test_fsspec_download0')

    def test_fsspec_download(tmp_path):
        # test a bunch of different (chained) URLs that point to the same content
        # on different persistent storage locations
        ops = FsspecUrlOperations()
        for url in (
            # included in a ZIP archive
            'zip://datalad-datalad-cddbe22/requirements-devel.txt::https://zenodo.org/record/7497306/files/datalad/datalad-0.18.0.zip?download=1',
            # included in a TAR archive
            'tar://datalad-0.18.0/requirements-devel.txt::https://files.pythonhosted.org/packages/dd/5e/9be11886ef4c3c64e78a8cdc3f9ac3f27d2dac403a6337d5685cd5686770/datalad-0.18.0.tar.gz',
            # pushed to github
            '***0.18.0/requirements-devel.txt',
        ):
>           props = ops.download(url, tmp_path / 'dummy', hash=['md5'])

../../../../.local/lib/python3.9/site-packages/datalad_next/url_operations/tests/test_fsspec.py:30: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../.local/lib/python3.9/site-packages/datalad_next/url_operations/fsspec.py:215: in download
    fs, urlpath, props = self._get_fs(from_url, credential=credential)
../../../../.local/lib/python3.9/site-packages/datalad_next/url_operations/fsspec.py:363: in _get_fs
    fs, urlpath, props = get_fs(
../../../../.local/lib/python3.9/site-packages/datalad_next/url_operations/fsspec.py:63: in get_fs_generic
    fs, urlpath = url_to_fs(url, **kwargs)
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/core.py:375: in url_to_fs
    fs = filesystem(protocol, **inkwargs)
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/registry.py:257: in filesystem
    return cls(**storage_options)
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/spec.py:76: in __call__
    obj = super().__call__(*args, **kwargs)
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/implementations/zip.py:54: in __init__
    fo = fsspec.open(
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/core.py:439: in open
    return open_files(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <List of 0 OpenFile instances>, item = 0

    def __getitem__(self, item):
>       out = super().__getitem__(item)
E       IndexError: list index out of range

/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/fsspec/core.py:194: IndexError
```

mih:
Thanks @adswa for the S3 access test. I think this is spot on!

adswa:
The crippledFS failure also shows up in #223, but I can't reproduce it locally.

jsheunis:
Some initial usage notes:
- wasn't initially aware that I had to run `pip install -r requirements-devel.txt` in addition to a normal `pip install -e .`, so it took me a while to get started. but after that no issues.
- can confirm pretty much the same stats as @mih reports here: https://github.com/datalad/datalad-next/pull/215#issuecomment-1387276658 (just about 50% longer on my network)
- command from @adswa works without errors: https://github.com/datalad/datalad-next/pull/215#discussion_r1183304766

mih:
I think we have learned a lot here, but nobody was able to figure out the speed issues. This will need to be picked up later, or replaced by something else entirely.

Thanks to everyone who tried.

adswa:
FTR, I'm linking two relevant comments to this PR:

@christian-monch found, I believe similar to what @mih found in https://github.com/datalad/datalad-next/pull/215#issuecomment-1387276658, that the block size makes a different for download speeds, in this particular case, for partial 7z archives: https://github.com/datalad/datalad-ria/issues/50#issuecomment-1711854929.

In @christian-monch's explorations, a block size of 1 resulted in a general speed up in different usecases.

I tried to test this in this PR, too, rediscovered @mih's initial comment about this, and found no such general improvements for s3 downloads: https://github.com/datalad/datalad-ria/issues/50#issuecomment-1717465527

PR #216, opened 2023-01-17
CredentialManager.obtain() helper for a standard credential workflow

mih:
See docstring for details.

This replaces a more specific implementation in the context of `create_sibling_webdav()`, and is now available as a more generic helper for use in other scenarios too.

### DIFF
--- a/datalad_next/utils/credman.py
+++ b/datalad_next/utils/credman.py
@@ -8,6 +8,9 @@
 # ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
 """Credential management and query"""
 
+# allow for |-type UnionType declarations
+from __future__ import annotations
+
 __docformat__ = 'restructuredtext'
 
 __all__ = ['CredentialManager']
@@ -15,6 +18,12 @@ __all__ = ['CredentialManager']
 from datetime import datetime
 import logging
 import re
+
+from typing import (
+    Dict,
+    List,
+    Tuple,
+)
 
 import datalad
 from datalad_next.exceptions import (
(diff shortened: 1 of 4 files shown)


mih:
Windows is back! The only failure is caused by a glitch with `httpbin`

mih:
Now complete with tests.

christian-monch (dismissed):
LGTM

datalad_next/utils/credman.py
```diff
@@ -553,6 +558,116 @@ def get_sort_key(x):
...
+        ------
+        ValueError
+          Raised when no matching credential could be found and none was
+          entered. Also raised, when a credential selected from a query result
+          or a manually entered one is missing any of the properties with a
+          name given in ``expected_props``.
+        """
+        cred = None
+        if not name:
+            if query_props:
```
adswa, line 635: Sorry if I'm missing something, but the docstring (first sentence) sounds as if I should be able to obtain a credential solely by prompt, without the need for query parameters. This conditional would prevent this, if I see it correctly

mih: Yes, that observation is correct. The type hint is needed such that the credential manager knows what to prompt for (some credentials need more than just a secret). This limitation could be lifted (to be able to just ask for a secret). Or we add docs, as you suggested.

I am always hesitant re examples. Unless executable (which most of the time required added/undesired complexity) they become invalid quickly, which often goes unnoticed. But I do see the issue.

mih: I have pushed some examples. Would that work for you?

adswa: Thx

Issue #198, opened 2022-12-14
Structured error reporting for `EnsureCommandParameterization`


mih:
ATM the encoding of validation errors and correction hints can be overwhelming for more complex constraint setups. There should be an option (or default) to get structured reports that could be rendered in different (environment-optimized ways), e.g. as an input form annotation in a GUI.

mih:
The most meaningful approach (known to me) for this is pydantic's `ValidationError`: https://docs.pydantic.dev/usage/models/#error-handling

mih:
I looked into https://docs.python.org/3/library/exceptions.html#exception-groups (backport available at https://pypi.org/project/exceptiongroup/).

This is not applicable directly, because it is location the subject of a constraint violation (we must know which parameter is wrong, not just that any parameter is wrong in a particular fashion). However, exception groups are essentiall exceptions that contain other exceptions (in a tuple). For the purpose here it would be sufficient to contain them in a mapping that associates an arbitrary key, representing the context, with the associated exception.

In the scope of a command parameter constraint the keys would be `str` matching parameter names (and possibly name special names for, e.g., joint-validation).

mih:
Some thoughts:

When I need to validate a single value, key bits of information when that validation fails would be 1) what was the violated criterion, and 2) what value violated it. Presently, we encode (1) in a handcrafted message that should also include (2) But sometimes it doesn't.
I am thinking to have a `ConstraintError` that is the base exception for all this. And I am debating what arguments its constructor should have.

Concept:

- it should receive the original input value -- to make it uniformly available for inspection reporting, etc. 
- it should receive some kind of error message -- strictly focused on the key violation. With that focus, and known that the original value need not (necessarily) be included in it (because it is also available), I suspect that reporting can be more stringent, and its verbosity better configurable from the outside
- it should also receive the violated constraint (instance). This would enable a reporting function to retrieve the necessary info for doing: "This error happened" + "because you provided X" + "here is what would have been acceptable"

This would clearly assign responsibilities to the underlying code. The constraint must be able to state clearly what is valid input, and can report compactly on what criterion was violated (without having to say in the message what value violated it -- the message could become more label-like). The `ConstraintError` would then bind all (message, input, constraint) together to make it jointly accessible -- avoiding the need for external tracking logic.

Raising a ConstraintError from a Constraint would then look like `ConstraintErrorDerived(self, msg, value)`

mih:
Consider this example:

```
cmd(a=5, b=6, c=6)
[...]

PR #227, opened 2023-02-06
Allow for opt-in parameter default validation

### LINKED ISSUE #225: `create-sibling-webdav` command cannot identify dataset - related to EnsureDataset constraint
### Problem
It should be possible to call `create-sibling-webdav` without specifying the dataset explicitly ("If no dataset is given, an attempt is made to identify the dataset based on the current working directory"), as long as the working directory is inside the dataset, like in the first example from the manpage:

```
datalad create-sibling-webdav "https://webdav.example.com/myds"
```
But, using current `main`, this gives:
```
[ERROR  ] 'NoneType' object has no attribute 'ds' 
```

### Description
The error is coming from https://github.com/datalad/datalad-next/blob/bc979ab85701226f0c3814c89606f34c09072d2f/datalad_next/commands/create_sibling_webdav.py#L289

The `CreateSiblingWebDAV` class is derived from `ValidatedInterface`; `EnsureDataset` constraint is used; the `__call__` has a default value `dataset=None` (abbreviated below):

```
class CreateSiblingWebDAV(ValidatedInterface):
    _params_ = dict(
        ...
        dataset=Parameter(...)
        ...
    )

    _validator_ = CreateSiblingWebDAVParamValidator(dict(
        ...
        dataset=EnsureDataset(installed=True, purpose='create WebDAV sibling(s)')
        ...
    )

    def __call__(
            url,
            *,
            dataset=None,
            ...
    ):
        ds = dataset.ds
```

Now, if I understand correctly the [ValidatedInterface](http://docs.datalad.org/projects/next/en/latest/generated/datalad_next.commands.ValidatedInterface.html) documentation, if `dataset` is not given (optional), it gets to the validator as None, but because this matches the default value in `__call__`, it just passes through. Then, the attempt to obtain the `dataset.ds` property crashes.

#168 gives me an impression (possibly wrong) that the use of `require_dataset` (which should try to figure out the path, right?) is discouraged, and `EnsureDataset` was meant to replace it. But this doesn't seem to be the case: I tried naively removing the default value for the dataset, but it ended with  an error "Validation of parameter 'dataset' failed -caused by- Cannot create Dataset from <class 'NoneType'>".

Edit: I re-read `EnsureDataset` and realised that it calls `require_dataset`, but it first checks whether the argument is `str` or `PurePath`. If I changed that to also allow `None`, and removed the defaul value in `__call__` it would probably work, (at least from CLI) but would that be intended way?

How to restore the "figuring-out-based-on-workdir" behaviour?

mih:
This change addresses a key shortcoming of the new command parameter
validation approach. Previously, it did not process a parameter's
default value (at all), and required each command implementation to
handle a default value directly.

This caused an inconvenience for the ubiquitous `dataset` parameter,
which often defaults to `None` to indicate "will discover in CWD".
However, given that `None` is the default, this discovery would need
to be implemented in each command, rather than using the standard
`EnsureDataset`.

With this change, `EnsureCommandParameterization` can be
parameterized with a new `validate_defaults` argument to cause
processing of an individual parameter's default.

Here, this is implemented for the `create_sibling_webdav()` command
as a demonstration. Likely other commands need a similar change.

Closes #225

### DIFF
--- a/datalad_next/constraints/dataset.py
+++ b/datalad_next/constraints/dataset.py
@@ -61,7 +61,9 @@ class EnsureDataset(Constraint):
                 elif not self._installed and is_installed:
                     raise ValueError(f'{value} already exists locally')
             return DatasetParameter(value, value)
-        elif not isinstance(value, (str, PurePath)):
+        # anticipate what require_dataset() could handle and fail if we got
+        # something else
+        elif not isinstance(value, (str, PurePath, type(None))):
             raise TypeError(f"Cannot create Dataset from {type(value)}")
 
         from datalad.distribution.dataset import require_dataset
--- a/datalad_next/constraints/dataset.py
+++ b/datalad_next/constraints/dataset.py
@@ -1,7 +1,11 @@
 """Validate DataLad datasets"""
 
-from pathlib import PurePath
+from pathlib import (
+    Path,
+    PurePath,
+)
 
+from datalad_next.datasets import Dataset
 from datalad_next.exceptions import NoDatasetFound
 
 from .base import Constraint
@@ -22,7 +26,7 @@ class DatasetParameter:
 
 
 class EnsureDataset(Constraint):
-    """Ensure a absent/present `Dataset` from any path or Dataset instance
+    """Ensure an absent/present `Dataset` from any path or Dataset instance
 
     Regardless of the nature of the input (`Dataset` instance or local path)
     a resulting instance (if it can be created) is optionally tested for
@@ -33,6 +37,18 @@ class EnsureDataset(Constraint):
     rather than a `Dataset` directly. Consuming commands can discover
     the original parameter value via its `original` property, and access a
     `Dataset` instance via its `ds` property.
+
+    In addition to any value representing an explicit path, this constraint
+    also recognizes the special value `None`. This instructs the implementation
+    to find a dataset that contains the process working directory (PWD).
+    Such a dataset need not have its root at PWD, but could be located in
+    any parent directory too. If no such dataset can be found, PWD is used
+    directly. Tests for ``installed`` are performed in the same way as with
+    an explicit dataset location argument. If `None` is given and
+    ``installed=True``, but no dataset is found, an exception is raised
+    (this is the behavior of the ``required_dataset()`` function in
+    the DataLad core package). With ``installed=False`` no exception is
+    raised and a dataset instances matching PWD is returned.
     """
     def __init__(self, installed: bool = None, purpose: str = None):
         """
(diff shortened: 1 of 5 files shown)


mih:
Discovered https://github.com/datalad/datalad/issues/7281 in the process.

mih:
I hope I have arrived at the end of this. A command that want to support auto-discovery of a `dataset` would need to

- have a `dataset` parameter with a default of `None`
- set an `EnsureDataset` validator for this parameter
- enable default value processing for this parameter via `EnsureCommandParameterization(validate_defaults=('dataset',), ...)`

This now guarantees that a `DatasetParameter` instance is passed to the command parameter after validation.

If a command requires such a dataset to either exist or not exist, the `installed` parameter of `EnsureDataset` has to be set accordingly. Unlike `require_dataset()` in datalad-core, there is no implicit behavior switching for `dataset=None`.

mih:
TODO:

- `download()` has: `cfg = dataset.ds.config if dataset else datalad.cfg`
- same for `credentials`

datalad_next/constraints/dataset.py
```diff
@@ -61,15 +77,32 @@ def __call__(self, value) -> DatasetParameter:
...
+                check_installed=self._installed is True,
+                purpose=self._purpose,
+            )
+        except NoDatasetFound:
+            # mitigation of non-uniform require_dataset() behavior.
+            # with value == None it does not honor check_installed
+            # https://github.com/datalad/datalad/issues/7281
+            if self._installed is True:
+                # if we are instructed to ensure an installed dataset
+                raise
```
mslw, line 98: `raise` or `from ... raise ValidationError` (if there is a ValidationError, if not then disregard the comment)?

mslw (approved):
So if I understand correctly, when the command uses a default `dataset=None` and the dataset is not there:
- EnsureDataset(installed=False) will return a Dataset object with PWD as path, but it will not be created, and the command may need to call `create()` on its own
- EnsureDataset(installed=True) will raise a validation error complaining that the dataset is not there

With regard to `ensure_dataset`, I wouldn't mind if it was made stricter altogether (i.e. raise NoDatasetError also for explicitkly given path), bit in either case I like the consistent behavior here.

Looks good to me! I left two minor comments.

datalad_next/constraints/dataset.py
mih: There is no ValidationError (yet?) -- only ValueError. I have no opinion ATM on what would be good to have here.

datalad_next/constraints/parameter.py
```diff
@@ -258,7 +270,9 @@ def joint_validation(self, params: Dict) -> Dict:
     def __call__(self, kwargs, at_default=None) -> Dict:
         validated = {}
         for argname, arg in kwargs.items():
-            if at_default and argname in at_default:
+            if at_default \
+                    and argname not in self._validate_defaults \
+                    and argname in at_default:
```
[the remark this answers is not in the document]
mih: I can see that. I will leave it unchanged for now (to avoid yet another CI run). I should be working on general docs on how this validation works at all. This is likely a better place for this kind of info

mih:
Thanks for the review @mslw !

datalad_next/constraints/dataset.py
mih: This connects to open issue https://github.com/datalad/datalad-next/issues/198

This `ValidationError` should exist and it should be the workhorse for structured error reporting.

Issue #141, opened 2022-11-22
Support command calls with validation bypass


mih:
No matter how cheap command parameter validation is, it will always add a cost.

Maybe validation should move into a dedicated decorator (not yet sure if before or after `eval_results`). This decorator would perform validation and make the underlying command call using the validated values. In addition, it would assign an attribute to the wrapped function that points to the original function directly and enables calls without incurring the validation cost.

`pydantic` does something very similar`.

PR #234, opened 2023-02-13
Structured error reporting and exhaustive parameter validation

### LINKED ISSUE #198: Structured error reporting for `EnsureCommandParameterization`

### LINKED ISSUE #199: Exhaustive validation mode for `EnsureCommandParameterization`
In particular in conjunction with #198 and for a GUI-type application it can be useful to obtain all validation errors at once, rather than having to work through them incrementally.

mih:
This is a further improvement of the `constraints` module (ping #115).

There is a new base class `ConstraintError` that can be used be
`Constraint` implementations to communicate validation errors. it
derived from `ValueError` (the former and current choice for raising
exceptions on validation errors) to keep `except` clauses in consumer
code valid.

With `ConstraintErrors` there is a second base class that can be used
for associating constraint violations with a particular context. The
specialization `CommandParametrizationError` uses this pattern to
communicate constraint violations for a full command parameterization
at once (for each parameter individually, and for the joint parameter
set in addition).

An improved `EnsureCommandParameterization` constraint uses these
new possibilities to offer more uniform and more detailed information
on parameter validations.

With this change, a caller can now also decide whether to perform an
exhaustive parameter validation, or to fail on first error.

The `Constraint` base class has received a new convenience method
`raise_for()` that can be used to raise the new `ConstraintError`
with minimal effort. Only a message (as before with `ValueError`),
and the value that was found to be in violation need to be passed.

Thew new approach to exception raising is demo'ed for `AltConstraints`,
and `EnsureChoice`. Beyond that, no other constraints have been adjusted
yet.

Closes #198
Closes #199 (ping #141)


bpoldrack (approved):
Makes sense at looks good to me.

PR #240, opened 2023-02-17
Deploy dedicated httpbin instance on GitHub actions

mih:
This should be more reliable than the performance of the public instance.


mih:
Workflow failed on coverage reporting. Should be OK to ignore.

PR #244, opened 2023-02-17
New ConstraintWithPassthrough

mih:
Type annotation fixes and generally bring things into a more sensible, more documented, more tested state.


mih:
The failures are https://github.com/datalad/datalad-next/issues/245

Issue #245, opened 2023-02-17
"UNAUTHORIZED for url: http://localhost:8765/bearer', 401)"


mih:
This is popping up everywhere. I thought it is related to a malfunction of the public httpbin instance, but I can finally replicate it locally with the docker-based one.

```
FAILED ../commands/tests/test_download.py::test_download_new_bearer_token_nosave
```

It is *never* happening when this test runs in isolation, but always, when all of `test_download.py` run.

mih:
So what is happening seems to be this:

- It is sufficient to have `test_download_new_bearer_token` run before `test_download_new_bearer_token_nosave` to make the error appear.
- both tests operate on the same URL, hence same realm, hence the latter would pick up on any residual information left by the former
- however, `test_download_new_bearer_token` has a `finally:` clause that removes any resulting credential
- what is happening is that the secret seems to be successfully removed, but the rest remains available in the credential manager inside the auth handler. The causes a credential to be returned on query that has no token. Given that having a token in the request is the only condition for success, its absence explains the 401 error

PR #247, opened 2023-02-18
Improve legacy credential handling (and test it)

mih:
With this change, we can now communicate to users what credentials are legacy and behave different from others.

```
% datalad credentials
crcns(user_password ✗): (legacy-credential)
```

Retrieval and update handling are also improved. Importantly this is now tested (only 5 lines are missing in the entire code base; 99% coverage).
The key improvment is that users no longer have to provide type hints
for legacy credentials that are known via a "provider configuration.


mih:
Test failures are #245 (I will get to you!!)

PR #248, opened 2023-02-18
Only work for complete credentials in `DataladAuth`

### LINKED ISSUE #245: "UNAUTHORIZED for url: http://localhost:8765/bearer', 401)"

mih:
Previously, the last used credential matching a `realm` was used unconditionally. With this change all matching credentials are filtered to exclude any credentials without a secret set.

A secret might not be present for any reason (removed by an external process, or never set to begin with). Regardless of the cause for such an anomaly, such a credential can never work for authentication. Therefore, we exclude them immediately. Should no other credential matching a realm provide a secret, prompting for credential entry is attempted.

Closes #245

### DIFF
--- a/datalad_next/utils/requests_auth.py
+++ b/datalad_next/utils/requests_auth.py
@@ -122,11 +122,17 @@ class DataladAuth(requests.auth.AuthBase):
             # get a realm ID for this authentication scheme
             realm = get_auth_realm(url, auth_schemes, scheme=ascheme)
             # ask for matching credentials
-            creds = self._credman.query(
-                _sortby='last-used',
-                type=ctype,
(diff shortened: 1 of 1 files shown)


mih:
This fixed using broken credentials, but it does not make the CI green again.

mih:
This is progress, but #249 remains to be fixed.

Issue #249, opened 2023-02-18
Credential type not saved in test


mih:
This is evident from https://github.com/datalad/datalad-next/pull/248 even after it fixes #245

```
FAILED ../commands/tests/test_download.py::test_download_new_bearer_token - AssertionError: Got 0 instead of 1 expected results matching
3179  {
3180   "cred_secret": "token123",
3181   "cred_type": "token"
3182  }
3183Inspected 1 record(s):
3184  [
3185   {
3186    "action": "credentials",
3187    "cred_secret": "token123",
3188    "name": "dataladtest_test_download_new_bearer_token",
3189    "status": "ok"
3190   }
3191  ]
3192
```

So far, this does not replicate for me locally, and also not outside the main CI run on an appveyor machine. It seems to be some evil interaction with the test setup. Grrrr.

adswa:
I tried this locally on my machine and the test passes.

mih:
Thx for trying!

I pushed a small push to get some insight. Here it is:

```diff
diff --git a/datalad_next/commands/tests/test_download.py b/datalad_next/commands/tests/test_download.py
index 45fff9f..96a83c1 100644
--- a/datalad_next/commands/tests/test_download.py
+++ b/datalad_next/commands/tests/test_download.py
@@ -208,9 +208,16 @@ def test_download_no_credential_leak_to_http(capsys):
     'dataladtest_test_download_new_bearer_token',
 ])
 def test_download_new_bearer_token(capsys):
+    from pprint import pprint
+    print('====query=====')
+    credentials('query')
+    print('====query done=====')
     try:
         download({f'{hbsurl}/bearer': '-'})
         # and it was saved under this name
+        print('====checkcred=====')
+        pprint(credentials('get', name='dataladtest_test_download_new_bearer_token', on_failure='ignore'))
+        print('====checkcred done=====')
         assert_result_count(
             credentials(
                 'get',
@@ -218,10 +225,15 @@ def test_download_new_bearer_token(capsys):
             1, cred_secret='token123', cred_type='token',
         )
     finally:
-        credentials(
+        print('====remove=====')
+        pprint(credentials(
             'remove',
             name='dataladtest_test_download_new_bearer_token',
-        )
+        ))
+        print('====remove done=====')
+        print('====checkcred=====')
+        pprint(credentials('get', name='dataladtest_test_download_new_bearer_token', on_failure='ignore'))
+        print('====checkcred done=====')
 ```

Which gives:

```
748====query=====
[...]

PR #252, opened 2023-02-22
RF of `CredentialManager.get()` related code

mih:
Trying to find what is happening with #249 by making the implemented logic less complex.


mih:
#249 still unresolved, the rest is good.

PR #253, opened 2023-02-22
More `CredentialManager` RF

mih:
- Uniform means to determine a secret.
- Use those means in `set()` too
- Document oddities
- No change in behavior


mih:
Still #249 but otherwise fine.

PR #254, opened 2023-02-22
New `memory_keyring` fixture

mih:
This fixture provides an isolated keyring in-memory-only backend for tests that needs to store/retrieve credentials.

The fixture is function-scope, hence the backend is destroyed at the end of each test.

Closes https://github.com/datalad/datalad/issues/7295


mih:
The failure pattern is still #249, but also only that.

PR #256, opened 2023-02-23
Make dataset-lookup behavior of `credentials` identical to `download` 

mih:
When called with `dataset=None`, `credentials()` would auto-locate
a dataset containing CWD (and use its configuration), whereas
`download()` would use the global configuration in this case.

This change makes the behavior uniform and chooses what `download()`
did as the model behavior.

This uniformatization is meaningful on its own.

However, due to datalad/datalad#7299 it is
also a necessity, because otherwise credentials deposited by
`download()` remain invisible to `credentials()` -- which is the
superficial trigger for datalad#249

Closes datalad#249


mih:
Still failing in a miriad of way -- BUT NOT A SINGLE ONE IS #249 !!!

PR #257, opened 2023-02-23
Do credential type detection for reporting

### LINKED ISSUE #251: Better support for lonely token credential specification
```sh
DATALAD_CREDENTIAL_MIKE_TOKEN=some datalad credentials get mike
mike(secret ✗): [token=some]
```

So right now it considers it to be a credential, but a credential without a secret and a `token` property.

This is suboptimal. This means that tokens can end up in a config file, more or less silently. It also means that we do not know anything about the nature of the credential (although we could), and we cannot say whether it is complete or not (although we could).

I'd say we should not only refuse to store credentials without a secret, but we should also disregard any credentials that can be determined to be incomplete.

In the case above a user could just use `secret` instead of `token`, but if they do not, we should either refuse (credential with no secret) or auto-convert (credential of type token with secret as given by the token env var).

mih:
For now the only scenario is a credential specification that comprises a token-property only.

However, with the recent RF there is now a dedicated piece of code that could be extended arbitrarily to support additional scenarios.

```
❯ DATALAD_CREDENTIAL_MIKE_TOKEN=some dl -f json_pp credentials get mike
{
  "action": "credentials",
  "cred_secret": "some",
  "cred_type": "token",
  "name": "mike",
  "status": "ok"
}

```
Closes #251

### DIFF
--- a/datalad_next/credman/manager.py
+++ b/datalad_next/credman/manager.py
@@ -201,8 +201,9 @@ class CredentialManager(object):
             # no secret, no credential
             if any(not p.startswith('_') for p in cred):
                 lgr.debug(
-                    'Not reporting on credential fragment with no secret: %r',
-                    cred,
+                    'Not reporting on credential fragment '
+                    '(name=%r) with no secret: %r',
+                    name, cred,
                 )
             return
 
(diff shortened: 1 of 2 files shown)


mih:
More green! All the failures are unrelated to this PR.

Issue #193, opened 2022-12-14
Provide `EnsureRemoteName`


mih:
Many commands needs sibling names, existing or prospective.

Checking for syntax compliance is essentially

`EnsureGitRefName(allow_onelevel=True, refspec_pattern=False)`

for any given dataset, it can be turned into a `EnsureChoice()`, but it would also need some kind of `EnsureNotIn()` (for setting a non-existing remote name). So likely the best is to implement dataset interaction directly in `EnsureRemoteName`, and make it a subclass of `EnsureGitRefName`

Issue #131, opened 2022-11-15
Implement `EnsureCredential`


mih:
suggested by https://github.com/datalad/datalad-next/issues/130#issuecomment-1313629085

mih:
This would also be a nice candidate to implement tests for `Constraint.for_dataset()`

PR #260, opened 2023-02-24
On-demand constraint-tailoring to dataset arguments

### LINKED ISSUE #200: Dataset-scope tailoring of `EnsureCommandParameterization` validation
A dataset often provides a critical piece of information to be able to validate properly and in-depth. For example, the name of an existing remote can only be validated once a dataset is determined in which this remote exists.

`Constraints` have `for_dataset()` that is meant to be used for this purpose.

`EnsureCommandParameterization` would need to call this method for any constraint in order to obtain the most adequate variant.

It could make sense to have the validation order be significant, i.e. have no `Dataset` identified, until a `dataset=` argument is processed, only only from there on call `for_dataset()` on the subsequent constraints.

It may also be necessary to support an alternative name declaration for a "dataset" command, in case one and the same command has two arguments of this type.

For that reason, it may also be beneficial to ignore the name of an argument and go by constraint-type (i.e. `EnsureDataset`).

Combined with validation order rules (a class author sets the order of arguments in `_validator_`) it may be sensible to start tailoring constraints when a first dataset-type (not name) argument is found, tailor all constraints from there on, and switch the dataset for tailoring, whenever another dataset-type argument is found.

mih:
This exposes a, so far dormant, feature of the `Constraint` class to
parameter validation: constraint tailoring for particular datasets.

`EnsureCommandParameterization` learned a `tailor_for_dataset` parameter
that can be used to identify which parameters' constraints should be
tailored for which `Dataset` instances. Tailoring will only be actually
done under the following conditions:

- the dataset-providing parameters need to evaluate to a
  DatasetParameter instance, typically via `EnsureDataset`
  (all regular conditions apply, e.g. a default `None` not being processed
  etc)
- the to-be-tailored parameter `Constraint` needs to implement its
  `for_dataset()` method to perform a tailoring.

A test with some custom-made constraints is included. At the moment,
no production-ready constraints do implement `for_dataset()`. However,
now that a consumer for that is available, it makes more sense to
address related issues, such as:

- #193
- #131

Closes #200

### DIFF
--- a/datalad_next/constraints/parameter.py
+++ b/datalad_next/constraints/parameter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Container
+from itertools import chain
 from typing import (
     Any,
     Callable,
@@ -26,7 +27,7 @@ from .compound import (
     EnsureIterableOf,
     EnsureMapping,
 )
-
+from .dataset import DatasetParameter
 from .exceptions import (
     ConstraintError,
     ParametrizationErrors,
@@ -291,6 +292,7 @@ class EnsureCommandParameterization(Constraint):
         validate_defaults: Container[str] | None = None,
         joint_constraints:
             Dict[ParameterConstraintContext, Callable] | None = None,
+        tailor_for_dataset: Dict[str, str] | None = None,
     ):
         """
         Parameters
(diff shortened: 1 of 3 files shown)


mih:
All CI failure is unrelated.

PR #269, opened 2023-03-02
BF: Teach _execute_command to handle DatasetParameters

### LINKED ISSUE #266: Failure to convert outcome of EnsureDataset() constraint to Dataset
I introduced an ``EnsureDataset()`` constraint to a command in ``metalad``. Validation succeeds, but the conversion of the outcome of ``EnsureDataset`` into a Dataset fails subsequently:

```
❱ datalad  meta-extract -d . metalad_example_file /tmp/ds1/some.txt 
[ERROR  ] Could not determine resource identifier type for <datalad_next.constraints.dataset.DatasetParameter object at 0x7fb64b692350> -caused by- expected str, bytes or os.PathLike object, not DatasetParameter 
```

The reason for this seems to be that the set of validated parameters contains a DatasetParameter object:

```
p final
{'context': None, 'result_xfm': None, 'extractorname': 'metalad_example_file', 'force_dataset_level': False, 'get_context': False, 'extractorargs': [], 'return_type': 'generator', 'result_renderer': 'tailored', 'on_failure': 'continue', 'path': PosixPath('/tmp/ds1/some.txt'), 'dataset': <datalad_next.constraints.dataset.DatasetParameter object at 0x7f4742171190>, 'result_filter': None}
```

After successful validation, this dict of parameters is given for further handling (at least in this case) to ``datalad_next/patches/interface_utils.py(146)_execute_command_()``. This function can not handle ``DatasetParameter``'s, though:

```
(Pdb) l
168  	    if dataset_arg is not None:
169  	        from datalad_next.datasets import Dataset
170  	        if isinstance(dataset_arg, Dataset):
171  	            ds = dataset_arg
172  	        else:
173  ->	            try:
174  	                ds = Dataset(dataset_arg)
175  	            except ValueError:
176  	                pass
177  	    # look for hooks
178  	    hooks = get_jsonhooks_from_config(ds.config if ds else dlcfg)
(Pdb) n
> /home/adina/repos/datalad-next/datalad_next/patches/interface_utils.py(174)_execute_command_()
-> ds = Dataset(dataset_arg)
(Pdb) 
ValueError: Could not determine resource identifier type for <datalad_next.constraints.dataset.DatasetParameter object at 0x7f4456b97690>
```.

My question is: Should the parameter validation return the actual dataset instead of the DatasetParameter (which could be accessed through the DatasetParameter), or should downstream code learn how to handle DatasetParameter?

Its also not entirely clear why this does not happen for ``next``'s ``download`` command (from which I copied how to use ``EnsureDataset()`` - maybe that one doesn't go through patches?

adswa:
fixes #266

This piece of code was not able to handle DatasetParameter objects, and instead crash with a ValueError. This patch should fix this.

### DIFF
--- a/datalad_next/patches/interface_utils.py
+++ b/datalad_next/patches/interface_utils.py
@@ -33,7 +33,7 @@ from datalad.interface.utils import (
 )
 from datalad_next.exceptions import IncompleteResultsError
 from datalad_next.utils.patch import apply_patch
-
+from datalad_next.constraints.dataset import DatasetParameter
 
 # use same logger as -core
 lgr = logging.getLogger('datalad.interface.utils')
@@ -169,6 +169,8 @@ def _execute_command_(
         from datalad_next.datasets import Dataset
         if isinstance(dataset_arg, Dataset):
             ds = dataset_arg
+        elif isinstance(dataset_arg, DatasetParameter):
+            ds = dataset_arg.ds
         else:
             try:
                 ds = Dataset(dataset_arg)


mih (approved):
Thx! LGTM.

All errors are relative to httpbin and have nothing to do with this code.

PR #271, opened 2023-03-02
Implement EnsurePath.for_dataset() and resolve against that dataset

### LINKED ISSUE #270: Implement resolve_path as a constraint
With #260 merged, we can enhance `EnsurePath` to automatically perform DataLad-unique way of resolving relative paths.

mih:
This implements DataLad's standard path resolution approach (relative
is relative to CWD, unless a `Dataset` instance is given).

This now makes it possible to perform this standard resolution pattern
during command parameter validation, like so:

```py
# example validator for a command with two args
EnsureCommandParameterization(
    param_constraints=dict(
        dataset=EnsureDataset(),
        path=EnsurePath(),
    ),
    tailor_for_dataset=dict(path='dataset'),
)
```

This will:

- cause the 'dataset' validator to run first
- auto-generate an `EnsurePath` variant that resolves against that
  dataset
- ensures that the command always receives absolute paths, resolved
  against the dataset whenever the rules require it.

Closes #270

### DIFF
--- a/datalad_next/constraints/base.py
+++ b/datalad_next/constraints/base.py
@@ -5,19 +5,25 @@ from __future__ import annotations
 
 __docformat__ = 'restructuredtext'
 
-__all__ = ['Constraint', 'Constraints', 'AltConstraints']
-
-from typing import (
-    TYPE_CHECKING,
-    TypeVar,
-)
+__all__ = ['Constraint', 'Constraints', 'AltConstraints',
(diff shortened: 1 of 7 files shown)


adswa:
I have used this check in https://github.com/datalad/datalad-metalad/pull/362 and it was very useful to have!

mih:
Failures are unrelated.

PR #279, opened 2023-03-03
RF: Simplify EnsureDataset and add optional ID check

### LINKED ISSUE #272: Add a dataset ID check for EnsureDataset?
In the context of metalad, I came across other custom checks that may make sense to integrate into existing Constraints. One custom additional check metalad does is a check for a valid dataset ID. Could ``EnsureDataset()`` get a check like this? Or should the constraints not get overloaded with options?

adswa:
This sits on top of #271 to prevent a merge conflict and make use of the changes required for type checking already implemented in it.

This PR adds two changes in individual commits:

The first commit refactors `EnsureDataset` - there was some duplication that I thought could be reduced by factoring out a ``_require_dataset()`` helper.

The second commit adds an optional `idcheck` to `EnsureDataset` to fix #272

### DIFF
--- a/datalad_next/constraints/dataset.py
+++ b/datalad_next/constraints/dataset.py
@@ -44,7 +44,7 @@ class EnsureDataset(Constraint):
     def __init__(self,
                  installed: bool | None = None,
                  purpose: str | None = None,
-                 idcheck: bool | None = None):
+                 require_id: bool | None = None):
         """
         Parameters
         ----------
@@ -60,7 +60,7 @@ class EnsureDataset(Constraint):
         """
         self._installed = installed
         self._purpose = purpose
-        self._idcheck = idcheck
+        self._require_id = require_id
         super().__init__()
 
     def __call__(self, value) -> DatasetParameter:
@@ -72,7 +72,7 @@ class EnsureDataset(Constraint):
         elif not isinstance(value, (str, PurePath, type(None))):
             raise TypeError(f"Cannot create Dataset from {type(value)}")
         else:
-            ds = _require_dataset(self, value)
+            ds = self._require_dataset(value)
         assert ds
         if self._installed is not None:
             is_installed = ds.is_installed()
@@ -82,7 +82,7 @@ class EnsureDataset(Constraint):
                 # for uniformity with require_dataset() below, use
                 # this custom exception
                 raise NoDatasetFound(f'{ds} is not installed')
-        if self._idcheck and not ds.id:
+        if self._require_id and not ds.id:
             raise NoDatasetFound(f'{ds} does not have a valid '
                                  f'datalad-id')
         return DatasetParameter(value, ds)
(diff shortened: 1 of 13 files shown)


datalad_next/constraints/dataset.py
```diff
@@ -50,7 +41,10 @@ class EnsureDataset(Constraint):
     the DataLad core package). With ``installed=False`` no exception is
     raised and a dataset instances matching PWD is returned.
     """
-    def __init__(self, installed: bool = None, purpose: str = None):
+    def __init__(self,
+                 installed: bool | None = None,
+                 purpose: str | None = None,
+                 idcheck: bool | None = None):
```
mih, line 47: What about `with_id: bool`? It would read "ensure dataset with id", not "ensure dataset idcheck". The code is als not checking the ID, only whether it is more than nothing.

Maybe "with" have too much of meaning in the Python context. `require_id` would also do it for me.

datalad_next/constraints/dataset.py
```diff
@@ -60,54 +54,64 @@ def __init__(self, installed: bool = None, purpose: str = None):
...
+                                 f'datalad-id')
         return DatasetParameter(value, ds)
 
     def short_description(self) -> str:
         return "(path to) {}Dataset".format(
             'an existing ' if self._installed is True
             else 'a non-existing ' if self._installed is False else 'a ')
+
+
+def _require_dataset(self, value):
```
mih, line 96: Better indent to make its nature as a method more obvious

datalad_next/constraints/tests/test_special_purpose.py
```diff
@@ -356,3 +356,14 @@ def test_EnsureDataset(tmp_path):
...
+    # smoke test for idcheck:
+    assert EnsureDataset(idcheck=True)(ds).ds == ds
+    assert EnsureDataset(idcheck=False)(ds).ds == ds
+    # unset the dataset ID to test whether an ID check would raise, but
+    # bring it back later in case future tests need it
+    id = ds.config.get('datalad.dataset.id')
+    ds.config.unset('datalad.dataset.id')
+    with pytest.raises(NoDatasetFound):
+        EnsureDataset(idcheck=True)(tmp_path)
+    ds.config.set('datalad.dataset.id', id)
```
mih, line 369: The dataset is temporary, there is no need to undo the changes. If you want to be more specific, you could use `scope='branch'` for `ds.config.unset()`.

If you want to be extra-safe re config leakage, use the `datalad_cfg` fixture. But all tests in `-next` are automatically verified to not change the persistent configuration. So if this one passes, we are safe.

mih (changes requested):
Thanks, this works for me! I only left minor, and a naming suggestion. I could be convinced otherwise, though.

datalad_next/constraints/dataset.py
adswa: I personally like ``require_id`` best

datalad_next/constraints/dataset.py
adswa: thx

datalad_next/constraints/tests/test_special_purpose.py
adswa: thx for the tip!

datalad_next/constraints/dataset.py
adswa: Renamed it to ``require_id`` now.

datalad_next/constraints/dataset.py
adswa: done

datalad_next/constraints/tests/test_special_purpose.py
adswa: I've changed the scope to branch.

mih (approved):
LGTM! Thx!

Issue #280, opened 2023-03-03
Decomplexify command parameter constraint setup


mih:
Right now, a command has one constraint that jointly covers all parameters and all API through which a command can be used.

This leads to quite complex constraints in complex cases.

One approach to making things less complex would be to support per API constraints, and effectively allow for trimming the complexity down to what is actually possible/needed. For example, the CLI-focused constraints will only ever have to does with `str` inputs. With a dedicated set of constraints, documenting them would also become API-specific, largely avoiding the need for all these `CMD` or `PY` macros in the docs.

Another approach would be to make constraints hierarchical, i.e. have a CLI-specific one that runs first. Its only purpose is to convert `str` input into Python dtypes that the Python API wants to support. A second constraint could then run and perform the rest/majority of argument validation code.
This second approach may have the advantage to provide a dedicated constraint to with with for the CLI, which could be less complex than with approach one, and possibly better suitable for things like auto-completion.

This second constraint would then only be triggered in `datalad.cli.exec.call_from_parser()` and process `kwargs`, before they get passed to a commands `__call__()`. That would offer a second advantage of approach 2: It requires no changes in general validation, but be a self-contained, API-specific step, that is implemented entirely in CLI code. 

Likewise, a similar GUI-focused implementation would happen entirely outside datalad-core code. Gooey already implements constraint overrides for its purpose. To be able to encapsulate that properly and uniformly across APIs would be a great step forward.

Issue #277, opened 2023-03-03
Rethink naming scheme for constraints


mih:
ATM most of them are called `Ensure...`. This is not only needlessly redundant, but also ambiguous. What is being ensured? The inputs to match a specification? Or the output to be of a certain kind?

mih:
It seems `EnsureMapping` is not a good name given the rules above. What it ensures is that there is a single-item-dict (IOW a single key-value-pair in `dict` format). It does accept the specification of such a mapping in different formats and types.

adswa:
I think a part of why I kept thinking along the lines of 

> what can go into a constraint instance, 

rather than 

> what comes out of a constraint instance

came from the fact that I translated "constraint" more in the sense of "limiting/restricting" than "force/coercion", but constraint is clearly a synonym for both. The dictionary defines it as

```
* The threat or use of force to prevent, restrict, or dictate the action or thought of others.
* The state of being restricted or confined within prescribed bounds.
* One that restricts, limits, or regulates; a check. 
```

 I'll attempt an overview of available constraints, their restriction on inputs, and their coercion of outputs. It might bring non-ideal names to the surface and could be a useful addition to the docs, too.
 
 
|Constraint name | Input limitation| Output coercion | Required additional constraints |
| ------------- | ------------- |------------- | -------------- |
|**Base constraints**|
|  ``EnsureBool`` | Input must be convertible to boolean  | Boolean  |
| ``EnsureCallable``  | Input must be a callable  | None  |
|  ``EnsureChoice`` |  Input must be an element of a set of allowed elements | None  |
| ``EnsureEnsureDType``  | Input must be convertible to a specified data type  |  returns input as specified data type |
|  ``EnsureFloat`` |  Input must be convertible to data type Float |  returns input as Float |
| ``EnsureInt``  |  Input must be convertible to data type Int | returns input as Int  |
|  ``EnsureKeyChoice`` | Input must contain a key that is an element of a set of allowed elements  | None  | #307: Needs an additional constraint that the input is a dict 
| ``EnsureNone``  |   |   |
|  ``EnsurePath`` | Input must be convertible to a Path, plus additional optional checks  | returns a Path Object  |
| ``EnsureRange``  | Input must be within a particular range  | None  | Needs and additional constraint that the input is int or float
|  ``EnsureStr`` | Input must be a string, and optionally comply to length and pattern matching rules  | None  |
| ``EnsureStrPrefix``  | Input must be a string and start with prefix  |  None |
[...]

PR #284, opened 2023-03-03
NF: Add EnsureSet to test for a required set of elements

### LINKED ISSUE #282: Create an EnsureSet() Constraint
``datalad-metalad``'s ``meta-add`` command has very specific requirements for metadata records. One of the requirements is the presence of specific sets of keys:

```
    required_keys = (
        "type",
        "extractor_name",
        "extractor_version",
        "extraction_parameter",
        "extraction_time",
        "agent_name",
        "agent_email",
        "dataset_id",
        "dataset_version",
        "extracted_metadata")

    optional_keys = (
        "path",
        "dataset_path")

    required_additional_keys = (
        "root_dataset_id",
        "root_dataset_version")
```

I found ``EnsureChoice()``, which could check that an input is element of a set of possible values, but I believe meta-add would need a check that an input contains all required values.

adswa:
Here's an attempt to write a new Constraint. Its name is not yet perfect, but so far the best I've come up with. I think the messaging also isn't yet ideal - any feedback is very welcome. My usecase for this constraint is a check whether all required keys in an input dictionary are set. Unlike EnsureChoice(), this constraint allows for more than one item. This constraint also allows additional elements to be present, as long as all requirement elements are found. Fixes #282

### DIFF
--- a/datalad_next/constraints/basic.py
+++ b/datalad_next/constraints/basic.py
@@ -283,6 +283,39 @@ class EnsureKeyChoice(EnsureChoice):
         return '%s:{%s}' % (self._key, ', '.join([repr(c) for c in self._allowed]))
 
 
+class EnsureSet(Constraint):
+    """Ensure an input contains all elements of a set of required values"""
+
+    def __init__(self, *values):
+        """
+        Parameters
+        ----------
+        *values
+           required values.
+        """
+        self._required = set(values)
+        super(EnsureSet, self).__init__()
+
+    def __call__(self, value):
+        if not self._required.issubset(set(value)):
+            self.raise_for(
+                value,
+                "missing {required}",
+                required=self._required,
+            )
+        return value
+
+    def long_description(self):
+        return 'All following fields are required: [CMD: %s CMD][PY: %s PY]' % (
+            str(tuple(i for i in self._required if i is not None)),
+            str(self._required)
+        )
+
+    def short_description(self):
+        return '{%s}' % ', '.join([repr(c) for c in sorted(self._required)])
+
+
+
 class EnsureRange(Constraint):
     """Ensure an input is within a particular range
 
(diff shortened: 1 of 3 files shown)


datalad_next/constraints/basic.py
```diff
@@ -283,6 +283,39 @@ def short_description(self):
...
+                value,
+                "missing {required}",
+                required=self._required,
+            )
+        return value
+
+    def long_description(self):
+        return 'All following fields are required: [CMD: %s CMD][PY: %s PY]' % (
+            str(tuple(i for i in self._required if i is not None)),
+            str(self._required)
```
mih, line 311: I see the point, and that `EnsureChoice` is doing the exact same thing. However, this is precisely what #280 is about. It is incomprehensible that this is because `None` cannot be given on the CLI, hence it should not be presented in the list of requirements (not sure if that is perfectly sound here, but for `EnsureChoice` this is what it is -- I made it such).

Just venting...

datalad_next/constraints/basic.py
```diff
@@ -283,6 +283,39 @@ def short_description(self):
         return '%s:{%s}' % (self._key, ', '.join([repr(c) for c in self._allowed]))
 
 
+class EnsureSet(Constraint):
```
mih, line 286: I am thinking about a good naming rule for this since #277. Presently I think, the name should reflect the outcome of the processing. Here we get `input is output` whenever a `required.issubset()` criterion is fulfilled. What about something like `EnsureContainsSet()`?

This should cause no expectations of type-changes.

Docs should state that valid input is any iterable.

mih (changes requested):
Thx! I left some comments that are mostly inspired/informed by issues that apply to the rest of this code base too.

datalad_next/constraints/basic.py
adswa: Yes, I struggled with messaging...

mih:
I am closing this PR after way too much negligence. ATM, we have no clear use case for this feature (the only one is metalad, and it is unclear to me, what the goals for it are).

This PR should be revisited once #282 is tackled again.

Thanks, nevertheless for this initiative!

PR #310, opened 2023-03-09
Let Constraint names reflect their function better

adswa:
This PR is up for debate as a response to #277.  Curious on everyone's thoughts whether this makes things better or worse. @mih @jsheunis @bpoldrack

### DIFF
--- a/datalad_next/__init__.py
+++ b/datalad_next/__init__.py
@@ -50,7 +50,7 @@ import datalad_next.patches
 from datalad.support.extensions import register_config
 from datalad_next.constraints import (
     EnsureBool,
-    EnsureChoice,
+    IsChoice,
 )
 register_config(
     'datalad.credentials.repeat-secret-entry',
@@ -80,7 +80,7 @@ register_config(
 register_config(
     'datalad.runtime.parameter-violation',
     'Perform exhaustive command parameter validation, or fail on first error?',
-    type=EnsureChoice('raise-early', 'raise-at-end'),
+    type=IsChoice('raise-early', 'raise-at-end'),
     default='raise-early',
     dialog='question',
 )
--- a/datalad_next/commands/create_sibling_webdav.py
+++ b/datalad_next/commands/create_sibling_webdav.py
@@ -34,11 +34,11 @@ from datalad.interface.common_opts import (
 )
 from datalad_next.constraints import (
     EnsureBool,
-    EnsureChoice,
+    IsChoice,
     EnsureInt,
     EnsureParsedURL,
-    EnsureRange,
-    EnsureStr,
+    IsRange,
+    IsStr,
 )
 from datalad_next.constraints.dataset import EnsureDataset
 from datalad_next.utils import CredentialManager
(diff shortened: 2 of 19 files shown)


mih (commented):
Thanks for putting the conceptual idea to this test. It is much easier to form an opinion this way!

Given what is see now, I have the feeling that this is a step in the right direction, but it has to be exercised with care:

I like
- the simplicity of the `Is` and `To` name prefixes
- I like being able to tell immediately whether something will only perform a check, or only a conversion, or something more complex

I dislike
- the semantic impression of the current implementation, e.g. `IsStrPrefix` does not check whether something IS a prefix, but whether it HAS as prefix. Same for `IsRange` which checks that something is within a particular range.
- the prospect of a proliferation of implementation enforced by this purity: See https://github.com/datalad/datalad-next/issues/277#issuecomment-1462352703 for an example that comes from the fact that `EnsureNone` turned into `IsNone` (with the implied expectation that only a check is made). More classes mean more need to document and more testing. In this example, it may well turn out that `EnsureNone` will be eventually replaced with `IsNone`, `ToTranslation`, *and* `EnsureNone`.

What to conclude from these feelings?

Maybe `Is...` should be `Matches...` to better fit the breadth of implementations, but still communicate that no transformation of inputs is done. `MatchesRange`, `MatchesPrefix`, `MatchesChoice`. Not as short, though.

Given the wide-spread impact on existing code, maybe we should propose a naming setup for *new* classes and transition the old ones more carefully and only when things fit perfectly, rather than in bulk. Stuff like `IsCallable`, or `IsChoice` (or the `Matches...` equivalents appear to be no-brainers).

adswa:
> Maybe Is... should be Matches... to better fit the breadth of implementations, but still communicate that no transformation of inputs is done. ```MatchesRange```, ``MatchesPrefix``, ``MatchesChoice``. Not as short, though.

I think brevity isn't the best dimension to optimize for, I wouldn't dislike any other prefix just because its longer - a longer prefix could even help to distinguish prefixes better visually :) 

There is, indeed, a semantic mismatch, and it arises naturally from the fact that we reduce the breadth of functionality performed by these Constraints to currently only 3 dimensions: Validity Checks, Output transformation, both. I would says this:
- Fully semantic correctness can only be achieved if we indeed name Constraints with an entirely free-form naming scheme; I personally would sacrifice semantic correctness partially for order
- Switching from one Prefix (Ensure) to three is an improvement over the status quo with regards to semantic correctness. 

Let me translate a few possibilities to view them side by side to see which one makes sense in most cases:
| ``Is`` | ``Matches``| ``Check`` | 
| ------ |---------------- |----------------|
| ``IsCallable`` | ``MatchesCallable`` |  ``CheckCallable`` 
| ``IsChoice`` | ``MatchesChoice`` |  ``CheckChoice``
| ``IsKeyChoice`` | ``MatchesKeyChoice`` |  ``CheckKeyChoice``
| ``IsRange`` | ``MatchesRange`` |  ``CheckRange``
| ``IsValue`` | ``MatchesValue`` |  ``CheckValue``
| ``IsStr`` | ``MatchesStr`` |  ``CheckStr``
| ``IsStrPrefix`` | ``MatchesStrPrefix`` |  ``CheckStrPrefix``
| ``IsNone`` | ``MatchesNone``|  ``CheckNone``
| ``IsURL`` | ``MatchesURL`` |  ``CheckURL``

(I'm personally split - anything that does value comparisons works well with ``Matches``, anything that does more general or type comparisons (``IsStr``, ``IsURL``) not 100%) - but I personally don't mind what we choose as long as we can identify and document a rule.

> Given the wide-spread impact on existing code, maybe we should propose a naming setup for new classes and transition the old ones more carefully and only when things fit perfectly, rather than in bulk. Stuff like IsCallable, or IsChoice (or the Matches... equivalents appear to be no-brainers).

That makes a lot of sense to me. So for this PR this would mean:
- [ ] rename only the no-brainers. I'd say they're ``MatchesCallable``, ``MatchesChoice``, ``MatchesStr``, ``MatchesStrPrefix``, ``MatchesValue`` - any others or different ones?
- [ ] add the naming setup to the developer guide

adswa:
Because this PR has quite a lot of conflicts, and needs to shrink in scope, I plan to reopen it scaled down. But if would be nice if someone can confirm the plan of action:

> That makes a lot of sense to me. So for this PR this would mean:

> - [ ] rename only the no-brainers. I'd say they're MatchesCallable, MatchesChoice, MatchesStr, MatchesStrPrefix, MatchesValue - any others or different ones?
> - [ ] add the naming setup to the developer guide

adswa:
I will close this PR as stale to not let it linger unresolved. I still think that standardizing the constraints to a set of names that reflects whether they only check or only transform or check and transform inputs makes sense - maybe we can reopen such effort soon. I think [this comment](https://github.com/datalad/datalad-next/pull/310#issuecomment-1463561460) still contains a lot of valuable considerations we can keep in mind in a future attempt.

PR #311, opened 2023-03-10
Replace implementation TODO with example how it already works

adswa:
I thought I found a small TODO to help with, but then realized that the code already does what the TODO wants. So instead, I've added an example how to do it.


mih (approved):
Ha, cool! That might have been the oldest todo in the code base. Pretty sure it predates datalad. Thx much!

Issue #261, opened 2023-02-25
has been failing testing in datalad core for abit


yarikoptic:
from a sample run https://github.com/datalad/datalad/actions/runs/4270688403/jobs/7434642131

```
=========================== short test summary info ============================
FAILED annexremotes/tests/test_uncurl.py::test_uncurl_checkurl - ValueError: unsupported URL 'garbledhttps://httpbin.org/status/404'
FAILED annexremotes/tests/test_uncurl.py::test_uncurl_addurl_unredirected - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none annex addurl --file=dummy garbledhttps://httpbin.org/bytes/24 -c annex.dotfiles=true' failed with exitcode 1 under /tmp/pytest-of-runner/pytest-0/test_uncurl_addurl_unredirecte0/ds [out: 'addurl garbledhttps://httpbin.org/bytes/24 
failed'] [err: 'Configuration of annex.security.allowed-url-schemes does not allow accessing garbledhttps://httpbin.org/bytes/24
addurl: 1 failed']
====== 2 failed, 204 passed, 10 skipped, 12 warnings in 492.97s (0:08:12) ======
```

<details>
<summary>Seems started to fail on Feb 23rd</summary> 

```shell
(git)smaug:/mnt/datasets/datalad/ci/logs/2023/02[master]git
$> GIT_TERMINAL_PROMPT=0 datalad foreach-dataset --o-s relpath -r -J10 'git grep "annexremotes/tests/test_uncurl.py::test_uncurl_checkurl"' | grep PASS
07/pr/7273/5852580/github-Extensions-6132-success/6_test (datalad-next).txt:2023-02-07T03:45:45.3941139Z annexremotes/tests/test_uncurl.py::test_uncurl_checkurl PASSED
07/pr/7273/5852580/github-Extensions-6132-success/test (datalad-next)/13_datalad-next tests using pytest.txt:2023-02-07T03:45:45.3940721Z annexremotes/tests/test_uncurl.py::test_uncurl_checkurl PASSED
07/pr/7273/5b40500/github-Extensions-6134-success/6_test (datalad-next).txt:2023-02-07T04:07:46.3473560Z annexremotes/tests/test_uncurl.py::test_uncurl_checkurl PASSED
[...]

PR #313, opened 2023-03-10
Spin up httpbin container via pytest fixture

jwodder:
Part of #261.

### DIFF
--- a/datalad_next/tests/fixtures.py
+++ b/datalad_next/tests/fixtures.py
@@ -1,9 +1,12 @@
 import logging
 import os
 from pathlib import Path
+import subprocess
 import pytest
 from tempfile import NamedTemporaryFile
+from time import sleep
 from unittest.mock import patch
+from urllib.request import urlopen
 
 from datalad_next.datasets import Dataset
 from datalad_next.tests.utils import (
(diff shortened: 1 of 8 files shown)


datalad_next/tests/fixtures.py
```diff
@@ -295,3 +298,57 @@ def http_server_with_basicauth(tmp_path_factory, http_credential):
...
+            else:
+                raise RuntimeError("httpbin container did not start up in time")
+        yield {
+            "standard": ciurl if container_id is not None else hbsurl,
+            "http": hburl,
+            "https": hbsurl,
+        }
+    finally:
+        if container_id is not None:
+            subprocess.run(["docker", "rm", "-f", container_id], check=True)
```
yarikoptic, line 354: May be just add `--rm` to `docker run` or there is benefit?

yarikoptic: Nevermind, it is a process in background, right

mih (approved):
Thank you! Maintains behavior, improves implementation. Much appreciated!

Note to self: Continues to ignore the possibility to use docker on windows. Watch failure pattern with public instance now and add later, if necessary.

mih:
Macos test on appveyor now need protection against missing docker again.

mih (approved):

PR #315, opened 2023-03-12
codespell: config, workflow + fixes

yarikoptic:


mih (approved):
Thx! I was finally able to look into this. Some of the edits I would have done differently, but none of them is worth a discussion.

After merging current `main` the check still passes, so I think we can give this a go.

Thx!

[I canceled the actual CIs]

PR #320, opened 2023-03-20
First draft of a basic deprecation decorator

adswa:
I was inspired by https://github.com/datalad/datalad/issues/6998 to try to write a decorator that could be used to automate deprecation warnings. It can be used to:

- deprecate entire methods:

```
     @staticmethod
     @datasetmethod(name="download")
     @eval_results
+    @deprecated(version='1.0.0', msg='Use <othercommand> instead')
     def __call__(...):
```

- deprecate individual parameters:

```
     @staticmethod
     @datasetmethod(name="download")
     @eval_results
+    @deprecated(version='1.0.0', msg='Use <otherparameter> instead', parameter='force')
     def __call__(...):
```

- deprecate individual parameter values:

```
     @staticmethod
     @datasetmethod(name="download")
     @eval_results
+    @deprecated(version='1.0.0', msg='Use <othermode> instead', parameter='mode', parameter_value='oneofthem')
     def __call__(...):
```

several decorators can be used to deprecate several parameters:
```
     @staticmethod
     @datasetmethod(name="download")
     @eval_results
+    @deprecated(version='1.0.0', msg='Use <othermode> instead', parameter='mode')
+    @deprecated(version='1.0.0', msg='Use <othermode> instead', parameter='force')
     def __call__(...):
```

several parameter choices can be deprecated as one:

```
     @staticmethod
     @datasetmethod(name="download")
     @eval_results
+    @deprecated(version='1.0.0', msg='Use <othermode> instead', parameter='mode', parameter_value=['oneofthem', 'onemode'])
     def __call__(...):
```

Whenever something deprecated is used (e.g., an entire deprecated function/method/class, a deprecated parameter, or a deprecated parameter choice), a DeprecationWarning will be emitted.

I think such a helper can be useful to prevent the need to issue DeprecationWarnings by hand within a command by manually checking specified parameter. What it doesn't do is any replacement, e.g., switching out a deprecated parameter choice against its replacement. I'd be curious to discuss whether such a helpful is useful in general, and whether ``datalad-next`` is the right target.


adswa:
The function has a too high complexity still. I tried an RF that reduced it in length in 37aa46ef4e3db58c, but that had a higher cognitive complexity than the current commit

mih:
I will review this soon, but I can already say that I very much appreciate the convenience that comes with this!

bpoldrack (commented):
In principle, I like the approach.

Edit: Ignore me, I need another pass.

datalad_next/utils/deprecate.py
```diff
@@ -0,0 +1,89 @@
...
+      Individual parameter choice of specified parameter being deprecated
+    """
+
+    base_template = "{func} was deprecated in version {version}. {msg}"
+    if parameter is None:
+        # the entire class/function is deprecated
+        def decorator(func):
+            @wraps(func)
+            def func_with_deprecation_warning(*args, **kwargs):
+                fname = (
```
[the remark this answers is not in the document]
bpoldrack: Ahhh, no. I am sorry. The renaming only applies to the `Dataset` method.

Hence, two things instead:
- include a check, that the class is a `subclass` of `Interface` to make sure what is decorated is a command implementation, so that the logic actually applies
- ideally transform the class name to the command name. That transformation should already exist in `interface`. (I would, however, not consider this point a blocker)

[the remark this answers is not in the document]
bpoldrack: Alternative, since this is making it more complex:

Rip out that logic. When a command is to be deprecated entirely, currently the class needs to be decorated and done. If someone decorates the wrong thing - not the problem of this decorator, I guess.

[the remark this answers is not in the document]
bpoldrack: I keep thinking about this. The issue with the simplest (the last) suggestion is, that a class deprecation would be triggered by importing the class, not by calling it. I wonder, whether that's actually a good thing. It could lead to the next step of forcing us to make imports leaner. If the deprecator would register that class in a singleton - say `datalad._depracted`, we could probably implement a delayed import into api. Just like extensions aren't necessarily imported on every datalad start up.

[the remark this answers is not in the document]
adswa: Thanks for your thoughts, although I'm not 100% certain I can follow them :)
Just for completeness, here is the current state:

> The parameter value overwrite-existing of parameter force of <class 'datalad_next.commands.download.Download'> was deprecated in version 1.0.0.

An easy alternative could also be to leave any messaging entirely to the user. So if you want to deprecate one parameter option, the ``msg`` could become "The option ABC of parameter D is deprecated. Use EFG instead". And then we wouldn't do any templating, but simply pass down the message into the deprecation warning

datalad_next/utils/deprecate.py
```diff
@@ -0,0 +1,89 @@
+import inspect
+import warnings
+from functools import wraps
+
+from datalad_next.utils import get_wrapped_class
+
+
+def deprecated(msg, version, parameter=None, parameter_choice=None):
```
mih, line 8:
```suggestion
def deprecated(msg, *, version, parameter=None, parameter_choice=None):
```

I presently consider it better to make the nature of the version declaration more obvious in the resulting code. It is already mandatory (which is essential, I think).

A custom message is also useful to be mandatory. Deprecation without guidance on adjustment is bad practice.

Given that there are two essential items, it would be great to have the different semantics be obvious in each declaration.

datalad_next/utils/deprecate.py
```diff
@@ -0,0 +1,89 @@
...
+    """Annotate functions, classes, or function parameters with standardized
+    deprecation warnings.
+
+    Parameters
+    ----------
+    version: str
+      Software version number at which the deprecation was made
+    msg: str
+      Custom message to append to a deprecation warning
```
mih, line 17:
```suggestion
    msg: str
      Custom message to append to a deprecation warning
    version: str
      Software version number at which the deprecation was made
```

Reordered to match signature.

datalad_next/utils/deprecate.py
```diff
@@ -0,0 +1,89 @@
...
+    parameter: str
+      Individual parameter being deprecated (instead of entire function/class)
+    parameter_choice: str
```
mih, line 20: I am wondering whether `value` would be better than `choice`. This will often be used for choices, but the implementation uses an equality test, so it would work for anything else too.

The type declaration seems to be incomplete, more than `str` is supported. See below.

datalad_next/utils/deprecate.py
```diff
@@ -0,0 +1,89 @@
...
+      Individual parameter choice of specified parameter being deprecated
+    """
+
+    base_template = "{func} was deprecated in version {version}. {msg}"
+    if parameter is None:
```
mih, line 25: It would make sense to me to move the whole body of this conditional into a dedicated helper function.

I think the two inline `decorator` functions could also become standalone functions. I might give this a try later.

datalad_next/utils/deprecate.py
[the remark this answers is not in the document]
mih: I have difficulties grasping the issue here. I think it would make sense to actually test the behavior that is discussed here and make it obvious, and verified. Afterwards it would be easier to debate whether it is problematic.

mih (commented):
Thank you for taking this on! I left a bunch of comments and suggestions.

mih:
Looking deeper into the code, I want to record that a feature provided here, but not in https://pypi.org/project/Deprecated/ is the possibility to deprecate individual parameters of a callable (and even particular parameter values). This should be pointed out in the documentation.

datalad_next/utils/deprecate.py
mih: Simplified and merged.

datalad_next/utils/deprecate.py
mih: Adjusted and type-annotated

mih:
I made a number of changes to the code. This was easy, because the tests are great! Thx much!

adswa:
This is really cool! Thanks, @mih!

bpoldrack:
@adswa 
> Do you think that 37aa46e would do the refactoring in a similar way to what you had in mind, or would it maybe be a better place to start a refactoring from?

Better place! However, I suppose this commit is superceded by what mih has already done.

mih (approved):
Thanks again for the PR. I think we can go ahead now.

PR #324, opened 2023-04-20
New EnsureRemoteName Constraint

### LINKED ISSUE #193: Provide `EnsureRemoteName`

adswa:
This PR is an initial approach to fix #193. It adds a new Constraint ``EnsureRemoteName`` that derives from ``EnsureGitRefName``. The Constraint checks that a sibling name is given, and, optionally, can check if the sibling already exists or not using the boolean ``existing`` flag. If ``existing=True`` the remote must exist, if ``existing=False`` it shouldn't exist. The Constraint can be tuned to a specific dataset.

Let me know if something like this is what you had in mind for #193.

### DIFF
--- a/datalad_next/constraints/__init__.py
+++ b/datalad_next/constraints/__init__.py
@@ -85,3 +85,8 @@ from .formats import (
     EnsureURL,
     EnsureParsedURL,
 )
+
+from .git import (
+    EnsureGitRefName,
+    EnsureRemoteName
+)
--- a/datalad_next/constraints/git.py
+++ b/datalad_next/constraints/git.py
@@ -81,21 +81,26 @@ class EnsureGitRefName(Constraint):
 
 
 class EnsureRemoteName(EnsureGitRefName):
-    """Ensures a remote name is provided, and if it fulfills optional
-    requirements"""
+    """Ensures a remote name is given, and optionally if such a remote is known
+    """
 
     def __init__(self,
-                 exists: bool | None = None,
+                 known: bool | None = None,
                  dsarg: DatasetParameter | None = None):
         """
         Parameters
         ----------
-        exists: bool, optional
-           If true, validates that the remote exists, fails otherwise.
-           If false, validates that the remote doesn't exist, fails otherwise.
-           If None, just checks that a sibling name was provided.
+        known: bool, optional
+           By default, a given value is only checked if it is a syntactically
+           correct remote name.
+           If ``True``, also checks that the given name corresponds to a
+           known remote in the dataset given by ``dsarg``. If ``False``,
+           checks that the given remote does not match any known remote
+           in that dataset.
+        dsarg: DatasetParameter, optional
+           Identifies a dataset for testing remote existence, if requested.
         """
-        self._exists = exists
+        self._known = known
         self._dsarg = dsarg
         super().__init__(allow_onelevel=True,
                          refspec_pattern=False)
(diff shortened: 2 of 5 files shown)


datalad_next/constraints/git.py
```diff
@@ -66,3 +69,69 @@ def short_description(self):
...
+    """Ensures a remote name is provided, and if it fulfills optional
+    requirements"""
+
+    def __init__(self,
+                 existing: bool | None = None,
+                 dsarg: DatasetParameter | None = None):
+        """
+        Parameters
+        ----------
+        existing: bool
```
christian-monch, line 84: This is just nitpicking ;-) 

I find the name `existing` and the logic of `True` and `False` slightly non-intuitive. Because to me I would think that `existing = True` means: "check for existing", and `existing = False` means: "don't check for existing".

But here `existing = False` means: "don't check for existing and check for not existing". And `existing = True` means: "don't check for not existing, but check for existing". I see the value in the checks. I understand that they are mutually exclusive. 

An (obviously useless) name that reflects the semantics seems to be: `if_not_none_and_true_check_existence_if_not_none_and_false_check_none_existance`.

An option might be to separate the request for checks and the type of check. E.g.,

- perform_check: bool: check the validation criteria if True, else don't check it.
- validation_criteria: str: one of "existance", "non-existance": check for .... if `perform check` is True.

As I said "nitpicking" ;-)

datalad_next/constraints/git.py
```diff
@@ -66,3 +69,69 @@ def short_description(self):
...
+            # simple, do here
+            raise ValueError('must state a sibling name')
+        super().__call__(value)
+
+        if self._existing is None:
+            # we only need to know that something was provided, no further check
+            return value
+
+        from datalad.runner import GitRunner, StdOutCapture
+        from datalad_next.exceptions import CommandError
```
christian-monch, line 105: `CommandError` is unused

datalad_next/constraints/tests/test_special_purpose.py
```diff
@@ -52,6 +54,35 @@ def test_EnsureGitRefName():
         'refs/heads/*') == 'refs/heads/*'
 
 
+def test_EnsureRemoteName(existing_dataset):
+    # empty sibling name must raise
+    with pytest.raises(ValueError):
+        c = EnsureRemoteName()('')
```
christian-monch, line 60: `c` is not used afterward.

christian-monch (changes requested):
A few comments w.r.t to `existing`.

datalad_next/constraints/git.py
christian-monch: (nitpick 2.0) another option would be a parameter named `validation` with values "nothing", "existence", or "non-existence".

adswa:
Thanks for the review, @christian-monch! I'll act on it, I promise, but I'm still waiting for @mih's verdict whether the constraint in general is what he had in mind with the issue.

datalad_next/constraints/__init__.py
```diff
@@ -51,3 +51,8 @@
     EnsureURL,
     EnsureParsedURL,
 )
+
+from .git import (
+    EnsureGitRefName,
+    EnsureRemoteName
+)
```
mih, line 58: I am still not sure which constraints should be imported at the toplevel. You seem to assume "all of them". I tend to think that too, sometimes. But then, any additional import takes time, and that time will need to be spent unconditionally, regardless of which constraint is actually needed. In this particular case the cost should be negligible.

mih (commented):
Here is my general comment: I think this is what I was aiming for, indeed. I am presently not well in touch with the whole constraint-topic, though.

It would be best, if this PR could already kook this into a suitable command. I believe the cheapest target would be `create_sibling_webdav` which has `name` and `storage_name` that both have `EnsureStr` right now. In the simplest case, this new one would replace that.

datalad_next/constraints/git.py
```diff
@@ -66,3 +69,69 @@ def short_description(self):
...
         )
+
+
+class EnsureRemoteName(EnsureGitRefName):
+    """Ensures a remote name is provided, and if it fulfills optional
+    requirements"""
+
+    def __init__(self,
+                 existing: bool | None = None,
+                 dsarg: DatasetParameter | None = None):
```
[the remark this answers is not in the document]
adswa: Thanks for catching this! 
I want to keep an explicit None to opt out of checks whether the remote exists or not, and just validate that a sibling name was given. I don't think we have any other check for this in constraints, and looking around, it seems like this task was mostly fulfilled by an ``EnsureStr``. But I am adjusting line 116 :+1:

datalad_next/constraints/git.py
adswa: alternative suggestion: what about ``pre-exists``?

datalad_next/constraints/git.py
```diff
@@ -66,3 +69,69 @@ def short_description(self):
...
+        """
+        Parameters
+        ----------
+        existing: bool
+           If true, validates that the remote exists, fails otherwise.
+           If false, validates that the remote doesn't exist, fails otherwise.
```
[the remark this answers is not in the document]
adswa: thx!

datalad_next/constraints/git.py
adswa: good catch!

adswa:
updated the branch

datalad_next/constraints/git.py
adswa: ping

mih:
Thanks for the changes!

Re naming of `existing` and now `preexists`: I think this is a bit overcomplicated thinking. My suggestion would be to call it simply `exists`. This is also what `EnsurePath(lexists=)` is and does, with identical argument semantics. I see no reason to implement a different logic or naming semantic here. I will push a corresponding change myself in a bit.

mih:
Playing with the code I believe that it should be done with a stricter handling of error conditions. When `exists` is not `None`, it should not tolerate an unset `dsarg` and run opportunistically in the CWD. It should fail. I will try to implement this now

mih:
OK, I ended up making more changes. I believe it is better now, but I had to rediscover some things, so this is a bit of a shot from the hip.

mih:
Meeged as #577

PR #329, opened 2023-05-03
Make special remote processes report progress logs to git-annex

### LINKED ISSUE #328: Let `uncurl` download progress reach git-annex
I am pulling a file via https and `datalad get` from the ICF data store (800M). Not a single progress update.

Pulling the same file from the same URL with `datalad download` (not `download-url`!) works fine and shows progress.

This is a serious issue, because progress reporting is necessary to avoid triggering annex stall detection -- which could easily happen for large-file downloads.

mih:
Add dedicated log handler is attached to both the datalad "root" logger and the special remote class instance. The handler gets a filter to only makes it see progress log records (this is something that also the main progress handler should be doing). In the generic main() of special remotes, this log setup modifications is added as an additional log output -- no interference with any existing setup.

This sketch is limited/incomplete in that it assumes only a single progress tracker to be used (although it is already set up to support more than one). So in case of multiple concurrent trackers, it would send rubbish progress logs to git-annex.

That being said, with this change we see git-annex native progress reporting with git annex get and datalad native progress reporting with datalad get.

Closes #328

### DIFF
--- a/datalad_next/patches/__init__.py
+++ b/datalad_next/patches/__init__.py
@@ -8,4 +8,5 @@ from . import (
     push_optimize,
     siblings,
     test_keyring,
+    customremotes_main,
 )
--- a/datalad_next/patches/customremotes_main.py
+++ b/datalad_next/patches/customremotes_main.py
@@ -1,0 +1,136 @@
+"""Connect ``log_progress``-style progress reporting to git-annex
+
+This patch introduces a dedicated progress log handler as a proxy between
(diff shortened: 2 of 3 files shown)


christian-monch (approved):
Looks good to me

mih:
Via @mslw:

> Checking out that PR made a progress bar appear for me.

mih:
It appear that all test failures are unrelated and also happen in other recent test runs.

Issue #331, opened 2023-05-04
FAILED test_clone.py::test_autoenabled_remote_msg


mih:
This is tripping the datalad-core test battery. It was fixed in datalad-core with https://github.com/datalad/datalad/pull/7367, but there has been no release.

We could consider switching the version of datalad-core to be tested here to the most recent state of its `maint`. This would be useful for development, but not for checking whether datalad-next's dependency declaration is fit-for-release.

In general https://github.com/datalad/datalad-extensions should provide insight, but it does not test the released datalad-core, and we only know about the continued failure in a released version, because here we test the latest datalad-core release (with datalad-next enabled).

adswa:
I think one possible way to tackle this is to adopt a faster release cycle in datalad. Yarik mentioned it in the past, pointing to how they do it in DANDI - essentially with every merged PR.

mih:
0.18.4 is out. We can undo to tmp fix.

PR #332, opened 2023-05-04
(Temporarily) use the latest `maint` state of datalad-core for testing

mih:
See https://github.com/datalad/datalad-next/issues/331


adswa (commented):
I would be fine with this, at least as a temporary means to make development for the current sprint easier

adswa (approved):

mih:
So this move seems to fix all CI issues. Going ahead with it as a quick remedy.

PR #339, opened 2023-05-05
Configuration-related updates

### LINKED ISSUE #325: Adjust `ConfigManager` to post overrides into the ENV
Overrides can come in via the CLI, or be directly set in a manager instance. Given that overrides are somewhat global, it makes sense to post them in the processes ENV too.

Here is a concrete example of where that would be useful https://github.com/psychoinformatics-de/knowledge-base/issues/15#issuecomment-1522907177 (instead of having to go git-native, a declaration via `datalad -c ...` would be sufficient).

This would also make a runner-specific change to pass on configuration, as proposed in https://github.com/datalad/datalad/pull/7344 (see issue https://github.com/datalad/datalad/issues/7352) unnecessary. While this would work as a solution for the clone issue linked above too, it would be limited to execution via the datalad runner. Simply posting overrides in the process ENV via `GIT_CONFIG_COUNT= GIT_CONFIG_KEY_ GIT_CONFIG_VALUE_` would be applicable to any means of executing subprocesses (e.g., plain `subprocess.run` -- something that is even done in the datalad codebase).

Issues this would address

- https://github.com/datalad/datalad/issues/3456 (closed but unresolved)
- https://github.com/datalad/datalad/issues/4119

#### Interference with Git

If the Git-native variables are already present in the environment (see above), we need to ensure that any DataLad modification amends the existing setup and does not strip settings.

#### Override semantics

The nature of "overrides" in the datalad configuration is underspecified. Relevant statements from the documentation are:

- environment variables can be used to override any datalad configuration, and have precedence over any other configuration scope (any further elaboration is specific to `DATALAD_*` type variables.
- re `DATALAD_CONFIG_OVERRIDES_JSON`: f both individual configuration variables *and* JSON-overrides are used, the former take precedent over the latter, overriding the respective *individual* settings from configurations declared in the JSON-overrides
- the `configuration` command is not able to query overrides separately
- `ConfigManager` docs say: 'override' limits the modification to the `ConfigManager` instance, and the assigned value overrides any setting from any other source [remark: the use of "instance" is a concern here]
- `ConfigManager` declares: Each instance carries a public `overrides` attribute. This dictionary contains variables that override any setting read from a file. The overrides are persistent across reloads [remark: limitation to "from a file" is relevant here]
- CLI config overrides (`datalad -c`) become `ConfigManager` instance overrides

#### Potential inconsistencies

- on `ConfigManager.reload()` instance overrides are applied unconditionally to the "merged" config. But overrides from ENV only when `scope` is not `branch`.  `ConfigManager.get_from_source()` implemented the same logic.
- `ConfigManager` runs `git config`. If `GIT_CONFIG_COUNT= GIT_CONFIG_KEY_ GIT_CONFIG_VALUE_` are declared, this should apply these git-native overrides, but datalad "sees" them. It needs to be checked whether scope associations are inferred properly. Possibly, `ConfigManager` could afterwards simply replace these settings, because it did already "read them in".
  - No, that is not possible. The settings are "seen" (git-config reports them as origin='command line' , and the config manager umps them all into a `git` category. At this point they are no longer detectable as overrides. If the git-nativ declaration is then replaced, any sub-process will no longer get to see them, because they only live in the parent process's `ConfigManager`
- Overrides from ENV are not reported by `ConfigManager.repr()` (only instance overrides)

#### Threats

If the implementation is changed to post config overrides in git-native format into the ENV, we blur the distinction between ENV-overrides (cannot be manipulated via the `ConfigManager`, are applied differently than instance overrides), and instance/CLI overrides. Essentially, any instance override would affect the ENV. However, posting in Git-native format does not simply remove the distinction of instance and ENV overrides (although that may be desirable, given the complexity increase).

It remains to be evaluated, how git-native overrides interact with the attribution of origin of a configuration. Importantly, Git distinguished a dedicated scope `command` for this (separate from `system`, `global`, `local`, `worktree` (which would be most similar to our `branch`)). For this `command` scope, git-config says:

> These environment variables will override values in configuration files, but will be overridden by any explicit options passed via `git -c`

This is both similar and different from our notion of overrides. Yes, they are overrides, but CLI setting has the last word, whereras in datalad, we have ENV override CLI.

The latter could be aligned by the datalad CLI resorting to posting its `-c` setting via the Git-native mechanism into the ENV. This would give them the highest precedence for datalad-internal processing too. It would also allow for a declaration of any `datalad -c X`, in exactly the same way and with the same semantics as `git -c X`, and also result in any internal git call to behave as if called in that fashion.

### LINKED ISSUE #337: Add cannonical import for `ConfigManager` class
This is an essential component, and we need the import, at minimum, for type annotation.

mih:
- Closes #337
- Closes #325

TODO
- [ ] Add high-level test, ideally also for the CLI
- [ ] Amend patch commit with proper explanation


mih:
This PR has been dissolved into its mutually independent change set. The key aspect is now addressed by #395

PR #342, opened 2023-05-08
Draft for Issue 183 list collection

christian-monch:
A draft for issue #183

-[ ] add credential management
-[ ] add tests

Issue #323, opened 2023-04-19
Consolidate "traverser" from gooey and metalad into a common implementation


mih:
This implementation should be layered, to be useful for listing, but also status/diff reporting.

It should also be benchmarked, relative to git-lsdir and and git-ls-tree.

Ping @christian-monch

christian-monch:
Will get to that in July

mih:
Looking into this further, it seems that it might be more meaningful to replace the implementation of change-detection in gooey's `status-light` which and iterator around `git diff-index HEAD` (instead of the present `diff-files`).

mih:
#580 now also adds `iter_gittree()` to the list of available iterators.

mih:
I have implemented a draft of `iter_gitdiff()` now.

PR #343, opened 2023-05-08
`iter_collections` and `ls-file-collection`

### LINKED ISSUE #183: Implement alternative to `add-archive-content`
**Note: Lots of thoughts here. The latest variant is https://github.com/datalad/datalad-next/issues/183#issuecomment-1532935562**

This is one of the oldest datalad command. It core purpose is to register special URL for the `datalad-archive` special remote that indicate that particular keys can be obtained by extracting a file from an archive that is also registered as an annexed file in the dataset.

On top of that, this command is also a front-end to archive extraction and content re-layouting.

It would be good this disentangle these aspects and provide a more targeted (and likely simpler alternative).

Concept

Parse an archive (already an annexed file), run `git annex calckey` on each contained file, add a `datalad-archive` URL to the computed key, if it exists.

This command would not interact with a dataset's worktree. Consequently, it needs no parameterization re archive content re-layouting. It should should work with bare repos.

From a user perspective, the composition of a dataset's content will be a more independent process. A user could extract files from an existing archive any way they see fit, rename, filter, restructure by any means and/or tools. Importantly, it does not matter whether an archive is the actual starting point of the operation, or whether an archive (with any structure) was just created from existing dataset content (think ORA dataset archive), for the purpose of depositing this archive only (rather than individual files) on some storage system.

### LINKED ISSUE #348: `str(CommandParametrizationError(...))` with with `KeyError`
```py
ipdb> p exceptions
{ParameterConstraintContext(parameters=('type', 'collection', 'hash'), description='collection iterator'): ConstraintError(LsFileCollectionParamValidator(), 'http://mike.com', '{type} collection requires a Path-type identifier', {'type': 'tarfile'})}

ipdb> e = CommandParametrizationError(exceptions)

ipdb> str(e)
*** KeyError: 'collection'
```

mih:
This PR is making `add-archive-content` (largely) obsolete. Closes #183 

Summary

- new module `iter_collections` to establish a location in the spirit of #323 . The idea is to use dataclasses instead of result records, and enums instead of string labels to report on collection content. Iterators also support hashing (as means to generate identifiers.
  - import `iterdir()` from `datalad-gooey` as an iterator for a directory as a collections (of directory content)
  - implemented `itertar()` for tar file archives -- reporting aligned with `iterdir()`
- `ls-file-collection` command that can drive these iterators.
  - factors any glue code necessary into a custom validator 
  - converts collection iterator items into result records and yields them

This is somewhat in conflict with #342 due to suboptimal communication. A future discussion needs to sort out how the two approaches can be consolidated.

TODO:

- [x] we need to be able hash collection items, ideally using a configurable algorithm. The command draft prepares for that. however, it is still unclear whether this would be done in a collection iterator (pretty strong requirement), or within the command itself (complicated to achieve, because it would require duplicating any possibly necessary data transport logic between the iterator implementation and the command). That being said, one could envision remote collections that can readily provide a hash -- for those it would make most sense to let the iterator report them. So maybe a `hash` parameter of `ls-file-collection` should be an instruction what to compute, if there is no hash already. If so, an iterator would no only need to report a hash, but also the type of the hash it is reporting.
  It could be a sensible middle-ground to have an iterator provide a zero-argument callable that provides a file-like that generic hash-computation can use, if the controlling command deems it necessary (e.g. a requested hash is not readily provided).
  That being said, it is unlikely that a generic implementation of a read-from-some-file-like is going to be optimal. Considering the bandwith of possibilities and necessities for `download` (i.e. anything remote; streaming, chunks size, etc) alone cries for configurability. So maybe going down the path of `UrlOperations` which can hash on the go is a better approach. Maybe factor that out into a dedicated class, and simply make it easy to use...
- [x] for #183 we need an iterator for archives (we could have two, a simple one for local archives, and a more sophisticated for remote archives (see #342) -- but they would not necessarily differ much in the end, if hashing must be perform -- rather than being able to report an existing hash.https://doi.org/10.1038/s41597-022-01163-2
- [x] #342 mentions auto-detection of collection types. I initially had the same idea, but then could not come up with a use case where one would want to (surrender to) auto-detect. What `auto` would do could/would change over time. Conflict would always be there (even if a had a gitrepo worktree, I might still just need a directory listing). `auto` could maybe provide a bit of perceived convenience for interactive use, but that convenience would need to be bough with added complexity of auto-detection and for preserving its outcome, long-term. Not worth it IMHO.
- [x] legit error on windows (symlink target looks broken)
```
       FAILED ..\iter_collections\tests\test_iterdir.py::test_iterdir - AssertionError: assert IterdirItem(path=WindowsPath('C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/symlink'), type=<PathType.symlink: 'symlink'>, symlink_target=WindowsPath('//?/C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/some_dir/file_in_dir.txt')) in [IterdirItem(path=WindowsPath('C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/random_file1.txt'), type=<PathType.file: 'file'>, symlink_target=None), IterdirItem(path=WindowsPath('C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/some_dir'), type=<PathType.directory: 'directory'>, symlink_target=None), IterdirItem(path=WindowsPath('C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/symlink'), type=<PathType.symlink: 'symlink'>, symlink_target=WindowsPath('C:/DLTMP/pytest-of-appveyor/pytest-0/dir_tree0/some_dir/file_in_dir.txt'))]
```
This seems to be related to the following doc snippet

> Changed in version 3.8: Added support for directory junctions, and changed to return the substitution path (which typically includes \\?\ prefix) rather than the optional “print name” field that was previously returned.

https://bugs.python.org/issue42957

- [x] report file sizes for all collection types
- [x] Stop using `StrEnum` -- only available from Python 3.11 onwards
- [x] Bring test coverage of `iter_collections` to 100%
- [x] Add tests for `ls-file-collection`
- [x] Fully document `iter_collections`
- [x] Document `ls-file-collection`. Include a demo on replacing `add-archive-content`
- [x] Make decision on representation of hardlinks (see TODO in `test_replace_add_archive_content`)
- [x] Something not correct with path type from itertar on windows yet
- [x] Add custom result renderer

Addressed issues

- Closes #348

### DIFF
--- a/datalad_next/__init__.py
+++ b/datalad_next/__init__.py
@@ -39,6 +39,10 @@ command_suite = (
         (
             'datalad_next.commands.download', 'Download', 'download',
         ),
+        (
+            'datalad_next.commands.ls_file_collection', 'LsFileCollection',
+            'ls-file-collection',
+        ),
     ]
 )
 
--- a/datalad_next/commands/__init__.py
+++ b/datalad_next/commands/__init__.py
@@ -52,7 +52,7 @@ class ValidatedInterface(Interface):
     should either be removed, or moved to the corresponding entry in
     ``_validator_``.
     """
-    _validator_ = None
+    _validator_: EnsureCommandParameterization | None = None
 
     @classmethod
     def get_parameter_validator(cls) -> EnsureCommandParameterization | None:
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -1,0 +1,265 @@
+"""
+"""
+
+from __future__ import annotations
+
+__docformat__ = 'restructuredtext'
+
+from dataclasses import (
+    asdict,
+    dataclass,
+)
+from logging import getLogger
+from pathlib import Path
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+)
+
+from datalad_next.commands import (
+    EnsureCommandParameterization,
+    ValidatedInterface,
+    Parameter,
+    ParameterConstraintContext,
+    build_doc,
+    eval_results,
+    get_status_dict,
+)
+from datalad_next.constraints import (
+    EnsureChoice,
+    EnsurePath,
+    EnsureURL,
+)
+from datalad_next.uis import (
+    ansi_colors as ac,
+    ui_switcher as ui,
+)
+from datalad_next.utils import ensure_list
+
+from datalad_next.iter_collections.directory import iterdir
+from datalad_next.iter_collections.tarfile import itertar
+from datalad_next.iter_collections.utils import FileSystemItemType
+
+
+lgr = getLogger('datalad.local.ls_file_collection')
+
+
+# hand-maintain a list of collection type names that should be
+# advertized and supported. it makes little sense to auto-discover
+# them, because each collection type likely needs some custom glue
+# code, and some iterators may not even be about *file* collections
+_supported_collection_types = (
+    'directory',
+    'tarfile',
+)
+
+
+@dataclass  # sadly PY3.10+ only (kw_only=True)
+class CollectionSpec:
+    """Internal type for passing a collection specification to
+    ``ls_file_collection``. it is created by the command validator
+    transparently.
+    """
+    orig_id: Any
+    iter: Iterator
+    item2res: Callable
+
+
+class LsFileCollectionParamValidator(EnsureCommandParameterization):
+    """Parameter validator for the ``ls_file_collection`` command"""
+    _collection_types = EnsureChoice(*_supported_collection_types)
+
+    def __init__(self):
+        super().__init__(
+            param_constraints=dict(
+                type=self._collection_types,
+                collection=EnsurePath(lexists=True) | EnsureURL(),
+                # TODO EnsureHashAlgorithm
+                # https://github.com/datalad/datalad-next/issues/346
+                #hash=None,
+            ),
+            joint_constraints={
+                ParameterConstraintContext(('type', 'collection', 'hash'),
+                                           'collection iterator'):
+                self.get_collection_iter,
+            },
+        )
+
+    def get_collection_iter(self, **kwargs):
+        type = kwargs['type']
+        collection = kwargs['collection']
+        hash = ensure_list(kwargs['hash'])
+        iter_fx = None
+        iter_kwargs = None
+        if type in ('directory', 'tarfile'):
+            if not isinstance(collection, Path):
+                self.raise_for(
+                    kwargs,
+                    "{type} collection requires a Path-type identifier",
+                    type=type,
+                )
+            iter_kwargs = dict(path=collection, hash=hash)
+            item2res = fsitem_to_dict
+        if type == 'directory':
+            iter_fx = iterdir
+        elif type == 'tarfile':
+            iter_fx = itertar
+        else:
+            raise RuntimeError('unhandled condition')
+        assert iter_fx is not None
+        return dict(
+            collection=CollectionSpec(
+                orig_id=collection,
+                iter=iter_fx(**iter_kwargs),
+                item2res=item2res),
+        )
+
+
+def fsitem_to_dict(item) -> Dict:
+    keymap = {'name': 'item'}
+    # FileSystemItemType is too fine-grained to be used as result type
+    # directly, map some cases!
+    fsitem_type_to_res_type = {
+        'specialfile': 'file',
+    }
+
+    # TODO likely could be faster by moving the conditional out of the
+    # dict-comprehension and handling them separately upfront/after
+    d = {
+        keymap.get(k, k):
+        # explicit str value access, until we can use `StrEnum`
+        v if k != 'type' else fsitem_type_to_res_type.get(v.value, v.value)
+        for k, v in asdict(item).items()
+        # strip pointless symlink target reports for anything but symlinks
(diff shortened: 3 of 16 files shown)


jsheunis:
I lack the experience and understanding wrt this functionality to comment specifically on the thought process or implementation, but just noting for the record that I ran a test locally and all seems to work nicely:

```
# create dataset with multi-level files

$ datalad create lscollection_test
$ mkdir lscollection_test/subdir
$ echo test1 > lscollection_test/topfile.txt
$ echo test2 > lscollection_test/subdir/subfile.txt
$ datalad save

# run command on root dir

$ datalad -f json ls-file-collection directory lscollection_test

{"action": "ls_file_collection", "collection": "lscollection_test", "item": "lscollection_test/subdir", "status": "ok", "type": "directory"}
{"action": "ls_file_collection", "collection": "lscollection_test", "item": "lscollection_test/topfile.txt", "status": "ok", "symlink_target": ".git/annex/objects/1Z/Z0/MD5E-s6--3e7705498e8be60520841409ebc69bc1.txt/MD5E-s6--3e7705498e8be60520841409ebc69bc1.txt", "type": "symlink"}
{"action": "ls_file_collection", "collection": "lscollection_test", "item": "lscollection_test/.gitattributes", "status": "ok", "type": "file"}
{"action": "ls_file_collection", "collection": "lscollection_test", "item": "lscollection_test/.git", "status": "ok", "type": "directory"}
{"action": "ls_file_collection", "collection": "lscollection_test", "item": "lscollection_test/.datalad", "status": "ok", "type": "directory"}

# run command on sub dir

$ datalad -f json ls-file-collection directory lscollection_test/subdir

{"action": "ls_file_collection", "collection": "lscollection_test/subdir", "item": "lscollection_test/subdir/subfile.txt", "status": "ok", "symlink_target": "../.git/annex/objects/fx/MG/MD5E-s6--126a8a51b9d1bbd07fddc65819a542c3.txt/MD5E-s6--126a8a51b9d1bbd07fddc65819a542c3.txt", "type": "symlink"}
```

mih:
Brief look at performance. Comparing the envisioned CLI use for a TAR file to get all infos needed to satisfy `datalad addurls`

```
datalad -f json ls-file-collection --hash md5 tarfile <TARFILE>
```

with the conventional equivalent (actually not, but in practice it would be possible to quickly grab the same info):

```
tar --to-command=md5sum -xf <TARFILE>
```

Test case is a 2.5GB `tgz` with 1470 files inside.

`datalad ls-file-collection`:

```
real        6.981       0.384       6.688       6.732       7.523       
user        6.366       0.445       5.935       6.187       6.978       
sys         0.584       0.133       0.452       0.535       0.765       
```

`tar + md5sum`:
```
real        14.247      0.730       13.535      13.954      15.251      
user        17.563      0.931       16.752      17.072      18.867      
sys         2.766       0.246       2.556       2.630       3.111
```

Promising!

mih:
Tests are good now, the failure is httpbin accessibility

mih:
Few more bits on performance. This time on the `directory` collection type. On a single directory with 36k files (I think that is a reasonable number of a directory with "a lot of files":

```
In [8]: timeit r = dl.ls_file_collection('directory', '/home/mih/dcm/study_20140724_095117357
   ...: ', result_renderer='disabled')
1.36 s ± 7.43 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: timeit !ls -l ~/dcm/study_20140724_095117357 > /dev/null
439 ms ± 62.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

With so many files, offset costs for argument validation etc. no longer matter much. Speed is about 1/3 of a plain `ls`.

This is not so bad for a first attempt IMHO.

jsheunis:
_(this comment disappeared from github when I posted it, so sorry if it's duplicated)_

I just realised that `ls-file-collection` can be a great lean replacement of `meta-extract` or `meta-conduct` when it comes to supplying metadata of the whole file tree of some file hierarchy to `datalad-catalog`. Obviously it provides somewhat different information, but some scripting can fill the gaps of what `datalad-catalog` needs and what `ls-file-collection` does not (currently) provide. And the bump in performance and ability to provide information on non-datalad datasets (or non-git repos) are major gains!

mih:
Hey @jsheunis @mslw just briefly confirming that this kind of reuse is indeed not only in scope, but more or less the focus. However, I want to point out that `ls-file-collection` is primarily a convenience "driver" for the underlying iterator. As can be seen from the benchmark above, there is a substantial speed difference between the command API and the iterator. This is something that continued to bug me about how I wrote `status` and `diff`. So I would suspect that within-datalad reuse will focus on the iterator, not `ls-file-collection` (which is still useful for CLI scripting).

That being said, the iterators can be made more customizable, once it is clear how it would make sense. However, I would prefer and aim for multiple, clear, and focused implementations, rather than sophisticated swiss-army-knifes.

Next up would be git-focused iterators

- #349 
- #350

They should make it possible to drop a substantial amount of custom code from gooey.

I did start thinking about annex-focused iterators, but I have not concluded on a concept yet (will file an issue once I know something). My current thinking is exploring the possibility of wrapping `annex find`.

jsheunis:
Thanks for the explanation @mih, it gives me better context.

PR #345, opened 2023-05-08
New `MultiHash` class

### LINKED ISSUE #344: Factor out multi-hash helper from `UrlOperations`
It hides in `_get_hasher`, `_get_hash_report` and too many duplicate loops. It could become a proper self-contained class, and also help in #343

mih:
This has factored been factored out from the `UrlOperations` classes. It simplifies that code and can also be used elsewhere.

Closes #344


yarikoptic:
FWIW name `MultiHash` might be a bit suboptimal since collides with a known https://github.com/multiformats/multihash .

Note that in datalad core we have `Digester` for pretty much the same functionality: https://github.com/datalad/datalad/blob/HEAD/datalad/support/digests.py#L25

mih:
Yes, I am aware of `Digester`, but it focuses on file-likes, hence it could not be used for the `UrlOperations`. This PR only took that original implementation and generalized and moved it into its own piece, because #343 also needs this.

Thanks for pointing out the name clash. I was not aware of this. Right now I think that a utility class conflicting with a package is not optimal indeed, but also not a major problem.

Issue #346, opened 2023-05-09
Add `EnsureHashAlgorithm`


mih:
There is detection logic in `MultiHash` that could/should become exposed in this more standard fashion. For commands that support it (`download`, `ls-file-collection`) it would also make sense to implement `EnsureMultiHash`

mslw:
`MultiHash` uses hashlib, and hashlib provides constant attributes `hashlib.algorithms_guaranteed` ([doc](https://docs.python.org/3/library/hashlib.html#hashlib.algorithms_guaranteed)) and `hashlib.algorithms_available` (both sets) - seems like a perfect fit.

mslw:
TIL from [hashlib docs](https://docs.python.org/3.11/library/hashlib.html):

> `hashlib.algorithms_available`: A set containing the names of the hash algorithms that are available in the running Python interpreter. These names will be recognized when passed to `new()` [but have no named constructors in hashlib -- mslw]
> 
> `hashlib.new` Is a generic constructor that takes the string name of the desired algorithm as its first parameter. It also exists to allow access to the above listed hashes as well as any other algorithms that your OpenSSL library may offer. The named constructors are much faster than `new()` and should be preferred.

So with the way `MultiHash` currently works (`hr = getattr(hashlib, h.lower(), None)`), we need to check against `algorithms_guaranteed`.

We could also extend `MultiHash` to use the additional ones through `new()` if not found in attributes. However, we probably care much less about the non-guaranteed algorithms. This is what's guaranteed and additionally available on my laptop:

``` pycon
>>> import hashlib
>>> hashlib.algorithms_guaranteed
{'shake_256', 'sha224', 'sha3_512', 'md5', 'sha3_256', 'sha3_224', 'shake_128', 'sha3_384', 'sha512', 'sha384', 'blake2s', 'blake2b', 'sha1', 'sha256'}
>>> hashlib.algorithms_available - hashlib.algorithms_guaranteed  # set difference
{'sm3', 'md4', 'ripemd160', 'sha512_256', 'whirlpool', 'md5-sha1', 'sha512_224'}
```

And example usage:
```pycon
>>> hasattr(hashlib, "ripemd160")
False
>>> h = hashlib.new("ripemd160")   # must go through new
>>> h.update(b"Nobody inspects the spammish repetition")
>>> h.hexdigest()
'cc4a5ce1b3df48aec5d22d1f16b894a0b894eccc'
```

I will probably use `algorithms_guaranteed` for now.

[...]

PR #357, opened 2023-05-15
Add EnsureHashAlgorithm constraint (draft)

mslw:
- [ ] handle list of names (currently expects string, gets list)
- [ ] handle letter casing
- [ ] deal with error messaging with `EnsureHashAlgorithm | EnsureListOf(EnsureHashAlgorithm)` https://github.com/datalad/datalad-next/issues/346#issuecomment-1549389089

Addresses #346

### DIFF
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -39,6 +39,7 @@ from datalad_next.constraints import (
     EnsureChoice,
     EnsurePath,
     EnsureURL,
+    EnsureHashAlgorithm,
 )
 from datalad_next.uis import (
     ansi_colors as ac,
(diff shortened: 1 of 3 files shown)


mslw:
I am using `ls-file-collection` as the model application of the validator. Hash parameter can be a `str | List(str)`; when coming from CLI it will always be the latter. I assumed `EnsureHashAlgorithm | EnsureListOf(EnsureHashAlgorithm)` would be the most intuitive constraint specification.

However, I got stupidly locked on how validation errors are reported with this constraint pattern: https://github.com/datalad/datalad-next/issues/346#issuecomment-1549389089

I would appreciate suggestions (@mih maybe you have a comment) for what would be the most logical solution:

- assume error messaging is good enough, keep the constraint and its usage as above
- assume error messaging needs to be improved separately, but the constraint and its usage is ok
- create an `EnsureMultiHash` constraint that would deal with strings & lists by itself, and create a `MultiHash` object already at the stage of parameter validation; adjust all code downstream to expect MultiHash as input

mih:
This PR has been in draft mode for a long time. I am closing it now as part of a cleanup run. Feel free to reopen once work on it restarts. Sorry.

PR #365, opened 2023-05-17
`HttpUrlOperations` decompresses download with appropriate encoding

### LINKED ISSUE #364: http-downloads via `download` (`HttpUrlOperations`) result in gzipped files on disk

mih:
Previously, a file downloaded from a server that supported `deflate`, or other types of compressions would end up in compressed form on disk. This is undesirable, as it changes the nature of files on download.

This change:

- adds a test that documents the faulty behavior, and the fix
- enables transparent decoding of compressed streams on download
- adjusts progress reporting to focus on bytes transferred, not bytes decompressed
- discontinued the use of `requests_toolbelt.downloadutils.tee`, arguably, using this (minimal) helper just makes things more opaque

Closes #364

### DIFF
--- a/datalad_next/url_operations/http.py
+++ b/datalad_next/url_operations/http.py
@@ -250,11 +250,16 @@ class HttpUrlOperations(UrlOperations):
             # compressed content
             expected_size = int(r.headers.get('content-length'))
         except (ValueError, TypeError):
+            # some responses do not have a `content-length` header,
+            # even though they HTTP200 and deliver the content.
+            # example:
+            # https://github.com/datalad/datalad-next/pull/365#issuecomment-1557114109
             expected_size = None
         self._progress_report_start(
             progress_id,
             ('Download %s to %s', from_url, to_path),
             'downloading',
+            # can be None, and that is OK
             expected_size,
         )
 
@@ -268,7 +273,6 @@ class HttpUrlOperations(UrlOperations):
             downloaded_bytes = 0
             # TODO make chunksize a config item, 65536 is the default in
             # requests_toolbelt
-            tell = 0
             for chunk in r.raw.stream(amt=65536, decode_content=True):
                 # update how much data was transferred from the remote server,
                 # but we cannot use the size of the chunk for that,
@@ -278,7 +282,7 @@ class HttpUrlOperations(UrlOperations):
                 if expected_size:
                     tell = r.raw.tell()
                 else:
-                    tell = tell + len(chunk)
+                    tell = downloaded_bytes + len(chunk)
                 self._progress_report_update(
                     progress_id,
                     ('Downloaded chunk',),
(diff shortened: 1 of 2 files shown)


mih:
I have exactly the same question in my head. thx!

The current test verifies that not decoding an uncompressed file obtained via a compressed download was broken, and is not fixed. The other way round (download something that should remain compressed) is not yet tested.

mih:
There is something weird going on with the test that ensures progress reporting. Locally it is susceptible to changes in download chunksize, but even with no change it works sometimes. Grrrr.

mih:
Now that the progress reporting reflects bytes-downloaded and not bytes decompressed, the test file size is too small to reliably take long enough for a progress update to appear

christian-monch:
First observation: in failing tests download logs progresses with `tell=0` and `download=0`.

christian-monch:
I got a 200 HTTP-response without `content-length` header. The reason that there is no `content-length` is, that the transfer-encoding is `chunked`.  I am not sure, when and why github returns the header, it might be related to locally cached data or CDN settings. Anyway, it is a valid response, and the client has no way of telling the overall size in advance. Commit 87858877ca9c32ed39f584591d530e8a5b59e072 should handle size reporting correctly in both cases, i.e. chunked and non-chunked

```python
{
    'Access-Control-Allow-Origin': 'https://render.githubusercontent.com',
    'content-disposition': 'attachment; filename=example-dicom-structural-f077bcc8d502ce8155507bd758cb3b7ccc887f40.zip',
    'Content-Security-Policy': "default-src 'none'; style-src 'unsafe-inline'; sandbox",
    'Content-Type': 'application/zip',
    'ETag': 'W/"55b2f976c68bdaeeec863a84b111afad6e788d3414cc34bf1ef4a925063bd72b"',
    'Strict-Transport-Security': 'max-age=31536000',
    'Vary': 'Authorization,Accept-Encoding,Origin',
    'X-Content-Type-Options': 'nosniff',
    'X-Frame-Options': 'deny',
    'X-XSS-Protection': '1; mode=block',
    'Date': 'Mon, 22 May 2023 11:53:24 GMT',
    'Transfer-Encoding': 'chunked',
    'X-GitHub-Request-Id': '0501:E603:C93E1E:F0E024:646B57B4'
}
```


For reference, a response header for non-chunked transfer from the same request:

```python
{
    'Access-Control-Allow-Origin': 'https://render.githubusercontent.com',
    'content-disposition': 'attachment; filename=example-dicom-structural-f077bcc8d502ce8155507bd758cb3b7ccc887f40.zip',
    'Content-Length': '16212183',
    'Content-Security-Policy': "default-src 'none'; style-src 'unsafe-inline'; sandbox",
    'Content-Type': 'application/zip',
    'ETag': 'W/"55b2f976c68bdaeeec863a84b111afad6e788d3414cc34bf1ef4a925063bd72b"',
    'Strict-Transport-Security': 'max-age=31536000',
    'Vary': 'Authorization,Accept-Encoding,Origin',
    'X-Content-Type-Options': 'nosniff',
    'X-Frame-Options': 'deny',
    'X-XSS-Protection': '1; mode=block',
    'Date': 'Mon, 22 May 2023 12:01:00 GMT',
    'X-GitHub-Request-Id': '055C:BBAA:9A54EC:BA29AC:646B597C'
}
```

PR #367, opened 2023-05-23
Enhance `FileSystemItem` and `iter_tar|dir()` to provide file IO

mih:
This enables something like this:

```
In [1]: from pydicom import dcmread
In [2]: from datalad_next.iter_collections.tarfile import iter_tar
In [3]: for i in iter_tar('/tmp/icfstore/dl-Z03/P000624_dicom.tar'):
   ...:     if i.fp:
   ...:         with dcmread(i.fp) as dcm:
   ...:             ...
```

Or in other words, arbitrary (post-)processing of files in a collection.

Here is a concrete use case that is enabled by this change: https://github.com/psychoinformatics-de/inm-icf-utilities/pull/6/commits/01349d87d34ec415dc55dd9b432fb8ffdebf9865

I had the concern that moving the hashing logic out of the iterators would remove the ability to not compute a particular hash, if the collection readily provides it (think dataverse offering MD5s). However, that is not true. Any such collection could provide this information unconditionally (it is free from a client POV), and the item2dict adapter for that collection in `ls_file_collections` can normalize such a report to use the standard `hash=<dict>` and skip computations as necessary and possible.

TODO

- [x] Add to documentation that the file pointers are only open until the iterator produces the next item.

### DIFF
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -48,7 +48,10 @@ from datalad_next.utils import ensure_list
 
 from datalad_next.iter_collections.directory import iter_dir
 from datalad_next.iter_collections.tarfile import iter_tar
-from datalad_next.iter_collections.utils import FileSystemItemType
+from datalad_next.iter_collections.utils import (
+    FileSystemItemType,
+    compute_multihash_from_fp,
+)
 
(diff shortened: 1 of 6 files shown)


christian-monch (approved):
LGTM

PR #223, opened 2023-01-26
Rough sketch of an archivist annex remote

mih:
This sits on top of #215 

It would be a replacement for the implementation of the
`datalad-archives` remote. In addition to its predecessor,
it would support sparse/partial access to remote archives,
avoiding the 200% storage overhead, and the requirement to
unconditionally download full archives.

Ultimately this will address #185

Related issues:

- datalad/datalad#7121 (archives would no
  longer be fully extracted, and need no removal, unless configured)
- datalad/datalad#3270 (uses UrlOperations
  framework with standard progress reporting)
- datalad/datalad#373 (partial access to
  remote archives would be supported for all FSSPEC-supported
  archive types and access protocols)

PR #368, opened 2023-05-23
`archivist` remote reborn

mih:
This is #223 again, but without the `fsspec` foundation. The special remote itself still has provisions for it, but the idea here is to be able to get it to a releasable state without having to wait for the FSSPEC functionality to mature.


mih:
This was already dead on arrival, another attempt is coming.

PR #371, opened 2023-05-24
Establish `datalad_next.config` for `ConfigManager`

### LINKED ISSUE #337: Add cannonical import for `ConfigManager` class

mih:
This will be the cannonical import location. This changeset starts using that for imports that are presently focused on type-annotation.

Closes #337 

This change was pulled from https://github.com/datalad/datalad-next/pull/339 -- which stalled and is now blocking this fix.

### DIFF
--- a/datalad_next/config/__init__.py
+++ b/datalad_next/config/__init__.py
@@ -1,0 +1,1 @@
+from datalad.config import ConfigManager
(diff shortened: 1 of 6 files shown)


christian-monch (approved):
LGTM

mih:
Thanks @christian-monch 

Error is httbin fluke.

Issue #184, opened 2022-12-12
Implement `ArchiveOperations`


mih:
Analog to `UrlOperations` but for creating and extracting from archives of any kind.

There a various implementations in datalad-core for this. Direct use of `tarfile`, `zipfile`, a `7z` wrapper, a `patool` wrapper. The main operations supported are `compress_file` and `decompress_file` (note the single-file granularity). For performance reasons, there is also a caching layer built around this to make extracting more than one file faster, at the expense of diskspace consumption.

I think we can apply the same concept as for `UrlOperations` -- a handler per archive type(combination), plus a set of operations.

- `extract`
- `extract_all` (to be efficient, when it is known that we need all)
- `add`
- `create_from` (to be efficient, when it is known that we just need to ingest some blobs into a new archive)

Whether or not to keep extracted copies around for caching, or do other maintenance or cleanup operations would be a configuration of the handler (or the task of alternative handlers).

In order to optimally support something like #183 , it would be good to provide an archive content iterator too.

Survey of archive operations in the datalad codebase:

- `datalad.local.export_archive`: Wrap `tarfile/zipfile` to create an archive from scratch, with timestamp customization.
- `datalad.distributed/export_archive_ora`: Create a `7z` archive from scratch (with annex keys), only `7z`-native customization options supported.
- `datalad.support.archive_utils_7z`: extract and add single files to an archive (within ArchiveCache framework)
- - `datalad.support.archive_utils_patool`: extract and add single files to an archive (within ArchiveCache framework)
- `datalad.distributed.ora`: extract a single file from a `7z` archive
- `datalad.customremotes.archives`: Extract a single file from an archive (within ArchiveCache framework)
- `datalad.distributed.export_to_figure`: Create archive from scratch using `export_archive`
- `datalad_next.gitremotes.datalad_annex`: create an zipfile from scratch, and fully extract all content from a zipfile
- `datalad_ukbiobank.update`: Index all zipfile content via `add-archive-content`
- `datalad_osf.git_remote`: create an zipfile from scratch, and fully extract all content from a zipfile

bpoldrack:
> Whether or not to keep extracted copies around for caching

Just leaving a reminder, that potential extraction cache location(s) should be made known to `datalad clean`.

mih:
[...]

PR #361, opened 2023-05-17
`iter_gitworktree()` for processing work tree content

mih:
The iterator is also integrated with `ls-file-collection` as collection type `gitworktree`.

Closes #350
Ping #323

Issue #350, opened 2023-05-10
Implement `iterworktree()` for `ls-file-collection`


mih:
Like #349 but centered around `lsfiles` instead of `lstree`. Together they can replace the custom implementation in `AnnexRepo.get_content_info()`

Ping #343

PR #372, opened 2023-05-24
Sketch of a base class for `ArchiveOperations`

mih:
Related to https://github.com/datalad/datalad-next/issues/184 but different in scope and aims.

### DIFF
--- a/datalad_next/archive_operations/__init__.py
+++ b/datalad_next/archive_operations/__init__.py
@@ -55,7 +55,7 @@ class ArchiveOperations(ABC):
 
     @abstractmethod
     def close(self) -> None:
-        raise NotImplementedError
+        pass
 
     @abstractmethod
     def __contains__(self, item: Any) -> bool:
--- a/datalad_next/archive_operations/__init__.py
+++ b/datalad_next/archive_operations/__init__.py
@@ -17,8 +17,8 @@ from typing import (
 
 import datalad
 
-from ..config import ConfigManager
-from ..iter_collections.utils import FileSystemItem
+from datalad_next.config import ConfigManager
+from datalad_next.iter_collections.utils import FileSystemItem
 
 
 lgr = logging.getLogger('datalad.ext.next.archive_operations')
@@ -53,7 +53,6 @@ class ArchiveOperations(ABC):
         """
         raise NotImplementedError
 
-    @abstractmethod
     def close(self) -> None:
         pass
 
--- a/datalad_next/archive_operations/__init__.py
+++ b/datalad_next/archive_operations/__init__.py
@@ -3,28 +3,31 @@
 # allow for |-type UnionType declarations
 from __future__ import annotations
 
-from contextlib import contextmanager
 import logging
-from pathlib import Path
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from contextlib import contextmanager
 from typing import (
     Any,
-    Dict,
+    Generator,
     IO,
 )
 
+import datalad
+
+from ..config import ConfigManager
+from ..iter_collections.utils import FileSystemItem
+
+
 lgr = logging.getLogger('datalad.ext.next.archive_operations')
 
-#
-# TODO
-# - add ConfigManager type annotation after
-#   https://github.com/datalad/datalad-next/pull/371 is available
-#
 
-
-class ArchiveOperations:
+class ArchiveOperations(ABC):
     """
     """
-    def __init__(self, location: Any, *, cfg=None):
+    def __init__(self, location: Any, *, cfg: ConfigManager | None = None):
         """
         Parameters
         ----------
@@ -38,19 +41,26 @@ class ArchiveOperations:
         self._cfg = cfg
 
     @property
-    def cfg(self):  # -> ConfigManager
+    def cfg(self) -> ConfigManager:
         if self._cfg is None:
             self._cfg = datalad.cfg
         return self._cfg
 
     @contextmanager
+    @abstractmethod
     def open(self, item: Any) -> IO:
         """
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def close(self) -> None:
+        raise NotImplementedError
+
+    @abstractmethod
     def __contains__(self, item: Any) -> bool:
         raise NotImplementedError
 
-    def __iter__(self) -> Generator[TarfileItem, None, None]:
+    @abstractmethod
+    def __iter__(self) -> Generator[FileSystemItem, None, None]:
         raise NotImplementedError
--- a/datalad_next/archive_operations/__init__.py
+++ b/datalad_next/archive_operations/__init__.py
@@ -1,0 +1,56 @@
+"""Archive operation handlers"""
+
+# allow for |-type UnionType declarations
+from __future__ import annotations
+
+from contextlib import contextmanager
+import logging
+from pathlib import Path
+from typing import (
+    Any,
+    Dict,
+    IO,
+)
+
+lgr = logging.getLogger('datalad.ext.next.archive_operations')
+
+#
+# TODO
+# - add ConfigManager type annotation after
+#   https://github.com/datalad/datalad-next/pull/371 is available
+#
+
+
+class ArchiveOperations:
+    """
+    """
+    def __init__(self, location: Any, *, cfg=None):
+        """
+        Parameters
+        ----------
+        location:
+          Archive location identifier (path, URL, etc.) understood by a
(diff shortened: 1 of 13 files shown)


datalad_next/archive_operations/__init__.py
```diff
@@ -0,0 +1,56 @@
+"""Archive operation handlers"""
+
+# allow for |-type UnionType declarations
+from __future__ import annotations
+
+from contextlib import contextmanager
+import logging
+from pathlib import Path
+from typing import
```
christian-monch, line 9:
```suggestion
from typing import (
```

datalad_next/archive_operations/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+from contextlib import contextmanager
+import logging
+from pathlib import Path
+from typing import 
+    Any,
+    Dict,
+    IO,
+)
+
```
christian-monch, line 14:
```suggestion

import datalad

from ..iter_collections.tarfile import TarfileItem

```

datalad_next/archive_operations/__init__.py
mih: Fixed

datalad_next/archive_operations/__init__.py
mih: It should not be TarfileItem below, but something more generic. Not yet sure in which way.

christian-monch: > It should not be TarfileItem below, but something more generic. Not yet sure in which way.

I thought so too. I guess `FilesystemItem` would be more appropriate. First, for the name and second because `TarfileItem` does not actually extend `FilesystemItem`. We could rename it to something like `ArchiveItem`

christian-monch: Changed it to FilesystemItem

datalad_next/archive_operations/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    @contextmanager
+    @abstractmethod
+    def open(self, item: Any) -> IO:
+        """
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def close(self) -> None:
+        raise NotImplementedError
```
mih, line 58: Likely a better default

```suggestion
        pass
```

datalad_next/archive_operations/tarfile.py
```diff
@@ -0,0 +1,93 @@
...
+# is nice, and `__iter__()` only has `self`, such that
+# any customization would need to be infused in the whole
+# class. Potentially cumbersome.
+from datalad_next.iter_collections.tarfile import (
+    TarfileItem,
+    iter_tar,
+)
+
+from . import ArchiveOperations
+from ..config import ConfigManager
```
mih, line 30: The must not be relative importing across `datalad_next` subpackages. Only within subpackage.

datalad_next/archive_operations/tests/common.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import annotations
+
+from .. import ArchiveOperations
+from ...iter_collections.tests.utils import TestCollection
```
mih, line 4: Must be absolute

datalad_next/archive_operations/tests/test_tarfile_archive_operations.py
```diff
@@ -0,0 +1,15 @@
+
+from .common import check_basic_functionality
+from ..tarfile import TarArchiveOperations
+from ...iter_collections.tests.utils import (
```
mih, line 4: Must be absolute

datalad_next/archive_operations/tests/test_zipfile_archive_operations.py
```diff
@@ -0,0 +1,15 @@
+
+from .common import check_basic_functionality
+from ..zipfile import ZipArchiveOperations
+from ...iter_collections.tests.utils import (
```
mih, line 4: Must be absolute

datalad_next/archive_operations/zipfile.py
```diff
@@ -0,0 +1,102 @@
...
+# is nice, and `__iter__()` only has `self`, such that
+# any customization would need to be infused in the whole
+# class. Potentially cumbersome.
+from datalad_next.iter_collections.zipfile import (
+    ZipfileItem,
+    iter_zip,
+)
+
+from . import ArchiveOperations
+from ..config import ConfigManager
```
mih, line 29: Must be absolute

datalad_next/archive_operations/tests/test_tarfile_archive_operations.py
christian-monch: Done

datalad_next/archive_operations/tarfile.py
christian-monch: Done

datalad_next/archive_operations/tests/common.py
christian-monch: Done

datalad_next/archive_operations/tests/test_zipfile_archive_operations.py
christian-monch: Done

datalad_next/archive_operations/zipfile.py
christian-monch: Done

datalad_next/archive_operations/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @property
+    def cfg(self) -> ConfigManager:
+        if self._cfg is None:
+            self._cfg = datalad.cfg
+        return self._cfg
+
+    @contextmanager
+    @abstractmethod
+    def open(self, item: Any) -> IO:
```
mih, line 51: Likely needs to become

```suggestion
    def open(self, item: PurePosixPath) -> IO:
```

to set concrete expectations. The `archivist`/`datalad-archives` use case will always come with such paths. I am not sure of particular archive formats use non-POSIX paths internally, but knowing the type of input path should help make any transformations that are necessary.

Are there any archives that use non-path item/member identification schemes?

datalad_next/archive_operations/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def close(self) -> None:
+        pass
+
+    @abstractmethod
+    def __contains__(self, item: Any) -> bool:
+        raise NotImplementedError
+
+    @abstractmethod
+    def __iter__(self) -> Generator[FileSystemItem, None, None]:
```
mih, line 64: I am not convinced that this is a good choice, but maybe it is. We are operating under the assumption that all iterator will produce that. I have no good feeling whether that is valid.

christian-monch: I share a "slightly weird" feeling. It is better than `TarfileItem`. I think we have to answer a general question:

- Should the yield-type of `ArchiveOperations.__iter__` and of `iter_<x>` be in the same class hierarchy?

For example:

```python
# The name below is stupid and just used to illustrate the point 
@dataclass
class CollectionOrArchiveItem:
    name: PurePosixPath




@dataclass
class FilesystemItem(CollectionOrArchiveItem):
    type: FileSystemItemType
    size: int
    mtime: float | None = None
    mode: int | None = None
    uid: int | None = None
    gid: int | None = None
    link_target: PurePath | None = None
```


Pro: it would make sense because both are members of some of set that distinguishes its elements by their name (although the elements usually have additional properties, the names are the unique distinction criteria). So there could be a common base-class for those types that basically contains a name.

Con: there are distinctions between the names in archives and names in collections. Names in archives are specified by the archive specs and should be identical across platforms. Names in collections might not be identical across platforms, i.e. Windows-dirs vs. Unix-dirs, unless they are represented as strings or a sequence of bytes (which we are currently doing)

mih: I arrived at similar conclusions. Please check https://github.com/datalad/datalad-next/issues/350 and specifically https://github.com/datalad/datalad-next/pull/361/commits/63d418e69ee1883c23185ca012e36cd8bc071256 for my approach from 1-2 weeks ago.

mih:
I am closing this PR now. The actual "Sketch of a base class for ArchiveOperations" has already been merged into `main` with https://github.com/datalad/datalad-next/pull/384. Other aspects of the diff here have been proposed at separate PR(s):

- https://github.com/datalad/datalad-next/pull/383

The remaining pieces also need to go down this path.

PR #374, opened 2023-05-24
Small remote access demo and performance baseline an experimental http_tarfile iterator

christian-monch:
This PR adds a simple version of a tar-file iterator over HTTP. The main idea is to replace a local file-like object with a file-like object that transparently reads from an HTTP-location. This leaves most of the iteration code untouched.

- [ ] TODO: add tests


christian-monch:
This is not going into main, but it might be useful to have as a performance baseline for tar-archive access over HTTP

PR #378, opened 2023-05-24
Explicit type for archive locators

mih:
These are the `dl+archive:...` style URLs used by the `datalad-archives` special remote and by the upcoming `archivist`.

This also brings an ENUM for archive types, and a place to put other enums too.


mih:
All errors are services outages (gitea in particular)

Issue #185, opened 2022-12-12
Re-Implement `datalad-archives` special remote 


mih:
When #184 is done, the `datalad-archives` special remote can be replaced with an implementation that sits on top of that code.

I see no reason to make any changes to the syntax and composition of the URLs themselves, so it would be a drop-in replacement.

mih:
There might be another way to do this and also make it possible perform partial downloads from remote archives. Here is the sketch:

If ....

- the archive key has one or more URLs assigned
- we know the archive type

we can autogenerate an FSSPEC-style URL chain, and hand it over to the FSSPEC-downloader (see https://github.com/datalad/datalad-next/pull/215). IF such an attempt fails, we can perform a regular `get` on the entire archive and extract a file as done in the original code.

FSSPEC also allows for several types of access caches to be embedded into such a URL chain.

So a complete plan would tie into #184 too:

- we would have an abstract representation of an `Archive`
- an `Archive` would have a location that is a URL
- an `Archive` would also have a type (possibly inferred from the URL). Something like `tar`, `zip`, `7z`, etc.
- we would have a `RemoteArchive` that supports some read-only operations (via FSSPEC). This may be merely an API wrapper around `FsspecUrlOperations` from #215)
- we would have a `LocalArchive` that is a complete interface to all operations we require to be able to do on archives (see #184) for a survey
- we would have `AnyArchiveOperations` (analog to `AnyUrlOperations`) that is a handler class for switching between concrete `Archive` implementations. This would be the main workhorse of a `datalad-archives` special remote.

Such an approach would complement (or make obsolete) #207, because there would be no need for supporting `dl+archive` URLs in `uncurl`, but at the same time, we would still use the code path via the `UrlOperations` downloaders to perform content access.

mih:
so the full dev goal would be:

- #183 to be able to annotate archive keys or urls pointing to archives with archive type information. This functionality should be accessible as a command (maybe replacing the implementation in add-archive-content) for existing annexed archives, but also for exporting new archives from existing dataset content. The latter would be inefficient when relying on the tooling of the former.
- #184 (or an equivalent) with a handler to return an appropriate representation of an archive (local or remote, with extracted cache or without) to be used in the special remote reimplementation.
- A collection of archive interface classes (FsspecArchiveOperations, 7zArchiveOperations, PatoolArchiveOperations) interfacing archive access tools not particular archive types. They may or may not support particular types of caching approaches.

PR #380, opened 2023-05-25
`archivist` git-annex special remote

mih:
This is a replacement for the implementation of the `datalad-archives` remote. In addition to its predecessor, it reduces the storage overhead from 200% to 100% by doing partial extraction from fully downloaded archives.

Ultimately, it will support sparse/partial access to remote archives, avoiding any storage overhead, and the requirement to unconditionally download full archives (see https://github.com/datalad/datalad-next/pull/215).

This implementation is trying to be efficient by:

- trying to fulfill requests using locally present archives
- trying to download smaller archives before larger ones

This implementation is aiming to be extensible by:

- using `ArchiveOperations` as a framework to uniformly implement (partial) access to local (or remote) archives.

This addresses #185

TODO:

- [x] Bring up test coverage
- [x] Announce new configuration via `register_config()`
- [ ] Add progress reporting for the extraction from an archive (look for `copyfileobj`). The archive download progress is already taken care of via `uncurl`
- [x] Proofread and update docs
- [x] Establish `SpecialRemote.repo` property (`LegacyAnnexRepo(self.annex.getgitdir())` on demand) -- done in #387
- [x] Establish `SpecialRemote.getcfg()` -- done in #387
- [ ]  Various cornercases and component are still in need of tests. Hints:

```
Name                                     Stmts   Miss  Cover   Missing
----------------------------------------------------------------------
datalad_next/annexremotes/archivist.py     159     16    90%   333-337, 341, 431, 449, 462-464, 479, 522, 532, 545-550
----------------------------------------------------------------------
TOTAL                                      159     16    90%
```

Related issues:

- datalad/datalad#7121 (archives would no
  longer be fully extracted, and need no removal, unless configured)
- datalad/datalad#3270 (uses UrlOperations
  framework with standard progress reporting)
- datalad/datalad#373 (partial access to
  remote archives would be supported for all FSSPEC-supported
  archive types and access protocols)

### DIFF
--- a/datalad_next/__init__.py
+++ b/datalad_next/__init__.py
@@ -88,6 +88,21 @@ register_config(
     default='raise-early',
     dialog='question',
 )
+register_config(
+    'datalad.archivist.legacy-mode',
+    'Fall back on legacy ``datalad-archives`` special remote implementation?',
+    description='If enabled, all `archivist` special remote operations '
+    'fall back onto the legacy ``datalad-archives`` special remote '
+    'implementation. This mode is only provided for backward-compatibility. '
+    'This legacy implementation unconditionally downloads archive files '
+    'completely, and keeps an internal cache of the full extracted archive '
+    'around. The implied 200% storage cost overhead for obtaining a complete '
+    'dataset can be prohibitive for datasets tracking large amount of data '
+    '(in archive files).',
+    type=EnsureBool(),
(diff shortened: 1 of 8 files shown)


mih:
I will merge this now. It has been tried out in the first target use case and worked according to soec

PR #381, opened 2023-05-26
Make `AnnexKey` hashable

### LINKED ISSUE #379: `AnnexKey` should be hashable
I would think that using this type as a key in a mapping would be a standard use case. It would require the following change:

```diff
diff --git a/datalad_next/types/annexkey.py b/datalad_next/types/annexkey.py
index 74bbcb6..5a85342 100644
--- a/datalad_next/types/annexkey.py
+++ b/datalad_next/types/annexkey.py
@@ -14,7 +14,7 @@ _annexkey_regex = re.compile(
 )
 
 
-@dataclass
+@dataclass(frozen=True)
 class AnnexKey:
     """Representation of a git-annex key
 
```

This would mean that post-init assignment of properties is not possible. I tend to think that this is OK.

mih:
I would think that using this type as a key in a mapping would be a standard use case. This would mean that post-init assignment of properties is not possible. I tend to think that this is OK.

Closes #379

### DIFF
--- a/datalad_next/types/annexkey.py
+++ b/datalad_next/types/annexkey.py
@@ -1,3 +1,5 @@
+"""git-annex key representation"""
+
 from __future__ import annotations
 
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


mih:
All test failures are service outages (gitea)

PR #383, opened 2023-05-26
`iter_zip()`

mih:
Extracted bolt-on from https://github.com/datalad/datalad-next/pull/372


christian-monch (approved):
I think that is the complete zip_iter implementation. Thanks for creating the PR. (This is not a comment on the code, since I wrote it)

mih:
OK, let's merge. Thx! Linking into `ls-file-collection` is still TODO.

mih:
Damn, also docs are missing... Will do

PR #385, opened 2023-05-26
Fix type annotation for `cfg`-parameter in `DataladAuth.__init__()`

christian-monch:
This PR replaces the faulty CredentialManager annotation for the cfg-parameter of DataladAuth.__init__() with the correct type, i.e. ConfigManager.

### DIFF
--- a/datalad_next/utils/requests_auth.py
+++ b/datalad_next/utils/requests_auth.py
@@ -9,6 +9,7 @@ from urllib.parse import urlparse
 import requests
 import www_authenticate
 
+from datalad.config import ConfigManager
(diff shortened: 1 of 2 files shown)


datalad_next/utils/requests_auth.py
```diff
@@ -9,6 +9,7 @@
 import requests
 import www_authenticate
 
+from datalad.config import ConfigManager
```
mih, line 12:
```suggestion
from datalad_next.config import ConfigManager
```

mih (commented):
Thanks. I suggested a smal change that I will commit and merge.

mih (approved):

PR #386, opened 2023-05-30
`repr(CommandError(...))` now informative

mih:
`CommandError`, the standard exception raised for any non-zero exit command execution now reports which command failed with `repr()` too. Previously, only `str()` would produce an informative message about a failure, while `repr()` would report `CommandError('')`, unless a dedicated message was provided.

Changes:

```
Could not obtain 'MD5E-s20--7fecdac24beaba574b1e9d7ff9da67f8.csv' -caused by- Exhausted all candidate archive handlers (previous failures [CommandError('')])
```

to

```
Could not obtain 'MD5E-s20--7fecdac24beaba574b1e9d7ff9da67f8.csv' -caused by- CommandError: 'git -c diff.ignoreSubmodules=none annex contentlocation MD5E-s391--15a48a2cb9efbdedece04ff16be413e9.tar.gz -c annex.dotfiles=true' failed with exitcode 1
```

Ping @christian-monch -- should likely also be done in datalad-core.

TODO:

- [ ] Transform into a patch of datalad-core. With the subclass approach any `except CommandError` using this subclass would not be catch a `CommandError` raised by datalad-core.

### DIFF
--- a/changelog.d/20230530_084701_michael.hanke.md
+++ b/changelog.d/20230530_084701_michael.hanke.md
@@ -1,0 +1,7 @@
+### 🐛 Bug Fixes
+
+- Patch `CommandError`, the standard exception raised for any non-zero exit
+  command execution to now reports which command failed with `repr()` too.
+  Previously, only `str()` would produce an informative message about a failure,
+  while `repr()` would report `CommandError('')`, unless a dedicated message was
+  provided. (by @mih)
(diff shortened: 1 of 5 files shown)


yarikoptic:
I am really not sure why such changes aren't just suggested directly in the core to start with.

mih:
Speed is the only reason.

PR #392, opened 2023-06-01
Diversify `Item` type classes

mih:
Originally introduced in
https://github.com/datalad/datalad-next/pull/361, but a need for this has also been expressed elsewhere, hence now a standalone changeset.

These types could also move into `datalad_next.types...`

### DIFF
--- a/datalad_next/iter_collections/utils.py
+++ b/datalad_next/iter_collections/utils.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 from enum import Enum
 from pathlib import PurePath
 from typing import (
+    Any,
     IO,
     List,
 )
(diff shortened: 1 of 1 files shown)


datalad_next/iter_collections/utils.py
```diff
@@ -28,9 +29,26 @@ class FileSystemItemType(Enum):
...
+@dataclass
+class TypedItem:
+    type: Any
+
+
+@dataclass
+class PathBasedItem(NamedItem):
+    # a path-identifier in an appropriate context.
+    # could be a filename, a relpath, or an absolute path.
+    # should match platform conventions
```
christian-monch, line 46: Would an item in a tar-file or zip-file be a "PathBasedItem" or only a "NamedItem"? In the former case, the "platform" from the comment "should match platform conventions" would refer to tar-file or zip-file spec. Which is OK, I am just mentioning it.

christian-monch (approved):
I like it, it makes sense to me (we already discussed the general idea here: https://github.com/datalad/datalad-next/pull/372#discussion_r1208964770).

As with most categorization efforts, things might change slightly in the future, but I think this is a good trajectory.

datalad_next/iter_collections/utils.py
mih: I think it would be `PathBasedItem`.

I have a comment somewhere that is wondering whether there are archive formats that use non-POSIX member names. Do you know any?

mih: Ah, now I see what you mean. No, the "platform" is not tar-file or zip-file. The question is posix path of windows path. However, given that this is nothing specific to this PR. I have filed https://github.com/datalad/datalad-next/issues/402 so we can move on here.

PR #396, opened 2023-06-01
Show realm in credential prompt

christian-monch:
This PR inserts information about the protection realm into the credential prompt for 401-error handling. This allows a user to enter a credential matching the protection realm.

RFC 2617 defines the protection realm as follows:

> The realm directive (case-insensitive) is required for all
> authentication schemes that issue a challenge. The realm value
> (case-sensitive), in combination with the canonical root URL (the
> absoluteURI for the server whose abs_path is empty; see section 5.1.[2](https://datatracker.ietf.org/doc/html/rfc2617#ref-2)
> of [2]) of the server being accessed, defines the protection space.
> These realms allow the protected resources on a server to be
> partitioned into a set of protection spaces, each with its own
> authentication scheme and/or authorization database. The realm value
> is a string, generally assigned by the origin server, which may have
> additional semantics specific to the authentication scheme. Note that
> there may be multiple challenges with the same auth-scheme but
> different realms.

That means, the realm, i.e. realm value, is a *name* that identifies a set of resources on the server. It cannot generally be determined from the URL, but only from the *WWW-Authenticate*-header. The realm might be identical to a prefix of the resource's local path, but that is not required.

For example, the fixture `http_server_with_basicauth` will always report the realm `Protected`, while the local path will typically be something like `/testfile.txt`.  Another example is the URL `https://docs.inm7.de/`. All paths in this host, e.g. `https://docs.inm7.de/learning`, are protected by a realm with the name `INM-7 Documentation`. Yet, another example would be the realm `Online Service Portal (Prodsystem)` for a local path `/Prod/InfoSystem/eTravel/Neu/Uebersicht.pl`, which is used in an employee-management portal that the authors use.

If we want to enable users to enter credentials for a realm via `datalad credentials set`, it would be helpful for them to report the realm in the protection prompt.

### DIFF
--- a/datalad_next/url_operations/http.py
+++ b/datalad_next/url_operations/http.py
@@ -117,7 +117,7 @@ class HttpUrlOperations(UrlOperations):
             try:
                 props['content-length'] = int(props['content-length'])
             except (TypeError, ValueError):
-                # but be resonably robust against unexpected responses
+                # but be reasonably robust against unexpected responses
                 pass
         return props
 
--- a/datalad_next/url_operations/tests/test_http.py
+++ b/datalad_next/url_operations/tests/test_http.py
@@ -96,3 +96,19 @@ def test_compressed_file_stay_compressed(tmp_path):
     # make sure it ends up on disk compressed!
     with gzip.open(dpath, 'rb') as f:
         f.read(1000)
+
+
+def test_header_adding():
+    default_headers = dict(key_1='value_1')
+    added_headers = dict(key_2='value_2')
+    url_ops = HttpUrlOperations(headers=default_headers)
+    assert 'key_1' in url_ops.get_headers()
+
+    # ensure that header entries from `headers` show up in result
+    combined_keys = {'key_1', 'key_2'}
+    result_1 = url_ops.get_headers(headers=dict(added_headers))
+    assert combined_keys.issubset(set(result_1))
+
+    # ensure that `headers` did not change the stored headers
+    result_2 = url_ops.get_headers()
+    assert 'key_2' not in set(result_2)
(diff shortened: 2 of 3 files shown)


mih (changes requested):
There is no demo of the new vs old prompt in the PR or a commit message (would insist on the latter and appreciate the former). Without that we would not know how it is supposed to look beyond the str-match test included.

My concern is a lot of duplication between `url` and `realm` -- but I might just not remember what they actually are.

christian-monch:
> Thanks for the PR.
> 
> There is no demo of the new vs old prompt in the PR or a commit message (would insist on the latter and appreciate the former). Without that we would not know how it is supposed to look beyond the str-match test included.
> 
> My concern is a lot of duplication between `url` and `realm` -- but I might just not remember what they actually are.

I amended the PR description and the commit. I added a number of examples that show that `url` and `realm` are often different and point to the definition (RFC 2617) to highlight that there is no defined relation between a `URL-path` and a realm, other than the `WWW-Authenticate`-header.

mih:
Thanks. I am aware how this works in general. What I was talking about is how a prompt actually looks (i.e. what a user sees). I see the template change in the diff, but it is unclear what the content of the variable `realm` actually is at that point. Putting the real output comparison in the description helps with that. Otherwise, I have to try to run it, as soon as I get to it.

christian-monch:
> Thanks. I am aware how this works in general.

Sorry, didn't mean to sound like a condescending prick. That wasn't my intention.

> What I was talking about is how a prompt actually looks...

I see now, what you mean. Yes, necessarily the scheme- and location-part would be identical. I still think it would be beneficial for the user to output the realm. I could change it to something like

```
Credential needed for realm ``https://docs.inm7.de/INM-7 Documentation´´ (which controls access to resource ``/learning´´)
```
WDYT?

christian-monch:
> What about [...]

I like it. Changed to your proposal in 39305eec52f20fa833de830b6a703e7b5d4fd9ab. Credential prompt is now:

```
Credential needed for accessing https://docs.inm7.de/learning (authentication realm 'INM-7 Documentation')
``` 

in contrast to the following from `main`-branch:

```
Credential needed for access to https://docs.inm7.de/learning
```

PR #403, opened 2023-06-04
Establish `FileSystemItem.from_path()`

mih:
Can be reused by other derived classes.


mih:
Ready to go, but a merge is waiting for #361 to catch up to this and confirm that it is useful for that too.

PR #408, opened 2023-06-07
Use correct return type of `ArchiveOperations.open()` to `Generator`

### LINKED ISSUE #406: Type annotations of `TarArchiveOperations.open` are not correct
This is an issue that is "caused" by the use of the `contextmanager`-decorator in the definition of the method `TarArchiveOperations.open`. 

The body of the implementation of `TarArchiveOperations.open()` returns a generator, since it uses a `yield`-statement. In combination with the `@contextmanager`-decorator the `open()`-method on an instance becomes a method that returns an `IO`-object. So, any caller of `open()` on a `TarArchiveOperations`-instance will see a return value that is an instance of `IO`.

The doc-strings should document the behavior that is observed on an instance of `TarArchiveOperations`.

But if we introduce type-checking, the annotations for the return-value, i.e. `IO`, will trigger an error, because the real return-type of the function specified by the code in `TarArchiveOperations.open()` is `Generator[IO, None, None]`.

There are a few options to mitigate the problem:

1. Do not introduce type-checks and ignore the discrepancy
2. Introduce type-checks, but ignore this particular definition (not sure if that is possible for all type checkers)
3. Correct the type annotation in the code, document the behavior of instances in the doc-string.

I would prefer option 3. The code annotations would be correct, and the documentation of the behavior of `open()` on instances of `TarArchiveOperations` would also be correct.

christian-monch:
Fixes #406 

This PR implements solution 3. described in item #406.

It changes the return type of the method `open()` of `ArchiveOperations` and its subclasses to reflect the return type that is defined in the method code. That means the return type is changed to `Generator`.

The PR also introduces a complete doc-string for `TarArchiveOperations.open()` to describe the return types that are observed y the caller of `open()` on an instance of `TarArchiveOperations`. Due to the use of the `@contextmanager`-decorator, the return type observed by a caller is `IO | None` instead of `Generator`.

### DIFF
--- a/datalad_next/archive_operations/__init__.py
+++ b/datalad_next/archive_operations/__init__.py
@@ -95,7 +95,7 @@ class ArchiveOperations(ABC):
 
     @contextmanager
     @abstractmethod
-    def open(self, item: Any) -> IO:
+    def open(self, item: Any) -> Generator[IO | None]:
         """Get a file-like for an archive item
 
         Parameters
--- a/datalad_next/archive_operations/tarfile.py
+++ b/datalad_next/archive_operations/tarfile.py
@@ -77,13 +77,28 @@ class TarArchiveOperations(ArchiveOperations):
             self._tarfile = None
 
     @contextmanager
-    def open(self, item: str | PurePath) -> IO:
+    def open(self, item: str | PurePath) -> Generator[IO | None]:
         """Get a file-like for a TAR archive item
+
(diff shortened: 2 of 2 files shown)


mih:
I tried rebasing this to resolve the conflicts, but I got confused. I will change this PR to draft mode. Please switch back once the conflicts are addressed.

mih (approved):
LGTM, thanks!

Issue #402, opened 2023-06-02
Decided on `PathBasedItem.name` being `PurePath` or `PurePosixPath`


mih:
Question has come up repeatedly, last https://github.com/datalad/datalad-next/pull/392/files#r1212725308

Which one should it be? Right now it is `PurePath`, hence platform (OS) specific.

PR #409, opened 2023-06-08
Use PurePosixPath for names of zipfile-entries (or even better: `str`)?

christian-monch:
This PR modifies zip-iterators to use `PurePosixPath` for names of archive members instead of `PurePath`. The rationale for this change is that according to [zip-APPNOTE.TXT version 6.3.3](https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.3.TXT) from September 1st, 2012, all slashes in file names (contained in zip) _must_ be forward slashes, i.e. '/'. From September 26th, 2006 onwards [zip-APPNOTE.TXT version 6.3.0](https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.0.TXT) states that all slashes _should_ be forward slashes.

It seems to be safe to assume that almost all zip-files we encounter will be using forward slashes. Because `PurePath` instances are OS-specific, a `PurePath` on Windows will contain `\`-slashes and not properly represent the content of the archive. Consequently, those paths cannot be used to retrieve items from a `zipfile.ZipFile` without conversion to a POSIX-style path.

We should therefore use `PurePosixPath`-instances to represent the name of elements in a zip-archive.

This touches again on issue #402. Maybe we should introduce different base classes, one for objects with platform-dependent paths, which would use `PurePath`, and one for objects with platform-independent, fixed, path-formats, e.g.  zip-file items.

### DIFF
--- a/datalad_next/iter_collections/tests/test_iterzip.py
+++ b/datalad_next/iter_collections/tests/test_iterzip.py
@@ -1,6 +1,6 @@
 import pytest
 import zipfile
-from pathlib import PurePath
+from pathlib import PurePosixPath
 
 from ..zipfile import (
     ZipfileItem,
@@ -47,7 +47,7 @@ def test_iter_zip(sample_zip):
         'SHA1': 'b5dfcec4d1b6166067226fae102f7fbcf6bd1bd4',
         'md5': 'd700214df5487801e8ee23d31e60382a',
     }
-    root = PurePath('test-archive')
+    root = PurePosixPath('test-archive')
     targets = [
         ZipfileItem(
             name=root,
--- a/datalad_next/iter_collections/tests/test_iterzip.py
+++ b/datalad_next/iter_collections/tests/test_iterzip.py
@@ -26,7 +26,7 @@ def sample_zip(tmp_path_factory):
         test-archive
         ├── onetwothree.txt
         └── subdir/
-            └── onetwothree_again.txt
+            └── onetwothree<>again.txt
     """
     path = tmp_path_factory.mktemp('zipfile') / 'sample.zip'
     file_content = b'zip-123\n'
@@ -35,7 +35,7 @@ def sample_zip(tmp_path_factory):
         zip_file.writestr('test-archive/subdir/', '')
         with zip_file.open('test-archive/onetwothree.txt', mode='w') as fp:
             fp.write(file_content)
-        with zip_file.open('test-archive/subdir/onetwothree_again.txt', mode='w') as fp:
+        with zip_file.open('test-archive/subdir/onetwothree<>again.txt', mode='w') as fp:
             fp.write(file_content)
 
     yield path
@@ -65,7 +65,7 @@ def test_iter_zip(sample_zip):
             size=0,
         ),
         ZipfileItem(
-            name=root / 'subdir' / 'onetwothree_again.txt',
+            name=root / 'subdir' / 'onetwothree<>again.txt',
             type=FileSystemItemType.file,
             size=8,
         ),
--- a/datalad_next/iter_collections/zipfile.py
+++ b/datalad_next/iter_collections/zipfile.py
@@ -11,7 +11,6 @@ import zipfile
 from dataclasses import dataclass
 from pathlib import (
     Path,
-    PurePath,
     PurePosixPath,
 )
 from typing import Generator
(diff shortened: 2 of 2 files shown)


mih (commented):
I am not convinced that this is a good move. I understand that this switch better reflects the internal nature of a zipfile. The cost of that switch is that now `FileSystemItem` or more generally a `PathBasedItem` is no longer having a platform-specific identifier. Consumers will now have to account for that in their code.

The use case you put forth is: "Consequently, those paths cannot be used to retrieve items from a zipfile.ZipFile without conversion to a POSIX-style path"

I would argue that this is not particularly relevant here. The `ZipFile` instance to retrieve from via path is an internal detail that is not exposed. At the same time, users can set `fp=True` and get direct access to file content -- plus a platform path to write to (given an externally supplied extraction base dir).

WDYT?

mih:
I implemented https://github.com/datalad/datalad-next/pull/415 that aims to show a different approach. It solves the problem posed here, but without the issues that I pointed out above.

christian-monch:
> I am not convinced that this is a good move. I understand that this switch better reflects the internal nature of a zipfile. The cost of that switch is that now `FileSystemItem` or more generally a `PathBasedItem` is no longer having a platform-specific identifier. Consumers will now have to account for that in their code.
> 
> The use case you put forth is: "Consequently, those paths cannot be used to retrieve items from a zipfile.ZipFile without conversion to a POSIX-style path"
> 
> I would argue that this is not particularly relevant here. The `ZipFile` instance to retrieve from via path is an internal detail that is not exposed. At the same time, users can set `fp=True` and get direct access to file content -- plus a platform path to write to (given an externally supplied extraction base dir).
> 
> WDYT?

I am not sure yet, but I would like to put forward a few thoughts:

1. I like the `PurePath.as_posix()`-approach from #415. But I would like to mention the following differences between Windows and Linux:

   a. `PurePath('d/a\\b\\c.txt').as_posix() == 'd/a/b/c.txt'` (on Windows)

   b. `PurePath('d/a\\b\\c.txt').as_posix() == 'd/a\\b\\c.txt'` (on Linux)

2. The previous means that the following code snippet will fail on Windows if an item with a name like `'d/a\\b\\c.txt'`, i.e. a name that contains backslashes, exists in the archive. That means `PurePath` cannot always be used to retrieve an item, even if we use the `PurePath`-instance stored in `item.name` (the `assert` below will fail on Windows):
```python
with TarArchiveOperations('<some archive>') as archive_ops:
    for item in archive_ops:
        assert item.name in archive_ops
``` 

The same problem exists for the zip-Archive operations.

We could probably ignore this and say that we do not support archives with backslashes in directory- of file-names on Windows platforms?! ... We could also only allows strings to name members of collections?!

I know, this seems like a never-ending story. I don't want to be nagging (naggy?), but I think we have to be careful to get this right in order to reduce maintenance. I this light I actually grow more and more fond of the idea of just using `str` for names of archive-members.

mih:
There is no need to apologize for identifying a flaw that needs to be fixed.

To break this down: This issue is not in the linked https://github.com/datalad/datalad-next/pull/415, but at https://github.com/datalad/datalad-next/blob/59a6317335119eafc97d7201b9b870ad018dceb5/datalad_next/iter_collections/tarfile.py#L66, because

```py
>>> PureWindowsPath(*PurePosixPath('d/a\\b\\c.txt').parts)
PureWindowsPath('d/a/b/c.txt')
```

This means that we must relay the POSIX nature of the archive member path to the users, because there is no way to express this as a platform (windows) path -- and also no way to extract this file under an equivalent name on an FS that uses windows-semantics.

If feel like a clean(er) solution would be to change `TarFileItem` to declare to have a `name` of type `PurePosixPath` (and likewise for `ZipFileItem`. Plus some documentation at that declaration that points out the motivation. We might also consider to derive both from an intermediate `ArchiveFileSystemItem` that does that.

mih (changes requested):
I think this can be merged, once the suggested changes to the ZipfileItem class (type change) are pushed.

christian-monch:
> I think this can be merged, once the suggested changes to the ZipfileItem class (type change) are pushed.

Changes are done.

mih (approved):
Thanks, LGTM!

PR #421, opened 2023-06-15
Fix `git ls-file` parsing in `iter_gitworktree()`

christian-monch:
This PR modifies the regular expression that is used during execution of `iter_gitworktree()` in order to parse the output of `git ls-tree`. 

The current version ensures that only one tab-character is considered to separate the file info from the file name. That ensures that file names that start with tab-characters are properly handled (the previous version,  would consider all tab-characters, including the tab-characters belonging to the name, as file-info/file-name separator).

The PR includes a regression test that verifies correct handling of file names starting with tab-characters

### DIFF
--- a/datalad_next/iter_collections/gitworktree.py
+++ b/datalad_next/iter_collections/gitworktree.py
@@ -69,7 +69,7 @@ class GitWorktreeFileSystemItem(FileSystemItem):
 
 # stolen from GitRepo.get_content_info()
 _lsfiles_props_re = re.compile(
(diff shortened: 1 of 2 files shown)


mih (commented):
Thanks. This looks good to me in general. However, the test conditions needs to be adjusted. This is not about windows, but about whether the FS can handle such filenames at all -- see the crippled FS test failures.

mih (approved):
LGTM! Thx!

PR #426, opened 2023-06-21
Adapt tar-archive name type to type of `TarFileItem.name`, i.e. `PurePosixPath`

christian-monch:
This PR adapts tararchive-operations to the recently updated type of `TarFileItem.name` to `PurePosixPath`.


mih (approved):
LGTM, thx!

PR #428, opened 2023-06-21
Fix package installation in appveyor on debian platforms

christian-monch:
Currently the construction of the test environment on appveyor is failing in debian-based environments. It fails during `apt-get update` due to changed release infos for debian packages. This PR allows allows `apt-get update` to continue in changes in release info. This should fix the tests on debian-based appveyor environments.


mih (approved):
Much appreciated!!!

The should go into the extension template too, I think.

PR #430, opened 2023-06-22
Fix directory name handling in zip-iterator

christian-monch:
This PR keeps the seemingly neverending iterator item name story alive. 

The problem is that directory names in zip-archives possess a trailing '/'. We currently use `PurePosixPath` instances to represent file- and directory-entries in all types of collections. This leads to the necessity of special handling in ZipArchiveOperations, for example, when checking for containment (because `ZipFileItem.name` cannot be used directly):

```python
...
for item in iter_zip(...):
    if (str(item.name) + '/' if item.type == FileSystemItemType.directory else item.name) in zip_archive:
        ...
``` 

Instead of the "canonical" way that is used in TarArchiveOperations:
```python
...
for item in iter_tar(...):
    if item.name in tar_archive:
        ...
``` 

This PR adds the class `ZipFileDirPath`, which is a direct subclass of `PurePosixPath`. Instances of the class represent directory names in zip-archives. The difference to `PurePosixPath` is that rendering appends a '/' to the name.

This allows us to use the "canonical" way of testing for containment in ZipArchiveOperations, i.e.
```python
...
for item in iter_zip(...):
    if item.name in zip_archive:
        ...
```

### DIFF
--- a/datalad_next/iter_collections/tests/test_iterzip.py
+++ b/datalad_next/iter_collections/tests/test_iterzip.py
@@ -4,7 +4,7 @@ from pathlib import PurePosixPath
 
 from ..zipfile import (
     FileSystemItemType,
-    ZipFileDirPath,
+    _ZipFileDirPath,
     ZipfileItem,
     iter_zip,
 )
@@ -51,7 +51,7 @@ def test_iter_zip(sample_zip):
     root = PurePosixPath('test-archive')
     targets = [
         ZipfileItem(
-            name=ZipFileDirPath(root),
+            name=_ZipFileDirPath(root),
             type=FileSystemItemType.directory,
             size=0,
         ),
@@ -61,7 +61,7 @@ def test_iter_zip(sample_zip):
             size=8,
         ),
         ZipfileItem(
-            name=ZipFileDirPath(root, 'subdir'),
+            name=_ZipFileDirPath(root, 'subdir'),
             type=FileSystemItemType.directory,
             size=0,
         ),
@@ -92,9 +92,9 @@ def test_iter_zip(sample_zip):
 
 
 def test_zip_dir_path():
-    zp1 = ZipFileDirPath("a/b")
-    zp2 = ZipFileDirPath("a/b")
-    zp3 = ZipFileDirPath("a/c")
+    zp1 = _ZipFileDirPath("a/b")
+    zp2 = _ZipFileDirPath("a/b")
+    zp3 = _ZipFileDirPath("a/c")
     pp = PurePosixPath("a/b")
 
     assert zp1.as_posix()[-1] == '/'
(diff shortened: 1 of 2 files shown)


datalad_next/iter_collections/zipfile.py
```diff
@@ -75,7 +86,11 @@ def _get_zipfile_item(zip_info: zipfile.ZipInfo) -> ZipfileItem:
         else FileSystemItemType.file
     )
     return ZipfileItem(
-        name=PurePosixPath(zip_info.filename),
+        name=(
+            ZipFileDirPath(zip_info.filename)
+            if zip_info.is_dir()
```
mih, line 91: `is_dir()` is already called a few lines above, we already know whether this is a directory at this point.

datalad_next/iter_collections/tests/test_iterzip.py
```diff
@@ -87,4 +88,16 @@ def test_iter_zip(sample_zip):
         # do not compare mtime
         r.mtime = None
     for t in targets:
+        print(t.name)
```
mih, line 91: I suspect that this could/should be removed again.

mih (commented):
Thx for the PR. I think this is an elegant way to solved the issue.

Am I correct that `__contain__` is to only operation (right now) affected by this problem?

christian-monch:
> Thx for the PR. I think this is an elegant way to solved the issue.
> 
> Am I correct that `__contain__` is to only operation (right now) affected by this problem?

For all intense and purposes yes. But one can call open on a directory entry in zip-files. Directories in zip-files are just empty files with a name that ends in `/`, AFAIK. Calling `ArchiveOperations.open` with an `item.name` of a directory-item would lead to a `KeyError`.

I am not sure how relevant this is in the wild, i.e. how many people would actually try to open a directory entry.

datalad_next/iter_collections/zipfile.py
christian-monch: Fixed in db878f5d15765651ea56d8b8c6fc081478213f30

datalad_next/iter_collections/tests/test_iterzip.py
christian-monch: That is correct. Thanks. Done in 9b44e57a6fab29af7b1c864ac8ad05f6637d161c

mih:
I will rebase this to remove 9a923ce4b17c038f7dca668eb07393260ee426ca -- I assume it got in here by accident. It seems totally unrelated. Not sure if there is/was an issue about that.

mih:
It seems this implementation is incomplete. https://github.com/datalad/datalad-next/pull/561 reveals that root directories are reported with a double-slash

```
❯ dl -f json ls-file-collection zipfile /home/mih/Neuromethods.zip | head -n1
{"action": "ls_file_collection", "collection": "/home/mih/Neuromethods.zip", "fp": null, "gid": null, "item": "//", "mode": null, "mtime": 1572969506.0, "size": 0, "status": "ok", "type": "directory", "uid": null}

❯ dl ls-file-collection zipfile /home/mih/Neuromethods.zip
?---------      - None:None Nov 05 2019 // (directory)
...
```

PR #440, opened 2023-09-01
Remove internal _getcfg() helper in favour of generic SpecialRemote helper

### LINKED ISSUE #439: Turn archivist special remotes' _getcfg() into flexible helper
The internal  ``_getcfg`` method is used in the archivist remote to get special remote configuration items:
 https://github.com/datalad/datalad-next/blob/0cb44b0b5185a89b01d8a045ef1ea770c7cc1cd1/datalad_next/annexremotes/archivist.py#L349

It has a comment that says
```
    # Helpers
    #
    # TODO this could be promoted to SpecialRemote as a generic helper
    # would need standardization of remote name query in `prepare()`
```

And indeed this helper would be useful for reuse in https://github.com/datalad/datalad-ria/pull/6, where a similar legacy fallback as the archivist special remote has with the dl-archives remote is desired.

adswa:
Fixes #439

Just a maintenance change removing an internal method in favor of an already existing generic helper of the ``SpecialRemote``.

### DIFF
--- a/datalad_next/annexremotes/archivist.py
+++ b/datalad_next/annexremotes/archivist.py
@@ -168,7 +168,8 @@ class ArchivistRemote(SpecialRemote):
         # let remote-specific setting take priority (there could be
         # multiple archivist-type remotes configured), and use unspecific switch
         # as a default, with a general default of NO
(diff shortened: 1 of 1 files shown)


mih (approved):
This looks good! Thanks for catching that!

PR #457, opened 2023-09-28
Simplify (subprocess) coverage setup

### LINKED ISSUE #453: Simplify subprocess coverage setup
We use the setup from datalad core here.

However, it turns out that almost none of that is needed, and pytest-cov does almost everything out of the box.

See https://github.com/datalad/datalad-ria/pull/6 for a demo on what is neeed for adding subprocess coverage.

One thing to keep in mind when exploring coverage capture for special remote processes: We may never be able to get full coverage for "abnormal" exits. When git-annex externally terminates a special remote process, it seems "expected" that the `onexit` handlers that coverage reporting relies on, do not reliably run.

mih:
The previous setup was taken from datalad-core. It comes from a time before `pytest`. With `pytest-cov` none of this complexity should be needed anymore.

A similar setup has been implemented as part of
https://github.com/datalad/datalad-ria/pull/6

Closes #453


christian-monch (approved):
LGTM

mih:
It is not clear to me why there are coverage fluctuations at all, but here is the overview of the PR from codecov:

image

Important are `uncurl.py` and `datalad_annex.py`, because this is code that *only* runs in subprocesses. Its coverage goes up.

PR #460, opened 2023-10-03
Draft of `iter_annexworktree()`

### LINKED ISSUE #459: Implement `iterannexworktree()` for `ls-file-collection`
Somewhat related to #349, we have a need to report on annex-managed (work)tree in the shape of an iterator.

We have four substantially overlapping implementations that provide a closely related functionality:

1. `AnnexRepo.get_content_annexinfo()` (from datalad-core)
2. `gooey-status-light` (from datalad-gooey): motivation for its implementation was to achieve a substantially faster performance than `get_content_annexinfo()`
3. https://github.com/datalad/datalad-metalad/blob/dev_0.5/iter_test/iterate_dataset.py (motivation for implementation not yet determined)
4. https://github.com/psychoinformatics-de/sfb1451-projects-catalog/pull/48 (motivation for implementation not yet determined)

These for need to be investigated re functionality and performance. Ideally, one and only one implementation should be used instead.

mih:
Closes #459

TODO:

- [ ] implement `fp=True`
- [ ] implement `link_target=True`

### DIFF
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -57,6 +57,9 @@ from datalad_next.iter_collections.gitworktree import (
     GitWorktreeFileSystemItem,
     iter_gitworktree,
 )
+from datalad_next.iter_collections.annexworktree import (
+    iter_annexworktree,
+)
 
 
 lgr = getLogger('datalad.local.ls_file_collection')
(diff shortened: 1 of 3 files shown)


mih:
Until I can figure out how to have a long running command that I can feed stdin and read stdout from (via the runner), I will not be able to go further here.

I will put this in draft mode, and hope to come back in a few weeks.

PR #465, opened 2023-10-03
Rough sketch for a batch(ed) command helper

mih:
This aims to address a long-standing frustration: my cluelessness re efficient execution of (long-running) processes that execute some (query) helper command. My inability has caused much awkward code that gather information upfront, and leads to sluggish behavior of iterators and similar loop-based code that needs to gather information from more than one source.

This changeset includes a minimal wrapper around `ThreadedRunner` that seems to accomplish the goal of provisioning a batched command via a familiar context manager approach.

A basic test demos how the usage could/would look like.

### DIFF
--- a/datalad_next/runners/batch.py
+++ b/datalad_next/runners/batch.py
@@ -1,0 +1,135 @@
+"""Helpers to execute batch commands
+
+Some of the functionality provided by this module depends on specific
+"generator" flavors of runner protocols, and additional dedicated
+low-level tooling:
+
+.. currentmodule:: datalad_next.runners.batch
+.. autosummary::
+   :toctree: generated
+
+    StdOutCaptureGeneratorProtocol
+    GeneratorAnnexJsonProtocol
+    _ResultGenerator
+"""
+
+from __future__ import annotations
+
+from contextlib import contextmanager
+
+from pathlib import Path
+
+from queue import Queue
+
+from typing import (
+    Any,
+    Generator,
+)
+
+from datalad.runner.nonasyncrunner import _ResultGenerator
+from datalad.support.annexrepo import GeneratorAnnexJsonProtocol
+
+from . import (
+    Protocol,
+    ThreadedRunner,
+)
+from .protocols import StdOutCaptureGeneratorProtocol
+
+
+class BatchProcess:
+    """Representation of a (running) batch process
+
+    An instance of this type is produced by any of the context manager variants
+    in this module. It is a convenience wrapper around an instance of a
+    :class:`_ResultGenerator` that is produced by a :meth:`ThreadedRunner.run`.
+
+    A batch process instanced is used by passing ``bytes`` input to its
+    ``__call__()`` method, and receiving the batch output as return value.
+
+    While the process is still running, it ``return_code`` property will
+    be ``None``. After the has terminated, the property will contain the
+    respective exit code.
+    """
+    def __init__(self, rgen: _ResultGenerator):
+        self._rgen = rgen
+
+    def __call__(self, data: bytes) -> Any:
+        self._rgen.runner.stdin_queue.put(data)
+        return next(self._rgen)
+
+    @property
+    def return_code(self) -> None | int:
+        return self._rgen.return_code
+
+
+@contextmanager
+def batchcommand(
+    cmd: list,
+    protocol_class: Protocol,
+    cwd: Path | None = None,
+) -> Generator[BatchProcess, None, None]:
+    """Generic context manager for batch processes
+
+    ``cmd`` is an ``argv``-list specification of a command. It is executed via
+    a :class:`~datalad_next.runners.ThreadedRunner` runner instance.  This
+    runner is parameterized with ``protocol_class`` (which can take any
+    implementation of a DataLad runner protocol), responsible for
+    (pre)processing the command's output.
+
+    On leaving the context, the manager takes care of closing ``STDIN``
+    of the underlying process, typically causing it to exit normally.
+
+    While this generic context manager can be used directly, it can be
+    more convenient to use any of the more specialized implementations
+    that employ a specific protocol (e.g., :func:`stdout_batchcommand`,
+    :func:`annexjson_batchcommand`).
+    """
+    input_q = Queue()
+    proc = ThreadedRunner(
+        cmd=cmd,
+        protocol_class=protocol_class,
+        stdin=input_q,
+        cwd=cwd).run()
+    try:
+        bp = BatchProcess(proc)
+        yield bp
+    finally:
+        input_q.put(None)
+        # TODO should we block until `bp.return_code != None`?
+        # TODO should we force terminate `bp` after a grace period?
+
+
+def stdout_batchcommand(
+    cmd: list,
+    cwd: Path | None = None,
+) -> Generator[BatchProcess, None, None]:
+    """Context manager for commands that produce arbitrary output on ``stdout``
+
+    Internally this calls :func:`batchcommand` with the
+    :class:`StdOutCaptureGeneratorProtocol` protocol implementation.
+    """
+    return batchcommand(
+        cmd,
+        protocol_class=StdOutCaptureGeneratorProtocol,
+        cwd=cwd,
+    )
+
+
+def annexjson_batchcommand(
+    cmd: list,
+    cwd: Path | None = None,
+) -> Generator[BatchProcess, None, None]:
+    """Context manager for git-annex commands that support ``--batch --json``
+
+    The given ``cmd``-list must be complete, i.e., include
+    ``git annex ... --batch --json``, and any additional flags that may be
+    needed.
+
+    Internally this calls :func:`batchcommand` with the
+    :class:`GeneratorAnnexJsonProtocol` protocol implementation.
+    """
+    return batchcommand(
+        cmd,
+        protocol_class=GeneratorAnnexJsonProtocol,
+        cwd=cwd,
+    )
--- a/datalad_next/runners/tests/test_batch.py
+++ b/datalad_next/runners/tests/test_batch.py
@@ -1,0 +1,54 @@
+import pytest
+
+from .. import CommandError
+
+from ..batch import (
+    annexjson_batchcommand,
+    stdout_batchcommand,
+)
+
+
+def test_batch_simple(existing_dataset):
+    # first with a simplistic protocol to test the basic mechanics
+    with stdout_batchcommand(
+        ['git', 'annex', 'examinekey',
+         # the \n in the format is needed to produce an output that hits
+         # the output queue after each input line
+         '--format', '${bytesize}\n',
+         '--batch'],
+        cwd=existing_dataset.pathobj,
+    ) as bp:
+        res = bp(b'MD5E-s21032--2f4e22eb05d58c21663794876dc701aa\n')
+        assert res.rstrip(b'\r\n') == b'21032'
+        # to subprocess is still running
+        assert bp.return_code is None
+        # another batch
+        res = bp(b'MD5E-s999--2f4e22eb05d58c21663794876dc701aa\n')
+        assert res.rstrip(b'\r\n') == b'999'
+        assert bp.return_code is None
+        # we can bring the process down with stupid input
+        with pytest.raises(CommandError) as deadexc:
+            bp(b'stupid\n')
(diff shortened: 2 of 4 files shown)


datalad_next/runners/tests/test_batch.py
```diff
@@ -0,0 +1,30 @@
...
+    with annexjson_batchcommand(
+        ['git', 'annex', 'examinekey', '--json',
+         '--batch'],
+        cwd=existing_dataset.pathobj,
+    ) as bp:
+        res = bp.proc(b'MD5E-s21032--2f4e22eb05d58c21663794876dc701aa\n')
+        # TODO the outcome should be the "result" of the protocol, which
+        # would be a decoded JSON data structure, not just the string
+        # read from the process stdout
+        assert res.rstrip(b'\r\n') == b'{"backend":"MD5E","bytesize":"21032","error-messages":[],"hashdirlower":"20f/01d/","hashdirmixed":"00/7F/","humansize":"21.03 kB","key":"MD5E-s21032--2f4e22eb05d58c21663794876dc701aa","keyname":"2f4e22eb05d58c21663794876dc701aa","mtime":"unknown","objectpath":".git/annex/objects/00/7F/MD5E-s21032--2f4e22eb05d58c21663794876dc701aa/MD5E-s21032--2f4e22eb05d58c21663794876dc701aa","objectpointer":"/annex/objects/MD5E-s21032--2f4e22eb05d58c21663794876dc701aa\\n"}'
```
mih, line 30: @christian-monch what would need to change in order to get the actual protocol result here?

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,51 @@
+"""Helpers to execute batch command
+"""
+from contextlib import contextmanager
+
+from queue import Queue
+
+from datalad.support.annexrepo import AnnexJsonProtocol
```
christian-monch, line 7:
```suggestion
from datalad.support.annexrepo import GeneratorAnnexJsonProtocol
```

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,51 @@
...
+        self._rungen.runner.stdin_queue.put(data)
+        return next(self._rungen)
+
+
+@contextmanager
+def batchcommand(cmd, protocol_class, cwd=None):
+    input_q = Queue()
+    proc = ThreadedRunner(
+        cmd=cmd,
+        protocol_class=StdOutCaptureGeneratorProtocol,
```
christian-monch, line 29:
```suggestion
        protocol_class=protocol_class,
```

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class BatchProcess:
+    """Representation of a (running) batch process
+
+    An instance of this type is produced by any of the context manager variants
+    in this module. It is a convenience wrapper around an instance of a
+    :class:`_ResultGenerator` that is produced by a :meth:`ThreadedRunner.run`.
+
+    A batch process instanced is used by passing ``bytes`` input to its
```
mih, line 46:
```suggestion
    A batch process instance is used by passing ``bytes`` input to its
```

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,135 @@
...
+    ``__call__()`` method, and receiving the batch output as return value.
+
+    While the process is still running, it ``return_code`` property will
```
mih, line 49:
```suggestion
    While the process is still running, it's ``return_code`` property will
```

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,135 @@
...
+    be ``None``. After the has terminated, the property will contain the
```
mih, line 50:
```suggestion
    be ``None``. After the has process terminated, the property will contain the
```

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,135 @@
...
+        protocol_class=protocol_class,
+        stdin=input_q,
+        cwd=cwd).run()
+    try:
+        bp = BatchProcess(proc)
+        yield bp
+    finally:
+        input_q.put(None)
+        # TODO should we block until `bp.return_code != None`?
+        # TODO should we force terminate `bp` after a grace period?
```
mih, line 99: @christian-monch do we have a "standard behavior" that should be implemented here?

datalad_next/runners/batch.py
```diff
@@ -0,0 +1,135 @@
...
+    While the process is still running, it ``return_code`` property will
+    be ``None``. After the has terminated, the property will contain the
+    respective exit code.
+    """
+    def __init__(self, rgen: _ResultGenerator):
+        self._rgen = rgen
+
+    def __call__(self, data: bytes) -> Any:
+        self._rgen.runner.stdin_queue.put(data)
+        return next(self._rgen)
```
christian-monch, line 58: Just as a remark. This method will stall, if the generator has no data stored and if `data` does not trigger any output from the underlying process.

Just returning `next(self._rgen)` might not work, if the result of processing `data` is split into multiple parts (either because the process delivered multiple parts, or because a communication channel split them up). For example, if any output is split into two parts, the next calls to `BatchProcess.__call__()` would return the second half of the answer of the previous call. The current code will most likely work with stdout and single line-results. But it would not work with multi-line JSON-results. In the latter case (and in general cases) the end of a response should not be determined by the chunking of the data that is returned by the running process.

For example, a test like this:
```
def test_batch_multi():
    # first with a simplistic protocol to test the basic mechanics
    with stdout_batchcommand(
            ['python', '-c', '''
import time
while True:
        a = input()
        print("you entered", end='', flush=True)
        time.sleep(1)
        print("->", a, flush=True)
''']) as bp:
        res = bp(b'A\n')
        print('A:', res)
        res = bp(b'B\n')
        print('B:', res)
        res = bp(b'C\n')
        print('C:', res)
```
Might print:
```
A: b'you entered'
B: b'->'
C: b' A\nyou entered'
```

This might lead to problems.

Since the caller of `BatchCommand`.__call__ has no access to the generator, a solution might be to add a method that detects how much of the returned data belongs to a proper response, whether the response is complete, and how much data belongs to the next response. This method could be default detect line-ends and could be overriden (or passed in as an argument) to support other response structures, e.g. multi-line JSON.

That would require to buffer any result-data that belongs to the next repsonse

 (There is  a `PatternFilter` in `datalad-ria.sshshell` (this branch:  https://github.com/datalad/datalad-ria/tree/ssh-shell) that detects pattern, e.g. '\n', in streams)

mih: > Just as a remark. This method will stall, if the generator has no data stored and if data does not trigger any output from the underlying process.

That was the idea: block until there is something. In the context of batch commands this is what I would expect. Am I wrong?

> Since the caller of BatchCommand.call has no access to the generator, a solution might be to add a method that detects how much of the returned data belongs to a proper response [...]

Can you point to any implementation that uses `ThreadedRunner` and does this correctly?

I believe what you are saying is that there cannot be an implementation that is generic and unaware of the nature/volume of the output.

Intuitively, I'd say: This is the domain of the `Protocol` implementation. For example `AnnexJsonProtocol` performs the tests that you listed. However, it is totally unclear to me what `next()` is actually calling.

For the example that you gave, I do not see how this problem can be resolved. Therefore, I conclude that there should not be a ` stdout_batchcommand`, but a ` stdoutline_batchcommand`, and `batchcommand` should get a warning that outlines the general problem.

However, before I can approach this, it would be good to know at which level the protocol implementation can be effective in preventing this problem.

christian-monch: > > Just as a remark. This method will stall, if the generator has no data stored and if data does not trigger any output from the underlying process.
> 
> That was the idea: block until there is something. In the context of batch commands this is what I would expect. Am I wrong?
> 
Not at all, that is perfectly fine. Just keep in mind that it puts the two processes into lockstep.

> > Since the caller of BatchCommand.call has no access to the generator, a solution might be to add a method that detects how much of the returned data belongs to a proper response [...]
> 
> Can you point to any implementation that uses `ThreadedRunner` and does this correctly?
> 

I would say `datalad.cmd.BatchedCommandProtocol` does the correct thing for a protocol that returns lines of output. It buffers lines in its `self.line_splitter` and sends each invidual line as a result to the generator (via `self.send_result`).

> [...]
> I believe what you are saying is that there cannot be an implementation that is generic and unaware of the nature/volume of the output.
> [...]

Depends on what you define by generic. Anything that is more structured than "streams of bytes with undetermined length" requires some processing. Still, that could be done in a generic fashion in the sense that we split the boilerplate code (assembling response data, decoding response data, ...) from the code that detects the response structure and can pinpoint individual response-"items". 


> Intuitively, I'd say: This is the domain of the `Protocol` implementation. For example `AnnexJsonProtocol` performs the tests that you listed. [...]

Yes, that would be another example besides the  `datalad.cmd.BatchedCommandProtocol`-example from above.

> [...]
However, it is totally unclear to me what `next()` is actually calling.
> 

`next()`  ends up calling `send` of `_ResultGenerator` (via the `__next__`-method of the `Generator`-baseclass of `_ResultGenerator`. The result generator will look into its internal queue to check for pending results, if none are there, it will go to the runner to fetch the next result. The runner might block until the next result (data or timeout) is available.

> For the example that you gave, I do not see how this problem can be resolved. Therefore, I conclude that there should not be a ` stdout_batchcommand`, but a ` stdoutline_batchcommand`, and `batchcommand` should get a warning that outlines the general problem. [...]

Yes, we might craft specific commands for specific response-structures. But we might also use a single command and delegate response structure "parsing" to either the protocol or the command.

> [...]
> However, before I can approach this, it would be good to know at which level the protocol implementation can be effective in preventing this problem.

I think `datalad.cmd.BatchedCommandProtocol` is not a bad example. It could be used to deliver single-line results, as intended by `stdoutline_batchcommand`. If you use `datalad.cmd.BatchedCommand` as protocol the previously described problems go away, i.e. the generator will only return single, complete lines (if you want to really use the class from `datalad.cmd` to test this, then, due to some specific timeout handling in `datalad.cmd` the use of `datalad.cmd.BatchedCommand` requires to instantiate `ThreadedRunner` with the keyword argument `protocol_kwargs={'batched_command': None}`. This will fail, if a timeout occurs, but will allow testing).



For a more general solution, a number of aspects have to be taken into account,  intended responsiveness (how much does the protocol buffer), expected size of the results, etc. Given my recent adventure with SSHShell and the need to detect complex patterns, I think a generic way for most applications would be the following:

1. Define a relatively simple protocol, much like `datalad.cmd.BatchedCommand` but without the LineSplitter-class.
2. Parameterize the protocol with an object that supports parsing data chunks in order to detect a response, for example, a method called `parse` which receives a data chunk and returns a triple consisting of optional response data, an indication if a response is complete, and optional remaining data from the data chunk:
```
     response_data, response_complete, remaining_data = parse(data_chunk).
```
This object can be provided by the user. In `datalad.cmd.BatchedCommand` the `LineSplitter`-class fulfills this role. If the expected responses are short, response_data would only ever be not-None if the response is complete. If response data is expected to be large, e.g. the result of a remote `cat <file>` operation, response data can already be handed to a caller before the response is complete.

I will leave it at this for now, but I think if we get this right in `datalad_next`, we can support a large number of use cases with a minimal amount of use-case specific code.

 


4.

mih: Thanks for the explanation and the pointer to `BatchedCommandProtocol` -- it did not know this existed or that `cmd.py` still contained stuff other than legacy imports.

I conclude the following:

- The general idea of having such a context manager like in this PR seems to be compatible with you statements.
- Robustness of behavior is largely the responsibility of protocol implementations

I will gut this PR to remove any code that exposes functionality with the simplistic "report report stdout" protocol and stick with the annex json one as the only helper.

This will be sufficient to get me going, and everything else can come later.

I saw no complaint about the general composition/implementation of the context manager, nor the proposed API.

mih:
I am closing this PR. I have adjusted the tests to use `stdout_batchcommand` only as a test helper. I also added a test that produces "chunked" JSON lines to check whether the protocol can handle this reliably. When doing so I discovered that the present implementation of `BatchProcess` is not even able to properly report `return_code` when a process exits even before anything is read from `stdin`.

I conclude that this is not a viable path forward.

I will rebase #469 to remove this changeset from it. And then also close that PR, because functionality like this is needed for it.

PR #469, opened 2023-10-04
Draft of iter_annexworktree()

### LINKED ISSUE #459: Implement `iterannexworktree()` for `ls-file-collection`

mih:
Sit on top of #465 

Closes #459

### DIFF
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -57,6 +57,9 @@ from datalad_next.iter_collections.gitworktree import (
     GitWorktreeFileSystemItem,
     iter_gitworktree,
 )
+from datalad_next.iter_collections.annexworktree import (
+    iter_annexworktree,
+)
 
 
 lgr = getLogger('datalad.local.ls_file_collection')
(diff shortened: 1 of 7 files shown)


mih:
I am closing this PR, because #465 failed to provide a helper to run a batch command -- at least one that I could understand.

Given that the original motivation for providing a faster dataset content lister is somewhate gone (`datalad status --annex` was found to be fast enough), this no longer seems like a useful way to spend energy.

PR #471, opened 2023-10-04
Add gitworktree to ls-file-collection's docstring

### LINKED ISSUE #470: `ls_file_collection` docstring does not mention all supported collections
Under "Supported file collection types are", the docstring for `ls_file_collection` lists only `directory` & `tarfile`. The `gitworktree` collection type is supported, but missing:

```
_supported_collection_types = (
    'directory',
    'tarfile',
    'gitworktree',
)
```

mslw:
This PR will include `gitworktree` among the available file collection types listed in `ls-file-collection`'s docstring. Fixes #470 

The proposed docstring favours brevity, but does not cover nuances which could be included:
- `hash` being required for `size` to be reported
- reporting being based on `git ls-files --stage --cached --exclude-standard --others`

### DIFF
--- a/changelog.d/20231004_180525_mslw.md
+++ b/changelog.d/20231004_180525_mslw.md
@@ -1,0 +1,6 @@
+### 📝 Documentation
+
+- Include `gitworktree` among the available file collection types
+  listed in `ls-file-collection`'s docstring.  Fixes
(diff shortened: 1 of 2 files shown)


mih (approved):
Thx!

At some point we need to think about limiting the number of text copies.

PR #473, opened 2023-10-08
`run()` context manager convenience for subprocess execution

mih:
This PR aims to pick up after the failed #465 -- with me continuing to try to understand the runner usage.

This changeset introduces a `run()` convenience context manager that can be used with generator on non-generator protocols. A set of tests documents that.

The primary purpose here is to establish a uniform way to use subprocesses safely and easily, while ensuring that such processes actually terminate when they are no longer needed. Previously, a persistent runner object was the primary (only?) possibility, and cleanup would need to be implemented on a case-by-case basis. This is error-prone and needlessly verbose.

TODO:

- [x] figure out, if this works
- [x] import unpublished test from #465 attempt for handling of a process that must terminate before anything is read from it (while using a generator protocol)
- [x] determine optional take-down approach for `finally`
- [ ] figure out and test, how the context manager could handle an external STDIN closing. Presently, we require an input `Queue` to be closed manually (`put(None)`). This crashes a `communicate()`: I/O operation on closed file). However, when we do not close `stdin`, a batch/upload process has no indication that it shall stop before it received a SIGTERM, which might stop an operation that should still be completed (upload the last chunk, etc)

### DIFF
--- a/datalad_next/runners/run.py
+++ b/datalad_next/runners/run.py
@@ -27,6 +27,8 @@ def run(
     *,
     cwd: Path | None = None,
     input: int | IO | bytes | Queue[bytes | None] | None = None,
+    # only generator protocols make sense for timeout, and timeouts are
+    # only checked when the generators polls
     timeout: float | None = None,
 ) -> dict | _ResultGenerator:
     runner = ThreadedRunner(
--- a/datalad_next/runners/run.py
+++ b/datalad_next/runners/run.py
@@ -9,7 +9,6 @@ from queue import Queue
 from subprocess import DEVNULL
 from typing import (
     IO,
-    Optional,
 )
 
 from datalad.runner.nonasyncrunner import _ResultGenerator
@@ -28,7 +27,7 @@ def run(
     cwd: Path | None = None,
     input: int | IO | bytes | Queue[bytes | None] | None = None,
     # only generator protocols make sense for timeout, and timeouts are
-    # only checked when the generators polls
+    # only checked when the generator polls
     timeout: float | None = None,
 ) -> dict | _ResultGenerator:
     runner = ThreadedRunner(
@@ -46,6 +45,12 @@ def run(
         # already be the case -- we make sure that now zombies
         # accumulate
         if runner.process is not None:
+            # TODO figure out what is the most graceful way to
+            # tell a process to stop. Possibly
+            # - process.terminate()
+            # - process.wait(with timeout)
+            # - catch TimeoutExpired exception and process.kill()
+            #
             # send it the KILL signal
             runner.process.kill()
             # wait till the OS has reported the process dead
--- a/datalad_next/runners/run.py
+++ b/datalad_next/runners/run.py
@@ -46,4 +46,7 @@ def run(
         # already be the case -- we make sure that now zombies
         # accumulate
         if runner.process is not None:
+            # send it the KILL signal
             runner.process.kill()
+            # wait till the OS has reported the process dead
+            runner.process.wait()
--- a/datalad_next/runners/run.py
+++ b/datalad_next/runners/run.py
@@ -6,7 +6,10 @@ from __future__ import annotations
 from contextlib import contextmanager
 from pathlib import Path
 from queue import Queue
-from subprocess import DEVNULL
+from subprocess import (
+    DEVNULL,
+    TimeoutExpired,
+)
 from typing import (
     IO,
 )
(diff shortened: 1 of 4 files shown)


datalad_next/runners/run.py
```diff
@@ -0,0 +1,64 @@
...
+        timeout=timeout,
+    )
+    try:
+        yield runner.run()
+    finally:
+        # if we get here the subprocess has no business running
+        # anymore. When run() exited normally, this should
+        # already be the case -- we make sure that no zombies
+        # accumulate
+        if runner.process is not None:
```
christian-monch, line 50: Generally, the result generator should get a chance to pull all data from the subprocess by calling something like `tuple(result_generator)`. Threaded runner has support for timeouts and that includes processes. The runner will 
`terminate`, but not `kill`, a process if the timeout handler returns `True` for a process timeout (that is a timeout where the `fd` is `None).  If the process triggers multiple timeouts, the timeout-callback could theoretically kill it. 

Having said that, the code here should work, as far as I can see. And it actually introduces "killing", which the ThreadedRunner does not automatically invoke (that would have to be done as a result of timeout-calls)

christian-monch:
> - [ ] figure out and test, how the context manager could handle an external STDIN closing. Presently, we require an input `Queue` to be closed manually (`put(None)`). This crashes a `communicate()`: I/O operation on closed file). However, when we do not close `stdin`, a batch/upload process has no indication that it shall stop before it received a SIGTERM, which might stop an operation that should still be completed (upload the last chunk, etc)

I might not completely understand the problem. Let me describe what I understand:

##### Scenario 1

1. there is a process that is running as long as its stdin is not closed.
2. there is an input queue connected to stdin of the process, if `None` is enqueued, the queue-reading thread will eventually close stdin of the subprocess.
3. the process is executed by a `run()` context manager.
4. the __exit__-function of the context manager is called, but `None` has NOT been enqueued to the input queue.
5. as a result, the process will either never end, or we terminate it (which gives the process the option to try to finish pending tasks), or we kill it, which would interrupt all processing and might lead to inconsistent states.

##### Comments on scenario 1

It seems that, for the process described in point 1,  the step in point 4. is not "correct". If the process behaves as described in 1., then its stdin should be closed before the context manager is exited. So the code should look something like this

```python
q = queue.Queue()
with run(stdin=q, ...) as rp:
    # do something 
        ...

    # close stdin
    q.put(None)
```

If that were the case, a generic `tuple(result_generator)` call in the __exit__-method of the `run()` context manager would wait until the process exits. The process might still go rogue and fail to exit for a number of reasons, e.g. bugs, maliciousness, locked synchronisation. In this case the timeout-based termination and killing is one option to continue with the "calling" process. Termination and killing might require some process-specific cleanup though. It should be noted that "closing stdin" might no be the way to end a process. A process might require a specific message and ignore stdin-closing. The point is, that proper process termination might require process specific operations. That is why point 4. above should not be reached.

In any case, the stdin-writer thread of ThreadedRunner would detect a closed stdin-descriptor and terminate itself. So random closing of stdin by the subprocess would not lead to an exception.

datalad_next/runners/run.py
```diff
@@ -0,0 +1,64 @@
...
+        protocol_class=protocol_class,
+        stdin=DEVNULL if input is None else input,
+        cwd=cwd,
+        timeout=timeout,
+    )
+    try:
+        yield runner.run()
+    finally:
+        # if we get here the subprocess has no business running
+        # anymore. When run() exited normally, this should
```
christian-monch, line 47: > \# if we get here the subprocess has no business running
        \# anymore. [...]

I am not sure that that is necessarily true. That would only be true if the code that is executed in the run-context has a way to ensure that the subprocess is actually done.  Either by means of communication with the process or by a `Popen.wait()`. If the context-code can ensure that, then it would be justified to always terminate() and possibly kill() the subprocess.

Is it the intended semantics that the code that is executed in the run-context ensures that the subprocess is actually done?!

datalad_next/runners/run.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@contextmanager
+def run(
+    cmd: list,
+    protocol_class: Protocol,
+    *,
+    cwd: Path | None = None,
+    input: int | IO | bytes | Queue[bytes | None] | None = None,
+    # only generator protocols make sense for timeout, and timeouts are
```
christian-monch, line 32: Timeouts make sense in non-generator execution as well, because they would trigger a call to the timeout-method of the protocol. The protocol implementation could then, for example, raise an exception.

mih: They make sense. But I found what you describe to not work/be implemented. I also found no test that would verify such behavior. I assume you write this from a perspective of a future development plan only.

datalad_next/runners/run.py
mih: > I am not sure that that is necessarily true.

I think so. This is a context manager. User using a context manager can and should have the expectation that the resources managed by in a context are freed/closed/whatever on exit.

What that means for an internal implementation is another business. but a context manager is typically used to encapsulate all the complexities that you point out, such that a user need not constantly be aware of everything.

datalad_next/runners/run.py
mih: It does not. I tried to use it for the SSH uploader implementation in here, and it fails to upload stuff.

PR #474, opened 2023-10-09
Approach to fix the batch isssues raised PR #473

christian-monch:
This PR contains an "explorational" version of the `batchprocess`-implementation that was started in PR #473. The PR:

- makes `finalize` work, i.e. process-status codes are harvested -> no zombies
- uses pluggable processors that convert input data in the final result, e.g. combine input data, split lines, convert lines to JSON-record.

This is intended as "take a look and see what you think". Doc-strings and comments are not in a final state.


mih:
Can you post an update what the state here is, please? While this is based on #473 is does not include any of its commits, hence I cannot see what is different without an exploration. But it looks as if only minimal code is retained, so that is fine. The amount of additional classes and and logic suggest that you believe none of the existing implementations in datalad-core can be reused or recombined to help here.

mih:
I believe this is superseded by #484. I could be wrong, so please reopen if needed. But we have so many unfinished things here, I'd rather err on that side.

PR #479, opened 2023-10-13
More properly present docs on the repo classes

mih:
Explicitly list the few `call_annex()` methods we aim to support. Most others (`annex_records()` etc) are better covered by iterators.


mih:
Took off the `minor` label. #487 makes it obsolete

PR #480, opened 2023-10-13
Remove call_annex_success(), available in datalad-core since v0.18.3

mih:


mih:
Let's not do this now

PR #484, opened 2023-10-13
A run context manager

christian-monch:
This PR adds a run-context manager. It is currently a draft PR, because there are open TODOs, but I wanted to start discussions early and would like to highlight that any input on this PR is appreciated.

#### What's in it

The run-context manager starts a subprocess and returns either a result structure or a generator to the code executed in the context.

If the protocol that is used to create the context does not inherit `GeneratorProtocolMixIn`, the context manager will execute the specified subprocess and, if the subprocess exits, return the result of `protocol._prepare_result()` in the `as`-variable.

If the protocol that is used to create the context does inherit `GeneratorProtocolMixIn`, the context manager will start the execution of the process and return the resulting generator in the `as`-variable. The generator can be used to fetch output from the subprocess and to trigger protocol callbacks, e.g. `pipe_data_received`, `timeout` etc. If the code exits the context, the context-manager will execute `tuple()` on the generator to ensure that the subprocess exit is properly handled, i.e. protocol-callbacks are called and the return code of the subprocess is fetched. 

The context handler supports the specification of terminate- and/or kill-times. If those are specified the context manager might send a terminate and/or a kill signal to the subprocess. The following happens if terminate- and/or kill-time are specified:

- In the non-generator case, the context manager will send a terminate-signal to the process if it does not exit with the terminate-time number of timeouts (if no timeout was specified by the caller, the context manager will use 1 second). It will send a kill-signal, if the process does not exist within (terminate-time + kill-time) number of timeouts.
- In the generator case, the context manager will start the terminate/kill logic after the code in the context left the context. The context manager will try to exhaust the generator and send a terminate or kill signal to the process, if the termination or kill condition was reached before the generator was exhausted.

The PR comes with a number of tests that demonstrate the behavior.

#### Data processors

The PR includes an implementation of a generator that can be configured to read data from a base generator, perform different steps of data processing, and yield the result of the last processing step to the caller. There are currently three data processors: 

- A "decode" processor that decodes bytes into strings
- A  "linesplit" processor that splits bytes or strings at line-ends or at the file-end
- A "jsonline" processor that converts lines into JSON-objects if possible.

Which processors are used is defined by putting them into a list. For example, the list `[decode_processor, splitlines_processor]` specifies that the content that is received from the underlying generator should first be decoded and then split into lines. The processing is "stream"-oriented, which means that data is returned as soon as possible, e.g. each individual line is returned the caller as soon as it was read.

#### Implementation remark

The run-context manager creates a subclass of the provided protocol class in order to amend the provided protocol class with the kill-logic. It uses this subclass to instantiate a runner. This construction allows the user to use any protocol class without the need to define yet another class, while it ensures that the class that is used by the runner supports the kill-logic.

#### TODO
- [ ] Improve documentation
- [x] Use the run-context manager in respective source code in datalad-next


datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+    Protocol,
+    ThreadedRunner,
+)
+
+
+def create_kill_wrapper(cls: type[Protocol]) -> type[Protocol]:
+    """ Extend `cls` to supports the "kill-interface"
+
+    This function creates a subclass of `cls` that contains the components
+    of the "kill-interface".  The two major components are amethod called `arm`,
```
mih, line 34:
```suggestion
    of the "kill-interface".  The two major components are a method called `arm`,
```

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+)
+
+from datalad_next.runners import (
+    GeneratorMixIn,
+    Protocol,
+    ThreadedRunner,
+)
+
+
+def create_kill_wrapper(cls: type[Protocol]) -> type[Protocol]:
```
mih, line 30: Looks like this is an internal helper

```suggestion
def _create_kill_wrapper(cls: type[Protocol]) -> type[Protocol]:
```

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+    Parameters
+    ----------
+    cls : type[Protocol]
+        A protocol class that should be extended by the kill-interface
+
+    Returns
+    -------
+    Any
+        If the `cls` is not a subclass of `GeneratorMixIn`, the result of
+        `cls._prepare_result` will be returned.
```
mih, line 48: Do you know why such an essential functionality is provided by a private method?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+        input: int | IO | bytes | Queue[bytes | None] | None = None,
+        timeout: float | None = None,
+        terminate_time: int | None = None,
+        kill_time: int | None = None,
+) -> Any | Generator:
+
+    introduces_timeout = False
+    if timeout is None:
+        introduces_timeout = True
+        timeout = 1.0
```
mih, line 115: What is the logic here? Unconditional timeouts? Would this even work with any long-running process?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+@contextmanager
+def run(
+        cmd: list,
+        protocol_class: type[Protocol],
+        *,
+        cwd: Path | None = None,
+        input: int | IO | bytes | Queue[bytes | None] | None = None,
+        timeout: float | None = None,
+        terminate_time: int | None = None,
+        kill_time: int | None = None,
```
mih, line 109: I think we need a precise definition in the documentation.

What is the difference between `timeout` and `terminate_time`?

Is `kill_time` on top of `terminate_time`?

What is the semantic difference that leads to one being called `timeout` and the other just be `time`?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+    introduces_timeout = False
+    if timeout is None:
+        introduces_timeout = True
+        timeout = 1.0
+
+    armed = (
+        False
+        if issubclass(protocol_class, GeneratorMixIn)
+        else True
+    )
```
mih, line 121: So `armed` means, it is a generated protocol. Why introduce a new term for that?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Create the wrapping class. This is done mainly to ensure that the
+    # termination-related functionality is present in the protocol class that
+    # is used, independent of the actual protocol class that the user passes as
+    # argument.
+    # A side effect of this approach is, that the number of protocol class
+    # definitions is reduced, because the user does not need to define
+    # terminate-capable protocols for every protocol he wants to use.
```
mih, line 129: Would this also be applicable to the case of generator protocol in general? Here we require exactly that "need to define <generator>-capable protocols for every protocol"

mih, line 129:
```suggestion
    # A side effect of this approach is, that the number of protocol class
    # definitions is reduced, because the user does not need to define
    # terminate-capable protocols for every protocol they want to use.
```

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+    kill_protocol_class = create_kill_wrapper(protocol_class)
+
+    runner = ThreadedRunner(
+        cmd=cmd,
+        protocol_class=kill_protocol_class,
+        stdin=DEVNULL if input is None else input,
+        protocol_kwargs=dict(
+            introduced_timeout=introduces_timeout,
```
mih, line 137: Is the change in tense on purpose?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+            encoding=None
+        ),
+        timeout=timeout,
+        exception_on_error=False,
+        cwd=cwd,
+    )
+    result = runner.run()
+    # We distinguish between a non-generator run, i,e. a blocking run and
+    # a generator run.
+    if not issubclass(protocol_class, GeneratorMixIn):
```
mih, line 151: Same as `armed is True`?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+        finally:
+            # Arm the protocol, that will enable terminate signaling or kill
+            # signaling, if terminate_time or kill_time are not None.
+            result.runner.protocol.arm()
+            # Exhaust the generator. Because we have set a timeout, this will
+            # lead to invocation of the timeout method of the wrapper, which
+            # will take care of termination or killing. And it will fetch
+            # the result code of the terminated process.
+            # NOTE: if terminate_time and kill_time are both None, this might
+            # loop forever.
```
mih, line 165: Something to protect against? Surely documentation is needed.

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+            tuple(result)
```
mih, line 166: What is the significance of `tuple()` here. It would collect all the remaining results into memory. This could be a lot.

In my previous attempts I used `proc.communicate()`, which seems to aim for the "same" thing, possibly on a lower level.

datalad_next/iter_collections/gitworktree.py
```diff
@@ -250,23 +255,27 @@ def _lsfiles_line2props(
...
+            ],
+            protocol_class=StdOutCaptureGeneratorProtocol,
+            input=DEVNULL,
+            cwd=path
+    ) as r:
+        # This code uses the data processor chain to process data. This fixes
+        # a problem with the previous version of the code, where `decode` was
+        # used on every data chunk that was sent tp `pipe_data_received`. But
+        # data is chunked up randomly and might be split in the middle of a
+        # character encoding, leading to weird errors.
```
mih, line 274: This comments make me think whether could be any legitimate use of `run()` without the decoder construct below, whenever output is to be processed?

I think not.

This raises the questions if anything useful/typical could/would be done with `r` (context manager yield), before it is consumed by `process_from()`? If there isn't something, then this would suggest that spiltting these components up may make sense internally, but is needless complexity for a user to deal with. `processors=` could become an argument of `run()`.

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
...
+        try:
+            result.append((True, json.loads(line)))
+        except json.decoder.JSONDecodeError:
+            result.append((False, lines))
+    return result, []
+
+
+def splitlines_processor(data_chunks: StrOrBytesList,
+                         final: bool = False
+                         ) -> tuple[StrOrBytesList, StrOrBytesList]:
```
mih, line 87: What is the relationship to `SplitLinesProcessor`? That one is undocumented but used. This one is documented but unused outside the tests, it seems.

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
...
+
+StrList = List[str]
+BytesList = List[bytes]
+StrOrBytes = Union[str, bytes]
+StrOrBytesList = List[StrOrBytes]
+
+
+def jsonline_processor(lines: StrOrBytesList,
+                       _: bool = False
+                       ) -> tuple[list[tuple[bool, Any]], StrOrBytesList]:
```
mih, line 58: Seems tested, but is otherwise unused.

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
...
+
+    # At this point we have joined whatever we can join. We still have to check
+    # whether the last chunk ends with a pattern-prefix.
+    if not final:
+        if ends_with_pattern_prefix(data_chunks[-1], pattern):
+            return data_chunks[:-1], data_chunks[-1:]
+    return data_chunks, []
+
+
+class SplitLinesProcessor:
```
mih, line 156: Undocumented

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
...
+            for line in self.line_splitter.process(data)
+        ]
+        if final:
+            remaining = self.line_splitter.finish_processing()
+            if remaining:
+                result.append(remaining)
+        return result, []
+
+
+class DecodeProcessor:
```
mih, line 179: Undocumented

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
...
+        try:
+            text = (b''.join(data_chunks)).decode(self.encoding)
+        except UnicodeDecodeError:
+            if final:
+                raise
+            return [], data_chunks
+        return [text], []
+
+
+class ProcessorPipeline:
```
mih, line 195: Wondering what the envisioned *public* API would be here. It seems `process()` and `finalize()` would/should not be called outside of `process_from()`. If that is true, could `process_from()` become `__call__()`?

datalad_next/runners/data_processors.py
```diff
@@ -0,0 +1,316 @@
+"""
+This module contains a generator that reads data-chunks from an iterable,
```
mih, line 2: I assume that generator is `ProcessorPipeline`?

From my POV it would make sense to cleanly split the pipelining aspect from the data processor implementations, and place them in different files.

mih:
I posted another round of comments. I am not sure if that is useful, I will stop until I got feedback.

FTR: At 28 commits it has reached my attention span limit. It would be useful for reviewing, if the PR is rebased and the commits resemble more reviewable units (new features and their tests, adoption of not features in previously existing code). ATM the commits resemble more the development history of the features themselves.

datalad_next/runners/run.py
christian-monch: I think it would very well be applicable. And I think it would make the code inside the protocol classes less "ad-hoc", more unified, and probably easier to understand.

datalad_next/runners/run.py
christian-monch: Nope, Thanks. Fixed in f63d800d3f9d7cc73d7dd7e782edc3a768461cb3

datalad_next/runners/run.py
christian-monch: Yes, it is the same, but I think the code is easier to understand this way.

datalad_next/runners/run.py
christian-monch: We do not expect too many results here. An alternative would be to do `for _ in result: pass`. We should exhaust the result generator to trigger cleanup and fetching of the process exit code.

datalad_next/iter_collections/gitworktree.py
christian-monch: > This raises the questions if anything useful/typical could/would be done with `r` (context manager yield), before it is consumed by `process_from()`?

`r` holds the exit code of the process, so it makes sense for the user to keep a copy around. But I like the idea of `processor=`.  We could copy the exit-code to the yielded  `process_from()`-generator, when the __exit__-method is executed. So the user-interface would actually not change.

datalad_next/runners/run.py
mih: If that is always the case (not many results), it would be good to have the comment say how this conclusion is made. From my POV it depends very much on the application and could potentially be huge. So I probably do not understand things properly.

datalad_next/runners/run.py
christian-monch: I am not sure how to protect against that. A user might intentionally not specify a terminate or kill time in order to let an important process, e.g. database-vacuum, do its job. On the other hand, a user might just not think about this and unintentionally let the process run forever. We would not be able to distinguish those two situations

Documentation and examples will go into the run-doc string and the "Runner Developers Handbook".

datalad_next/runners/run.py
christian-monch: You are right, it depends on the application. The idea was that, when we reach this point, we have put the process into a state in which it should exit. Depending on the application in question, there might indeed be a large amount of output, e.g. a summary of operations etc.

I will change it so that it does not store output from the subprocess at this point

datalad_next/runners/data_processors.py
christian-monch: Yes, it is unused. It could be used to process JSON output from `git annex --json ...` calls.

datalad_next/runners/data_processors.py
christian-monch: Yes, it makes sense. I wanted to prevent too many modules that we have to import from. We might still put the implementations into different files and import the processors into the name space of the pipelining code, so that they could be imported from there!?

christian-monch:
> I posted another round of comments. I am not sure if that is useful, I will stop until I got feedback.
> 
> FTR: At 28 commits it has reached my attention span limit. It would be useful for reviewing, if the PR is rebased and the commits resemble more reviewable units (new features and their tests, adoption of not features in previously existing code). ATM the commits resemble more the development history of the features themselves.

Will do that as soon as the batch code is settled

datalad_next/runners/data_processors.py
mih: From the POV of the `datalad_next` code base, module number minimization is not a goal. If some things are "standard" components, they can be propagated to the mini-package root. However, a new feature is hardly a "standard" instantaneously.

datalad_next/runners/run.py
mih: I want to add a few more comments after having tried to exercise this through in a concrete use case.

It is unclear where the timeouts apply, and thereby it is unclear how the  must be used.

Example: I have a process that downloads some information. This information in reported in a json-lines format. The process is batch-mode of some kind (an input triggers a response, continued until stdin closes).

Now there are several layers a user of this helper needs to be aware of.

- the process that downloads and eventually reports a JSON-line
- the protocol that assembles a JSON-line and decodes it, once complete

Let's assume the remote bandwidth is really low (few bytes per second), and the information that needs to go into a single JSON-line is a few kb, so it needs minutes to assemble a response, and have the "outer" protocol generator yield something.

What would the timeouts apply to? The inner layer (i.e. nothing coming down the remote connection), or the outer layer (no JSON-line-based result yielded for X seconds)?

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
+"""
+This module provides a context-manager that executes a subprocess.
+
+The context manager supports the termination and/or killing of the subprocess,
+if it is executed with a not-None `terminate_time` and/or `kill_time`.
+
+The context manager supports generator- and non-generator-protocols.
```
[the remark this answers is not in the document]
christian-monch: done

datalad_next/runners/run.py
christian-monch: Done (did not use the suggestion, because there were a few other doc changes as well)

datalad_next/runners/run.py
christian-monch: Done

datalad_next/runners/run.py
```diff
@@ -0,0 +1,166 @@
...
+
+def create_kill_wrapper(cls: type[Protocol]) -> type[Protocol]:
+    """ Extend `cls` to supports the "kill-interface"
+
+    This function creates a subclass of `cls` that contains the components
+    of the "kill-interface".  The two major components are amethod called `arm`,
+    and logic inside the timeout handler that can trigger a termination or
+    kill signal to the subprocess, if the termination time or kill time has
+    come.
+
```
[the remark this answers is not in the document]
christian-monch: It is only used internally and a mixture between a real interface, i.e. it contains the method `arm`, and a behavior overloading, i.e. wrap `timeout` of the base-class to implement the kill-logic.

Since it is not a simple interface I put it in quotes.

I left it as it is for now. I think I might rather replace "kill-interface" with "termination and kill mechanisms" or simillar

datalad_next/runners/run.py
christian-monch: The kill and termination mechanism is implemented largely in the `timeout`-callback. If the user does not specify timeouts, we introduce one second timeout to ensure that the `timeout`-callback is called. We also record the fact that we introduced the timeout in order to do the right thing at the end of our `timeout`-handler

datalad_next/runners/run.py
christian-monch: If you use a generator-protocol, then the timeouts that aid the terminating and killing mechanism are only triggered when the context is left. I extended the documentation to clarify that. So for `stdout_batchcommand` and `annexjson_batchcommand`, the timeout-counting starts when the context is left, which means when the code in the context decides that all operations are done and exits the context.

Timeout-handling for the transmission should be done in the protocol that is used, e.g.  `StdOutCaptureGeneratorProtocol`. For example, the protocol might send timeout-messages to the result generator and the in-context code can decide what to do. In another example, if the protocol raises an exception in its `timeout`-callback, the `batchcommand`-context would be left and the terminating- and killing- timeouts would be started.

datalad_next/runners/run.py
christian-monch: `armed` means that the `KillWrapper`-protocol will start counting timeouts in order to terminate the process. Currently it is set up like this:

- if the protocol is a non-generator-protocol, it (more precisely the `KillWrapper`) will be armed when the process is started, i.e. when the context is entered.
- if the protocol is a generator-protocol, it will be armed when the context is left.

That happens to coincide with `armed == not issubclass(protocol_class, GeneratorMixIn)` (which would be a nice change in the code; will do that), but that is not the only imaginable way to arm the wrapper-class. Since it is an idependent "variable", I kept it separate.

datalad_next/runners/run.py
christian-monch: Dcoumentation is also added to the doc-string of the context manager

datalad_next/runners/run.py
christian-monch: Done

datalad_next/iter_collections/gitworktree.py
christian-monch: > This comments make me think whether could be any legitimate use of run() without the decoder construct below, whenever output is to be processed?

Output processing does not necessarily include decoding. A remote ssh cat-command would want to receive bytes and still require processing, for example, to look for a specific pattern in the returned bytes.

datalad_next/runners/data_processors.py
christian-monch: `SplitLineProcessor` is gone.

datalad_next/runners/data_processors.py
christian-monch: I removed the class `SplitLinesProcessor`. The `LineSplitter` that was used by `SplitLinesProcessor` has shortcomings that make it not feasible for the pipelines, i.e. it does not support `bytes`  and, as a minor issue, it cannot be properly re-used.

datalad_next/runners/data_processors.py
christian-monch: Documentation added.

datalad_next/runners/data_processors.py
christian-monch: `process` and `finalize` could be called by protocol implementations in their `pipe_data_received` and `connection_lost`-callbacks. This could be especially useful for non-generator protocols. I think it would help us to reduce code duplication in those protocols and robustify the data handling.

But we could still make `process_from` a `__call__`.

christian-monch:
> I posted another round of comments. I am not sure if that is useful, I will stop until I got feedback.
> 
> FTR: At 28 commits it has reached my attention span limit. It would be useful for reviewing, if the PR is rebased and the commits resemble more reviewable units (new features and their tests, adoption of not features in previously existing code). ATM the commits resemble more the development history of the features themselves.

Thanks a lot for the thorough commenting. I think your comments are really helpful.

datalad_next/runners/data_processors.py
christian-monch: I will split them up in a new commit.

mih:
Thanks for the update. Before doing another review, I will rebase https://github.com/datalad/datalad-next/pull/469 on top of this PR and get some 1st-hand experience with the tooling from a user POV.

datalad_next/runners/run.py
mih: I understand the first paragraph and appreciate the clarification.

I have difficulties following the reasoning in the 2nd paragraph. When you say "should be done", you are saying "it is not implemented anywhere"? Do we have any protocol implementation that behaves like this? If not, how do we know that it would work?

datalad_next/runners/run.py
christian-monch: No idea. It was like then when I met Datalad

datalad_next/runners/run.py
```diff
@@ -0,0 +1,258 @@
...
+    cmd: list,
+    protocol_class: type[Protocol],
+    *,
+    cwd: Path | None = None,
+    stdin: int | IO | bytes | Queue[bytes | None] | None = None,
+    timeout: float | None = None,
+    closing_action: Callable | None = None,
+    terminate_time: int | None = None,
+    kill_time: int | None = None,
+    **protocol_kwargs,
```
mih, line 114: https://github.com/datalad/datalad-next/pull/484#discussion_r1367971873 made me suspicious re whether this is a good idea. I do not see how/why the keyword arguments of `run()` would be guaranteed to be non-conflicting with kwargs of an arbitrary protocol implementation.

If they are indeed not, there should be a strong incentive for claiming that namespace in this way. Any protocol would need to obey this space, and any future addition to the `run()` API would need to be conscious of any protocol implementation out there.

datalad_next/runners/run.py
```diff
@@ -0,0 +1,258 @@
...
+        The protocol class that should be used to process subprocess events.
+    cwd: Path, optional
+        If provided, defines the current work directory for the subprocess
+    stdin: int | IO | bytes | Queue[bytes | None], optional
+        Input source or data for stdin of the subprocess. See the constructor
+        of :class:`ThreadedRunner` for a detailed description
+    timeout: float, optional
+        If provided, defines the time after which the timeout-callback of the
+        protocol will be enabled. See the constructor
+        of :class:`ThreadedRunner` for a detailed description
```
mih, line 167:
```suggestion
    timeout: float, optional
        If provided, defines the time (in seconds) after which the timeout-callback of the
        protocol will be triggered. See the constructor
        of :class:`ThreadedRunner` for a detailed description
```

It seems that the critical doc is in `Protocol.timeout()`.

> Called if the timeout parameter to WitlessRunner.run() is not `None` and a process file descriptor could not be read (stdout or stderr) or not be written (stdin) within the specified time in seconds, or if waiting for a subprocess to exit takes longer than the specified time.

In general I understand the desire to avoid duplication of documentation. However, the reference `:class:ThreadedRunner` is not resolved, and a readers is tasked with a scavenger hunt of finding that class (which is not in a module of this name, in a different package, which once found points to another class and method which only then has this IMHO critical piece of information).

mih:
This needs

```diff
diff --git a/datalad_next/runners/__init__.py b/datalad_next/runners/__init__.py
index cca244f..821cd51 100644
--- a/datalad_next/runners/__init__.py
+++ b/datalad_next/runners/__init__.py
@@ -16,6 +16,8 @@ of a Git repository.
 
    GitRunner
    Runner
+   run
+   batch
 
 Additional information on the design of the subprocess execution tooling
 is available from https://docs.datalad.org/design/threaded_runner.html
```

to get any rendered documentation.

datalad_next/runners/run.py
```diff
@@ -0,0 +1,258 @@
...
+    ----------
+    cmd : list[str]
+        The command list that is passed to ``ThreadedRunner.run()``
+    protocol_class : type[Protocol]
+        The protocol class that should be used to process subprocess events.
+    cwd: Path, optional
+        If provided, defines the current work directory for the subprocess
+    stdin: int | IO | bytes | Queue[bytes | None], optional
+        Input source or data for stdin of the subprocess. See the constructor
+        of :class:`ThreadedRunner` for a detailed description
```
mih, line 163: I have the same issue with having minimal docs here in favor of an unresolved link. But at minimum it needs to say that `stdin=None` is the same as `stdin=DEVNULL` by default.

datalad_next/runners/run.py
```diff
@@ -0,0 +1,258 @@
...
+    The run-context manager supports the guaranteed exit of a subprocess through
+    either:
+
+     a) natural exit of the subprocess
+     b) termination of the subprocess via SIGTERM, if ``terminate_time`` is
+        specified
+     c) termination of the subprocess via SIGKILL, if ``kill_time`` is specified
+
+    If the process terminates the run-context manager will ensure that its exit
+    status is read, in order to prevent zombie processes.
```
mih, line 132: This would be a good place to provide information on how the exit code could be obtained when using `run()`

datalad_next/runners/run.py
```diff
@@ -0,0 +1,258 @@
...
+
+    If neither ``terminate_time`` nor ``kill_time`` are specified, and the
+    subprocess does not exit by itself, for example, because it waits for some
+    input, the ``__exit__``-method of the run-context manager will never return.
+    In other words the thread will seem to "hang" when leaving the run-context.
+    The only way to ensure that the context is eventually left is to provide a
+    ``kill_time``. It is a good idea to provide a ``terminate_time`` in
+    addition, to allow the subprocess a graceful exit (see ``kill_time``- and
+    ``terminate_time``-argument descriptions below).
```
mih, line 141: That makes me wonder: If I have a batch process (no terminate nor kill time set), and my code crashes due to a bug on the receiving end of that batch processes output, the `__exit__` would hang forever?

Wouldn't it make sense to close the process input, and have that trigger a normal exit?

mih: I now see that there is `closing_action`. Why would this behavior not be default, and require a user to remember to write this out on each use?

datalad_next/runners/run.py
christian-monch: Very good point. You are right that there is potential for conflict. Will change that.

christian-monch: Fixed in 25bf3cfc9cf2cbab0b502836b0aab0ff6f93edc7

That brought up the question of whether we should prefix keyword arguments of `datalad`-provided protocols with a `dl_` and document that we reserve the `dl_`* keyword space in protocols in order to prevent name clashes with external protocol developers.

datalad_next/iter_collections/gitworktree.py
```diff
@@ -250,23 +254,27 @@ def _lsfiles_line2props(
...
+        yield from process_from(
+            data_source=r,
+            processors = [
+                decode_processor('utf-8'),
+                splitlines_processor(separator='\0', keep_ends=False)
+            ]
+        )
```
mih, line 280:
```suggestion
    with run(
            cmd=[
                'git', 'ls-files',
                # we rely on zero-byte splitting below
                '-z',
                # otherwise take whatever is coming in
                *args,
            ],
            protocol_class=StdOutCaptureGeneratorProtocol,
            processors = [
                decode_processor('utf-8'),
                splitlines_processor(separator='\0', keep_ends=False)
            ],
            cwd=path,
    ) as r:
        yield from r
```

I brought it up before, but still think this usage should not be more complex than this. If `run()` must not have `processors=` it could be renamed and the replacement could have it.

datalad_next/runners/data_processors/decode.py
```diff
@@ -0,0 +1,86 @@
+""" This module contains data processors for the data pipeline processor
+
+The data processors contained here are:
+
+  - decode_utf8_processor
+
+"""
```
mih, line 7: None of this documentation is accessible in the renderer docs. This hides the fact that all the `data_processors` modules are documented in the exact same way, making them indistinguishable from each other.

I think the pattern to manually list the main components in a module in the top-level docstring is not optimal. In the renderer documentation this would e automatic/obvious already. Here it adds content that needs to be kept in sync. I believe an `__all__` declaration would be a more appropriate alternative.

In general I think that that sorting of code in all the modules is artificially creating the need to describe what matters most. The "public" component tends to be last in the file (although it could be first).

datalad_next/iter_collections/gitworktree.py
christian-monch: Thanks. I think I missunderstood your earlier comments.


I have a commit which defines a `StdOutCapturePipelineGeneratorProtocol`. The protocol takes an optional processor list as a keyword-argument and sends the process-pipeline results to the generator. This is transparent to the run-context-manager and achieves the same terse "ls-files"-implementation, i.e. the code in the context is just:

```
yield from r
```

I did not yet put it in because it requires selecting the "right" protocol and providing a processor list in the `protocol_kwargs`. That seemed complex to me. And, in contrast to your suggestion (put the pipeline processing into `run`) it requires support by the protocol. It would be transparent for the `run`-context manager code though.

Your suggestion should in principle support any protocol. It would require different treatment of generator- and non-generator protocols, if we want to apply pipelines to non-generator protocols as well. The latter would require the result of the non-generator protocols to adhere to a fixed convention. That seems the be the case, all non-generator protocol results are a dictionary with the keys `stdout` and `stderr`. If we want to support processor pipelines on non-generator protocols, the `run`-context manager would have to pass the values of `result_dict['stdout']` and possibly `result_dict['stderr']` through the processing pipeline. This would also require a way to specify processor lists for stdout and stderr. Alternatively, we could limit the `processors`-keyword validity to generator protocols and stdout only.

I am not sure what would be the best approach here. WDYT?

christian-monch: After having worked with adding processing pipelines to non-generator protocols, I would say: "let's not do that". The non-generator protocols are very much centered around joining individual byte-chunks in their `pipe_data_received`-method and decoding them into strings in their `_prepare_result`-method.  Without overwriting the `_prepare_result`-method, it is not even possible to receive undecoded byte-content from a subprocess.

To make use of arbitrary pipelines, `pipe_data_received` and `_prepare_result` have to be overridden. That would introduce another "subtree" of protocol-class definitions. Alternatively, we could introduce a "synchronized"-run mode, that would collect all results from a  generator-protocol-based `run` and return them.

christian-monch: I brought the processing protocol classes back. Using those classes the code that executes `git ls-files` looks like this:

```python
    with run(
            cmd=[
                'git', 'ls-files',
                # we rely on zero-byte splitting below
                '-z',
                # otherwise take whatever is coming in
                *args,
            ],
            protocol_class=StdOutCaptureProcessingGeneratorProtocol,
            stdin=DEVNULL,
            cwd=path,
            protocol_kwargs=dict(
                processors=[
                    decode_processor('utf-8'),
                    splitlines_processor(separator='\0', keep_ends=False)
                ]
            )
    ) as r:
        # This code uses the data processor chain to process data. This fixes
        # a problem with the previous version of the code, where `decode` was
        # used on every data chunk that was sent tp `pipe_data_received`. But
        # data is chunked up randomly and might be split in the middle of a
        # character encoding, leading to weird errors.
        yield from r
```

mih:
I want to take one aspect from above and bring it up for a dedicated discussion:

> I would say: "let's not support processing pipelines in non-generator protocols"

What about "let's not support non-generator protocols" instead?

This PR adds 2k+ lines already. This is ~10% of the entire code base. It is a LOT. I get the impression that some of that is just due to "need to support old stuff that happened to be done". But this is not what this extension is about.

If we can find a paradigm that can be supported with a lean implementation that does not multiply the code base, I would very much prefer that. I do not mind is that has consequences, as long as they are spelled out.

What would happen, if only generator-protocols are supported? What would be the performance impact on *real* use cases (I am specifically not talking about theory -- I understand that thread-management adds complexity).

christian-monch:
> I want to take one aspect from above and bring it up for a dedicated discussion:
> 
> > I would say: "let's not support processing pipelines in non-generator protocols"
> 
> What about "let's not support non-generator protocols" instead?

That is a **good** question. I think it is very interesting idea (I had a similar hunch when I wrote the developer handbook). From the top of my head, an probably incomplete initial answer:

> 
> This PR adds 2k+ lines already. This is ~10% of the entire code base. It is a LOT. I get the impression that some of that is just due to "need to support old stuff that happened to be done". But this is not what this extension is about.
> 
> If we can find a paradigm that can be supported with a lean implementation that does not multiply the code base, I would very much prefer that. I do not mind is that has consequences, as long as they are spelled out.
> [...]

If we only support generator-protocols, we might get away with exactly one protocol that just forwards all events to a result generator. Check the `GenericGeneratorProtocol` example in the "Runner with Generators Protocols - Developer handbook" (https://datalad--476.org.readthedocs.build/projects/next/en/476/developer_guide/generator-runner.html) for an example. That approach moves almost all processing out of protocol callbacks into the code that uses the runner.

I think it is conceptually simpler because there are no more "opaque" operations performed in protocol classes. The drawback is that it would require new abstractions to write nice, correct, and re-usable code that processes the events in the `for event in event_source:`-loop. The data processing pipeline is one concept that could be useful here. So the code base would probably not be reduced in total.

The sweet spot is probably somewhere in the middle. The `*ProcessingGeneratorProtocol`-classes might be a combination of abilities and flexibility that could: a) reduce the overall number of classes, while b) support a large range of usage scenarios.


> [...] What would happen, if only generator-protocols are supported? [...]

##### Within `datalad-next`:

1. A reduced number of protocol classes.
2. Possibly shorter protocol class names due to the removal of the `Generator`-part.
3. If we still want to support a synchronous mode, i.e. a mode that behaves as if there was a non-generator-protocol, we could use a wrapper around asynchronous runs. The wrapper would collect outputs from the generator until the process terminates and return those outputs together with the exit code to its caller. (A generic wrapper would either require that all generator-protocols signal data sources, i.e. stdout or stderr, in the same way, or that a provided function is capable of determining the data source.)


##### Within `datalad`:

1. A number of runner invocations would have to be changed, to use a wrapper function as described above.


##### Overview of protocols that are currently used in `datalad-next`:

`datalad-next` currently uses the following protocols in its application code, i.e. not in the tests:

- `NoCaptureGeneratorProtocol`
- `StdOutCaptureGeneratorProtocol`
- `StdOutCaptureProcessingGeneratorProtocol`
- `GeneratorAnnexJsonProtocol` (from datalad-core)

The tests use the following protocols in addition:

- `StdOutErrCaptureProcessingGeneratorProtocol`
- `NoCapture` (from datalad-core)
- `StdOutCapture` (from datalad-core)
- `StdErrCapture` (from datalad-core)
- `StdOutErrCapture` (from datalad-core)


> What would be the performance impact on _real_ use cases (I am specifically not talking about theory -- I understand that thread-management adds complexity).

The performance impact should be negligible because the non-generator protocols and the generator protocols use the same thread-based implementation. Compared to non-generator protocols the generator protocols add the overhead of:

1. sending results to the result generator queue
2. going through the result generator state machine, when the next result is fetched

which does not create a lot of performance overhead.


As a remark: if performance becomes a problem, we could look into a `select`- or `poll`-based runner on `posix`, which would expose the same interface as `ThreadedRunner` and be a drop-in replacement that could automatically be used in `posix`-environments.

mih:
Registering https://github.com/datalad/datalad-next/pull/497#issuecomment-1780665496 here.

I did not expect that, but I have to admit that I lost track a bit.

christian-monch:
This PR is superseded by PR #516. I will close it for now.

PR #491, opened 2023-10-19
Add two tests that document portability issues of run-records

mih:
- IO specifications are stored in platform-native conventions
- IO specifications are reported as-is

combined with the absence of any platform type record, this makes run-records non-portable across unix/windows scopes.

Ping https://github.com/datalad/datalad/issues/7512

TODO

- [ ] rethink the interpretation of old run-records. I tend to believe that we should never guess. We should add a config switch how run-records are to be intepreted. By default, all recorded paths should be considered POSIX. By config, we should be able to switch that to `windows` (not to `native`). This could/would be considered for reading (`rerun`) AND writing (`run`). If incoming paths are incompatible with the setting (e.g., absolute windows path and a POSIX setting), an error would be issued.

### DIFF
--- a/datalad_next/patches/tests/test_run.py
+++ b/datalad_next/patches/tests/test_run.py
@@ -1,4 +1,7 @@
+from pathlib import Path
 import pytest
+
+from datalad.local.rerun import get_run_info
 
 from datalad_next.exceptions import IncompleteResultsError
 from datalad_next.tests.utils import (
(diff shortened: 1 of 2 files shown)


datalad_next/patches/tests/test_run.py
```diff
@@ -23,3 +26,79 @@ def test_substitution_config_default(existing_dataset):
...
+    outfile.parent.mkdir()
+    infile.touch()
+    ds.save()
+    assert not outfile.exists()
+    # script copies any 'inputs' to the outputs dir
+    res = ds.run(
+        '{python} -c "'
+        'from shutil import copyfile;'
+        'from pathlib import Path;'
+        r'[copyfile(f, Path.cwd() / \'outputs\' / Path(f).name)'
```
adswa, line 46:
```suggestion
        r'[copyfile(f, Path.cwd() / \"outputs\" / Path(f).name)'
```

There is one current test failure because of this
```
[INFO] == Command start (output follows) ===== 
  File "<string>", line 1
    from shutil import copyfile;from pathlib import Path;[copyfile(f, Path.cwd() / \'outputs\' / Path(f).name)  for f in "inputs/testfile.txt".split()]
                                                                                                                                                      ^
SyntaxError: unexpected character after line continuation character
[INFO] == Command exit (modification check follows) ===== 
FAILED
```

mih: I actually changed it from the alternative you propose -- but that fails on windows due to an unfortunate quoting behavior

PR #492, opened 2023-10-20
Add EnsureHashAlgorithm

adswa:
This PR revives #357, which I couldn't push into directly. I didn't add much to it, I committed a patch @mih proposed a while ago already on top of it, and I added a basic set of tests for the constraint.


mih (approved):
Thanks for reviving this feature! I added a changelog, minimal documentation, and a missing test to reach 100% coverage. It tested it, and it LGTM!

PR #497, opened 2023-10-23
Revival of #469 to test #484

mih:


datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+    annexkey: str | None = None
+    annexsize: int | None = None
+
+
+def iter_annexworktree(
+    path: Path,
+    *,
+    untracked: str | None = 'all',
+    link_target: bool = False,
+    fp: bool = False,
```
christian-monch, line 65: `fp` is currently not used. It seems to be included for signature-compatibility. If that is the case, the documentation should be adjusted and we might report a `True` value as an error

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+
+    When no reporting of link targets or file-objects are requested, items of
+    type :class:`GitWorktreeItem` are yielded, otherwise
+    :class:`GitWorktreeFileSystemItem` instances. In both cases, ``gitsha`` and
+    ``gittype`` properties are provided. Either of them being ``None``
+    indicates untracked work tree content.
+
+    .. note::
+      The ``gitsha`` is not equivalent to a SHA1 hash of a file's content,
+      but is the SHA-type blob identifier as reported and used by Git.
```
christian-monch, line 86: Copy and past from `iter_gitworktree`, not applicable here.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+    PurePosixPath,
+)
+from typing import (
+    Generator,
+)
+
+from datalad_next.runners import (
+    DEVNULL,
+    ThreadedRunner,
+)
```
christian-monch, line 23: unused

```suggestion
```

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+                annexkey=ainfo.get('key'),
+                annexsize=asize,
+                annexobjpath=aobjpath,
+            )
+
+
+def _get_annexinfo(path: Path) -> dict:
+    from datalad.support.annexrepo import GeneratorAnnexJsonNoStderrProtocol
+
+    map = {
```
christian-monch, line 164: The `map` dict shadows the `map`-function. IMHO, this makes the code below confusing to read. Maybe replace with `key_mapper` or similar?

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+    ) as annexfind:
+        return {
+            # git-annex reports the file path in POSIX conventions,
+            # even on windows, but the hashdir report comes in
+            # platform conventions.
+            # we need to go to platform conventions to match the behavior of
+            # iter_gitworktree()
+            PurePath(PurePosixPath(r['file'])):
+            {
+                k: map[k](r[k]) for k in map.keys() if k in r
```
christian-monch, line 191: `map.keys()` is not required, `for k in map` would iterate over all keys. Also, instead of mapping a key to a value, i.e. `map[k]`, one could use `items()`. e.g.:

```suggestion
                k: m(r[k]) for k, m in map.items() if k in r
```

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,194 @@
...
+    map = {
+        'bytesize': lambda x: x,
+        'key': lambda x: x,
+    }
+
+    with run(
+        [
+            'git', 'annex', 'find',
+            # we want everything regardless of local availability.
+            # the faster `--anything` came only in 10.20230126
```
christian-monch, line 173: I am not sure how much of a performance difference it makes, but since we know the annex version we could choose the faster command on applicable annex versions.

christian-monch (commented):
Not sure about the maturity yet, so: just my two cents.

mih:
Thanks for the review @christian-monch but there is no need for it right now. This is draft mode, because I am not sure if this is the way to go at all. I am just using it to explore #484 from a user perspective.

mih:
Due to the significant changes in #484 I rebased this trial and tried to get an sense of the performance changes. However, it turns out that the code no longer yields any result and merely hangs.

See https://github.com/datalad/datalad-next/pull/469/commits/dafafb31002f1069aa8192c5f315752624f80082 for the test cases.

mih:
We can close this in favor of #520

PR #516, opened 2023-10-26
Minimal run-context-mananger

christian-monch:
This PR is a minimized version of PR #484, that concentrates on the run-context-manager and the batch-context-manager and their application only.

I will close #484 for now in favor of this PR

TODO:

- [x] figure out https://github.com/datalad/datalad-next/issues/529#issuecomment-1784757155


mih:
Thanks! This seems to be a sensible step!

Can we identify a concrete objective for this PR, so it can be reviewed adequately?

Would it be fair to say:

> This aims to make any *direct* use of any `Runner` class obsolete, in favor of a context-manager based pattern for (concurrent) subprocess execution.

Would that fit?

mih (changes requested):
I went through the changes with the mindset of the objective being to take away needless flexibility of subprocess execution and to establish a context-manager-based pattern.

`ThreadedRunner` is not only a single private import: nice!

I think that `runners/batch.py` and `runners/tests/test_batch.py` should not be part of this PR. This is self-contained and tested code, but it is not used. Even if we have plans for adoption (later), this just adds weight to this PR and makes it less clear what this change is about IMHO.

christian-monch:
> Thanks! This seems to be a sensible step!
> 
> Can we identify a concrete objective for this PR, so it can be reviewed adequately?
> 
> Would it be fair to say:
> 
> > This aims to make any _direct_ use of any `Runner` class obsolete, in favor of a context-manager based pattern for (concurrent) subprocess execution.
> 
> Would that fit?

Yes

> Edit: I will review based on the assumption that the answer is "yes"

👍🏻 
 
> FTR: It would be good to have PRs from forks to avoid running all CI twice.

Yeah, sorry. I messed up the upstream location on this one. Will take better care next time

christian-monch:
> I went through the changes with the mindset of the objective being to take away needless flexibility of subprocess execution and to establish a context-manager-based pattern.
> 
> `ThreadedRunner` is not only a single private import: nice!
> 
> I think that `runners/batch.py` and `runners/tests/test_batch.py` should not be part of this PR. This is self-contained and tested code, but it is not used. Even if we have plans for adoption (later), this just adds weight to this PR and makes it less clear what this change is about IMHO.

I see your point. Will take it out into its own PR.

christian-monch:
> [...]
> I think that `runners/batch.py` and `runners/tests/test_batch.py` should not be part of this PR. This is self-contained and tested code, but it is not used. Even if we have plans for adoption (later), this just adds weight to this PR and makes it less clear what this change is about IMHO.

Done

christian-monch:
Marking this as "Ready for review". Remaing appveyor-errors are related to:

- `test_create_sibling_ghlike.py::test_gogs`
- `test_create_sibling_gogs.py::test_gogs`
-

mih:

PR #520, opened 2023-10-27
POC: implement `iter_annexworktree` with one run-context manager and two batch-context managers

christian-monch:
This PR is not meant for merging. It is a POC to demonstrate how `iter_annexworktree` can be implemented with run-context-managers and batch-context-managers in the following fashion:

```python
with \
        run(['git', 'ls-files', ...], ...) as git_ls_files, \
        annexjson_batchcommand(['git', 'annex', 'examinekey', ...], ...) as examine_key, \
        annexline_batchcommand(['git', 'annex', 'lookupkey', ...], ...) as lookup_key:
    for item in git_ls_files:
        annex_key = lookup_key(item.name + '\n')
        if annex_key:
            key_properties = examine_key(annex_key)
            yield AnnexWorkTreeItem(item, key_properties)
        else:
            yield AnnexWorkTreeItem(item, None)
```

It also introduces a wrapper around the result iterators that will trigger a termination if an iteration step is longer then a given timeout.

Note: the provided git-ls-files implementation is not as complete as the "original", since this is a POC. It might not work on some corner cases. So for large scale performance comparisons at little more work is most likely to be done.

### DIFF
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -163,7 +163,7 @@ def iter_annexworktree(
 
         # Remaining git ls-files results are all unannexed, yield them.
         assert len(gaf_store) == 0
-        for path, glf_item in glf_store.items():
+        for glf_item in glf_store.values():
             yield AnnexWorktreeItem(
                 name=glf_item.name,
                 gitsha=glf_item.gitsha,
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -1,0 +1,174 @@
+"""Report on the content of a Git-annex repository worktree
+
+The main functionality is provided by the :func:`iter_annexworktree()`
+function.
+"""
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+from itertools import zip_longest
+from pathlib import (
+    Path,
+    PurePath,
+    PurePosixPath,
+)
+from typing import Generator
+
+from datalad.support.annexrepo import GeneratorAnnexJsonNoStderrProtocol
+from datalad_next.runners.batch import annexjson_batchcommand
+from datalad_next.runners.run import run
+
+from .gitworktree import (
+    GitWorktreeItem,
+    GitWorktreeFileSystemItem,
+    iter_gitworktree,
+)
+
+lgr = logging.getLogger('datalad.ext.next.iter_collections.annexworktree')
+
+
+# TODO Could be `StrEnum`, came with PY3.11
+class AnnexTreeItemType(Enum):
+    """Enumeration of item types of Git trees
+    """
+    file = 'file'
+    executablefile = 'executablefile'
+    symlink = 'symlink'
+    directory = 'directory'
+    submodule = 'submodule'
+
+
+@dataclass
+class AnnexWorktreeItem(GitWorktreeItem):
+    annexkey: str | None = None
+    annexsize: int | None = None
+    annexobjpath: PurePath | None = None
+
+
+@dataclass
+class AnnexWorktreeFileSystemItem(GitWorktreeFileSystemItem):
+    annexkey: str | None = None
+    annexsize: int | None = None
+
+
+def iter_annexworktree(
+    path: Path,
+    *,
+    untracked: str | None = 'all',
+    link_target: bool = False,
+    fp: bool = False,
+) -> Generator[AnnexWorktreeItem | AnnexWorktreeFileSystemItem, None, None]:
+    """ Iterate over the git annex worktree of a Git repository
+
+    This iterator uses ``git ls-files``, ``git annex lookupkey``, and
+    ``git annex examinekey`` to report on a git annex work tree of a Git repo.
+
+    .. note::
+      This is a POC implementation to demonstrate the concurrent use of
+      run-context-managers and two batch-context-managers. Most parameters are
+      ignored, the protocol that is used does not take care of pending results.
+
+    Parameters
+    ----------
+    path: Path
+      Path of a directory in a Git repository to report on. This directory
+      need not be the root directory of the repository, but must be part of
+      the repository's work tree.
+    untracked: {'all', 'whole-dir', 'no-empty'} or None, optional
+      If not ``None``, also reports on untracked work tree content.
+      ``all`` reports on any untracked file; ``whole-dir`` yields a single
+      report for a directory that is entirely untracked, and not individual
+      untracked files in it; ``no-empty-dir`` skips any reports on
+      untracked empty directories. Any untracked content is yielded as
+      a ``PurePosixPath``.
+    link_target: bool, optional
+      This flag behaves analog to that of ``iter_gitworktree()``. In addition,
+      enabling link target reporting will provide the location
+      of an annex object (annexed file content) via the
+      ``AnnexWorktreeItem.annexobjpath`` property. Annex object path reporting
+      is supported whether or not a particular key is locally present.
+    fp: bool, optional
+      If ``True``, each file-type item includes a file-like object
+      to access the file's content. This file handle will be closed
+      automatically when the next item is yielded.
+
+    Yields
+    ------
+    :class:`AnnexWorktreeItem` or `AnnexWorktreeFileSystemItem`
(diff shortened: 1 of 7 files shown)


datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,167 @@
...
+    _mode_type_map,
+)
+
+lgr = logging.getLogger('datalad.ext.next.iter_collections.annexworktree')
+
+
+# The following protocol is part of the POC implementation. It allows a simple
+# `with run()`-commandline that returns `GitWorkTreeItem`-objects or
+# `GitWorkTreeFileSystemItem`-objects. In the real implementation we would
+# probably use `iter_gitworktree()`.
```
mih, line 38: This raises interesting points for me:

- yes, we would definitely use `iter_gitworktree()`, but `iter_gitworktree()` could be implemented this way for max alignment
- `GitLsFilesProtocol` should probably not sit on `StdOutCaptureGeneratorProtocol` directly, but on an intermediate protocol that implements the zero-byte-delimited-items output, or even delimited-items output. Right now the protocol assume that `ls-files` is called correctly but in the line
  ```
    run(['git', 'ls-files', '-z', '--stage', '--cached'], protocol_class=GitLsFilesProtocol...
  ```

  I cannot immediately see this requirement/implication.

- it seems we should/could aim for assembling a library of such common protocols

mih:
I left a first comment. I also tried to run it on my standard lots-of-files dataset. But it seems to hang there:

```py
from datalad_next.iter_collections.annexworktree import iter_annexworktree
res = list(iter_annexworktree('.', untracked='all', link_target=False))
<Ctrl-C> after some time

File ~/env/datalad-dev/lib/python3.11/site-packages/datalad/runner/nonasyncrunner.py:637, in ThreadedRunner.process_queue(self)
    627 while True:
    628     # We do not need a user provided timeout here. If
    629     # self.timeout is None, no timeouts are reported anyway.
   (...)
    634     # `ThreadedRunner.process_check_interval`, to check whether the
    635     # process is still running.
    636     try:
--> 637         file_number, state, data = self.output_queue.get(
    638             timeout=ThreadedRunner.timeout_resolution)
    639         break
    640     except Empty:

File /usr/lib/python3.11/queue.py:180, in Queue.get(self, block, timeout)
    178         if remaining <= 0.0:
    179             raise Empty
--> 180         self.not_empty.wait(remaining)
    181 item = self._get()
    182 self.not_full.notify()

File /usr/lib/python3.11/threading.py:324, in Condition.wait(self, timeout)
    322 else:
    323     if timeout > 0:
--> 324         gotit = waiter.acquire(True, timeout)
    325     else:
    326         gotit = waiter.acquire(False)

KeyboardInterrupt: 

```

datalad_next/iter_collections/annexworktree.py
christian-monch: > This raises interesting points for me:
> 
> * yes, we would definitely use `iter_gitworktree()`, but `iter_gitworktree()` could be implemented this way for max alignment

That's what I thought as well. I was tempted to do that. But I refrained, in order to focus on the annex-batch-related parts.

> * `GitLsFilesProtocol` should probably not sit on `StdOutCaptureProcessingGeneratorProtocol` directly, but on an intermediate protocol that implements the zero-byte-delimited-items output, or even delimited-items output. Right now the protocol assume that `ls-files` is called correctly but in the line
>   ```
>     run(['git', 'ls-files', '-z', '--stage', '--cached'], protocol_class=GitLsFilesProtocol...
>   ```
>     
>   I cannot immediately see this requirement/implication.
> * it seems we should/could aim for assembling a library of such common protocols

The, now-closed, big-runner PR with data_processors has a protocol, i.e. `StdOutErrCaptureProcessingGeneratorProtocol`, that passes all data through a data processor pipeline. The pipeline is configured by the protocol keyword argument `processors`.

With this concept, the `iter_gitworktree` call would look something like this:

```python
with run(
    ['git', 'ls-files', '-z', '--stage', '--cached'],
    protocol_class=StdOutCaptureProcessingGeneratorProtocol
    protocol_kwargs={'processors': [decode_processor('utf-8'), splitlines_processor('\x00', False), gititem_processor]}
) as git_ls_files:
    pass
```

That could of course be encapsulated into a specialized function, reducing the call to something like this

```python
with git_ls_files(path=path, ...) as gls:
    ...
```

christian-monch:
> I left a first comment. I also tried to run it on my standard lots-of-files dataset. But it seems to hang there:
> 
> ```python
> from datalad_next.iter_collections.annexworktree import iter_annexworktree
> res = list(iter_annexworktree('.', untracked='all', link_target=False))
> <Ctrl-C> after some time
> 
> File ~/env/datalad-dev/lib/python3.11/site-packages/datalad/runner/nonasyncrunner.py:637, in ThreadedRunner.process_queue(self)
>     627 while True:
>     628     # We do not need a user provided timeout here. If
>     629     # self.timeout is None, no timeouts are reported anyway.
>    (...)
>     634     # `ThreadedRunner.process_check_interval`, to check whether the
>     635     # process is still running.
>     636     try:
> --> 637         file_number, state, data = self.output_queue.get(
>     638             timeout=ThreadedRunner.timeout_resolution)
>     639         break
>     640     except Empty:
> 
> File /usr/lib/python3.11/queue.py:180, in Queue.get(self, block, timeout)
>     178         if remaining <= 0.0:
>     179             raise Empty
> --> 180         self.not_empty.wait(remaining)
>     181 item = self._get()
>     182 self.not_full.notify()
> 
> File /usr/lib/python3.11/threading.py:324, in Condition.wait(self, timeout)
>     322 else:
>     323     if timeout > 0:
> --> 324         gotit = waiter.acquire(True, timeout)
>     325     else:
>     326         gotit = waiter.acquire(False)
> 
> KeyboardInterrupt: 
> ```

Thanks for the test. Weird, it seems to not be able to get the lock for the queue... 

I will take a look at that. Not sure what triggers it, but i might be a thread that crashed while holding a lock.

christian-monch:
I used the iterator on a dataset with 33.000 entries and it took about 3:10 minutes on my development machine.

If you interrupt the execution with <CTRL-C> you are very likely to end up in the queue-lock acquire code.

mih:
> I used the iterator on a dataset with 33.000 entries and it took about 3:10 minutes on my development machine.
> 
> If you interrupt the execution with you are very likely to end up in the queue-lock acquire code, so that has no specific significance.

I did not wait for 3min, indeed. As you can see in https://github.com/datalad/datalad-next/pull/469/commits/dafafb31002f1069aa8192c5f315752624f80082 I would expect a runtime of substantially below 10s.

christian-monch:
> > I used the iterator on a dataset with 33.000 entries and it took about 3:10 minutes on my development machine.
> > If you interrupt the execution with you are very likely to end up in the queue-lock acquire code, so that has no specific significance.
> 
> I did not wait for 3min, indeed. As you can see in [dafafb3](https://github.com/datalad/datalad-next/commit/dafafb31002f1069aa8192c5f315752624f80082) I would expect a runtime of substantially below 10s.

It takes between 2 and 3 seconds to do the git-ls-files iteration only (comment out batch processing).

I did the following test:
```
> git ls-files >/tmp/glf-out
> time git annex lookupkey --batch < /tmp/glf-out >/tmp/alk-out

real    2m45,258s
user    1m16,367s
sys     1m22,510s
```

And this one:
```
> time git annex examinekey --batch --json < /tmp/alk-out >/tmp/aek-out

real    0m2,382s
user    0m2,111s
sys     0m0,261s
```

So, looking up a large number of keys is too slow. I will create a version that uses a run-context for `git ls-files`, a run-context for `git annex find`, and a batchcommend-context for `git annex examinekey`.

christian-monch:
I pushed a version that uses two run-contexts to read from `git ls-files` and `git annex find` at the "same" time. It only uses a batchcommand-context for `git annex examinekey`. `git annex lookupkey` is gone.

This version tries to keep the maximum number of stored results low while having a single pass over both run-context iterators (the actual number is dependent on the ordering of names that are returned from the two run-contexts).

The overall time is around 14 seconds. It is more than 10 seconds, but a lot less than 3:15 minutes.

The code is not as simple anymore though :-/

mih:

PR #532, opened 2023-11-06
Add batch command-context manager on top of run-context manager

christian-monch:
This PR is a Draft-PR because it is against another PR. 

This PR adds a `batchcommand`-context manager. A `batchcommand`-context manager starts a subprocess and supports batch-style communication with it. That means, the user can send a request to the subprocess and receive the response from the subprocess.

How bytes that are read from the subprocess are assembled into a response is determined by the protocol that is used when the `batchcommand`-context manager is created. Typical responses would be single lines of text or JSON-encoded data.

The PR contains three convenience functions that provide context-manager for common interaction patterns (common in `datalad` and `datalad_next` anyways):

1. `stdout_batchcommand`: this context-manager writes to `stdin` of the subprocess and returns the next, yet unread, chunk of bytes that was written to `stdout` of the subprocess.
2. `stdoutline_batchcommand`: this context-manager writes to `stdin` of the subprocess and returns the next, yet unread, line that was written to `stdout` of the subprocess. Line-separators and whether they are included in the output can be chosen when creating the context-manager.
3. `annexjson_batchcommand`: this context-mananger writes to `stdin` of the subprocess and returns a JSON-object, that is created from the next, yet unread, line that is written to `stdout` of the subprocess. This context-mananger is geared towards git-annex batch commands with JSON-result output, i.e. git-annex invocations with the parameters `--batch`, `--json`, or `--json-error-messages`.
    
The tests contain an example for a protocol, `PythonProtocol` that handles output from an interactive python interpreter, which varies in size.

### DIFF
--- a/datalad_next/runners/batch.py
+++ b/datalad_next/runners/batch.py
@@ -1,0 +1,276 @@
+"""Helpers to execute batch commands
+
+Some of the functionality provided by this module depends on specific
+"generator" flavors of runner protocols, and additional dedicated
+low-level tooling:
+
+.. currentmodule:: datalad_next.runners.batch
+.. autosummary::
+   :toctree: generated
+
+    StdOutCaptureGeneratorProtocol
+    StdOutLineCaptureGeneratorProtocol
+    GeneratorAnnexJsonProtocol
+    _ResultGenerator
+"""
+from __future__ import annotations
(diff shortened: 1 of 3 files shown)


mih:

PR #538, opened 2023-11-15
Include copy of https://github.com/uktrade/iterable-subprocess

### LINKED ISSUE #536: Performance issues in batch-commands
### The performance-issue of `datalad_next.runners.batch.batchcommand`.

This is the current result of my investigations into the performance problems that became apparent when implementing iter_annexworktree with a batch-process that reads keys from stdin and writes JSON-objects with key information to stdout.

There are a number of possible reasons for the low performance:

- Overhead through the use of protocols
- Synchronization requirements in stdin-Queues due to the use of threads
- Generally slow implementation
- ...

In the following section different potential causes for the problem are identified and explored

### Stdin Queue synchronisation

In `iter_annexworktree` we use the `batchcommand`-context manager, more precisely the `annexjson_batchcommand`-context manager to communicate with the batch-process. The runtime of `iter_annexworktree` is longer than expected. I.e. in Linux on a dataset with 33k annexed files, it takes about 14 seconds. Of those 14 seconds, about 10 are spent in the batch-process, i.e. calling `BatchProcess.__call__` and retrieving the result.

Isolating the key-lookup and performing a key-by-key lookup similar to this:
```
from datalad_next.runners.batch import annexjson_batchcommand

# Read 33163 keys from a file
with open('/tmp/kkk', 'rb') as f:
    keys = f.readlines()

with annexjson_batchcommand(['git', 'annex', 'examinekey', '--json', '--batch']) as examine_key:
    for key in keys:
        _ = examine_key(key)
```
It takes about 7.8 seconds. A profiler reveals that 57% of the time are spent when acquiring the queue-lock in `Queue.get` that is called in `ThreadedRunner.process_loop`. 

** Experiment 1:** use a non-thread based runner, e.g. a runner based on `select`.
 
I implemented a select-based runner that is API-equivalent to the ThreadedRunner, i.e. supports protocols.

The select-based runner uses about 5.3 seconds for a key-by-key lookup (in contrast to 7.8 seconds that were used by the `ThreadedRunner`).

#### Comparison with shell-based lookup

If the batch process is driven by the shell via:
```
> cat key-list | git annex examinekey --json --batch > /tmp/out
```
It takes only about 2.5 seconds to complete, instead of 7.8 seconds that `ThreadedRunner` used as described above.
But the shell-command does not reflect how `iter_annexworktree` uses the batch command, because it does not wait for a reply from the batch-process. If we "simulate" this behavior by "sending" all keys to the batch command at once to the `annexjson_batchcommand`, we get an execution time between 3.3 and 4.4 seconds in the `annexjson_batchcommand`-context, using the following source:
```
...
with annexjson_batchcommand(['git', 'annex', 'examinekey', '--json', '--batch']) as examine_key:
    examine_key._stdin_queue.put(b''.join(keys))
    for _ in enumerate(examine_key._rgen):
        pass
```

The select-based runner uses about 3.1 seconds for a bulk-lookup. This is slightly faster than the  `ThreadedRunner`-based lookup. Large differences are not expected because with bulk-lookup the `ThreadedRunner` uses `Queue.put` only once, and therefore locks the stdin-queue only once.

#### `queue.Queue` synchronisation speed.

Why is so much time spent in `Queue.get`?

### Protocol overhead

Another reason for unsatisfactory performance might be the time spent in the protocol machines. 

problem and possible solutions (expect some edits in the next hours).

The good news is that the datalad-core runner (ThreadedRunner) generally performs well under most circumstances

- [x] copy the non-batch results here

The bad news is that it does not perform that well, if input is enqueued into the stdin-Queue at certain rates. On the branch `iter-annexworktree-run-and-batch`, the following test shows unexpectedly slow performance:

```
for _ in iter_annexworktree(Path('.'), untracked='all', link_target=False):
    pass
```

Takes about 12 seconds on a dataset with 33k annexed files, i.e. `studyforrest-data-visualrois` (a shell-based implementation takes a few seconds).

In order to get to the bottom of this, I implemented a few more runner classes:

- A simple runner that yields results directly and does not implement a `_ResultGenerator`, and does not use a protocol
- A simple runner as outlined above that uses `select.select` instead of threads for I/O multiplexing and can therefore work with `collection.deque`  instead of `queue.Queue`.
- A replacement for `ThreadedRunner` that uses `select.select` instead of threads for I/O multiplexing  and supports the API of the current `ThreadedRunner`, i.e. supports protocols.

It seems that the problem is related to the use of `queue.Queue` to send stdin-data to the subprocess. If data is put into the queue at a certain rate, the runtime increases by 3 or 4 times.

The following two tests demonstrate the different behavior. Both tests send lines to an s"echo server" that returns `read: <line content>\n`. The test `test_batch_performance_intertwined` sends a single element to the stdin-Queue of the runner and reads the result. The seconds test, i.e. `test_batch_performance_bulk` sends all requests to the stdin-Queue of the runner and reads all results afterwards.

The first test takes about `11.4` seconds, the second test about `2.4` seconds, which is almost 5 times faster (both use the ThreadedRunner).

```
def test_batch_performance_intertwined():
    prog = line_echo_batch_prog.format(end="\\n")
    with stdoutline_batchcommand(
        cmd=[sys.executable, "-u", "-c", prog]
    ) as batch_process:
        for i in range(200000):
            r = batch_process(f"{i}\n".encode())
            assert r == f"read: {i}"
    assert batch_process.return_code == 0


def test_batch_performance_bulk():
    prog = line_echo_batch_prog.format(end="\\n")
    with stdoutline_batchcommand(
            cmd=[sys.executable, "-u", "-c", prog]
    ) as batch_process:
        for i in range(200000):
            batch_process._stdin_queue.put(f"{i}\n".encode())
        for i in range(200000):
            r = next(batch_process._rgen)
            assert r == f"read: {i}"
    assert batch_process.return_code == 0
```

To validate the findings, I implemented a version of `iter_annexworktree` that uses "bulk" batch processing. That means, all keys to examine are sent in a tight loop to the `annex examinekey`-batch process before any output is read from the batch-process (which obviously is not in the spirit of the generator-approach). The runtime is 6.7 seconds, which is almost twice as fast as the "intertwined" approach

I have used the select-based simple-runner to feed the command `git annex examinekey --batch --json --json-error-messages` with the 33k from `studyforrest-data-visualrois`. It takes about:

- 2.9 seconds for "intertwined", i.e. typical batch-style, processing, and
- 3.0 seconds for "bulk" processing.

which is as fast as one could expect.

The third implemented runner, i.e. a select-based runner that supports protocols, executes the intertwined `iter_annexworktree` on 33k files in 9.5 seconds, which is about 3 seconds faster than `ThreadedRunner`.

#### Possible Solutions:

- Figure out the Queue-issue root and fix it
- Use a different programming approach, i.e. more bulk processing. (That somehow defeats the generator-style)
- Use a select-based runner that allows to use `deque` because we do not have to wait at a `Queue` in this runner.
- ....

### LINKED ISSUE #537: Alternative paradigm for concurrent batch processes
I searched for alternatives to our current `ThreadedRunner`-based implementation for batch process handling (with context managers).

Related:

- https://github.com/datalad/datalad-next/pull/532
- https://github.com/datalad/datalad-next/pull/520
- https://github.com/datalad/datalad-next/pull/516

and some more prior attempts by myself. All solutions so far appear to be rather heavy, and do not deliver the necessary performance.

Below is a demo of a leaner solution:

It runs the equivalent of the following pipeline:

```bash
git ls-files | git annex find --anything --format="\${key}\n" --batch | grep -v '^[[:space:]]*$' | git annex examinekey --format="\${objectpath}\n" --batch
```

As shown below, it connects three subprocesses via a "pipe". A custom iterator takes the conceptual place of our "protocols" and performs some output processing (pipe-inline).

```py
from iterable_subprocess import iterable_subprocess

# small helper to ensure that only complete records are passed on
# here a record is a full line
# XXX this implementation is not fully correct, few lines seem to be missing
def glue(iter):
    count = 0
    trailer = None
    for i in iter:
        lines = i.splitlines()
        if trailer is not None:
            count += 1
            yield trailer + lines[0]
            trailer = None
            lines = lines[1:]
        if not lines[-1].endswith(b'\n'):
            trailer = lines[-1]
            lines = lines[:-1]
        for line in lines:
            if not line:
                continue
            count += 1
            yield line + b'\n'
    if trailer is not None:
        count += 1
        yield trailer


result_recs = []
with \
        iterable_subprocess(
            ['git', 'ls-files'], tuple()) as glsf, \
        iterable_subprocess(
            ['git', 'annex', 'find', '--anything', '--format=\${key}\n', '--batch'], glsf) as gaf, \
        iterable_subprocess(
            ['git', 'annex', 'examinekey', '--format=\${objectpath}\n', '--batch'], glue(gaf)) as gek:
    for r in glue(gek):
        print(r)
```

Benchmark:

```
❯ multitime -n5 python ~/hacking/datalad/next/iter_subproc.py > /dev/null
===> multitime results
1: python /home/mih/hacking/datalad/next/iter_subproc.py
real        3.208       0.073       3.099       3.239       3.302       
user        2.806       0.037       2.754       2.805       2.864       
sys         2.113       0.036       2.084       2.093       2.181       

❯ multitime -n 5 git ls-files | git annex find --anything --format="\${key}\n" --batch | grep -v '^[[:space:]]*$' | git annex examinekey --format="\${objectpath}\n" --batch > /dev/null
===> multitime results
1: git ls-files
real        3.276       0.040       3.234       3.267       3.327       
user        0.009       0.003       0.005       0.008       0.013       
sys         0.004       0.003       0.000       0.004       0.009       
```

So on my machine and test dataset with 36k files, I achieve the same performance as with the shell command.

I like this solution, because it has minimal (code) overhead. The underlying implementation of `iterable_subprocess` is ~100 lines of code on top of the stdlib.

While this looks good and solves an urgently needed use case, it is unclear to me what this code cannot do, that `ThreadedRunner` does for our existing implementations.

@christian-monch please have a look!

mih:
This brings the preconditions for #539, which in itself has the evidence for the utility of this approach.

Historic context from here.

ATM` it is still unclear whether it will be adopted. This changeset is primarily a preparation for running the tests in all contexts that we need.

The sources where included via

```
git subtree add --prefix datalad_next/iterable_subprocess https://github.com/uktrade/iterable-subprocess.git main --squash
```

Closes #536
Closes #537

TODO:

- [x] some tests use external commands, generally not to be expected on windows: `ls`, `cat`, `funzip`. The first two should be replacable with `dir` and `type`, but the latter (from unzip) might need to become a conditional for running the test. It is installed on appveyor, though.
      SOLVED: All tools, also `funzip` come with Git (also on Windows). It may need a PATH adjustment on a particular machine, but not additional install.
- [x] some tests fail on windows. These all seem to be related to improper exit handling on windows. STDIN is closed unconditionally, which is not appreciated by Windows (as the underlying process is no longer running). This leads to `OSError Errno 22`. The conditions under which this happens are not completely clear to me right now. Simply catching `OSError` seems not enough.
  - test_if_process_closes_standard_input_but_exits_with_non_zero_error_code_then_broken_pipe_error FAILED
  - test_if_process_closes_standard_input_but_exits_with_non_zero_error_code_then_iterable_subprocess_error FAILED
  - test_program_that_outputs_for_a_long_time_is_interrupted_on_context_exit FAILED

Test log

```
FAILED ..\iterable_subprocess\test_iterable_subprocess.py::test_if_process_closes_standard_input_but_exits_with_non_zero_error_code_then_broken_pipe_error - OSError: [Errno 22] Invalid argument
FAILED ..\iterable_subprocess\test_iterable_subprocess.py::test_if_process_closes_standard_input_but_exits_with_non_zero_error_code_then_iterable_subprocess_error - OSError: [Errno 22] Invalid argument
FAILED ..\iterable_subprocess\test_iterable_subprocess.py::test_program_that_outputs_for_a_long_time_is_interrupted_on_context_exit - assert b'BrokenPipeError' in b'Traceback (most recent call last):\r\n  File "<string>", line 3, in <module>\r\nOSError: [Errno 22] Invalid argument\r\nException ignored in: <_io.TextIOWrapper name=\'<stdout>\' mode=\'w\' encoding=\'cp1252\'>\r\nOSError: [Errno 22] Invalid argument\r\n'
 +  where b'Traceback (most recent call last):\r\n  File "<string>", line 3, in <module>\r\nOSError: [Errno 22] Invalid argument\r\nException ignored in: <_io.TextIOWrapper name=\'<stdout>\' mode=\'w\' encoding=\'cp1252\'>\r\nOSError: [Errno 22] Invalid argument\r\n' = IterableSubprocessError(120, b'Traceback (most recent call last):\r\n  File "<string>", line 3, in <module>\r\nOSError: [Errno 22] Invalid argument\r\nException ignored in: <_io.TextIOWrapper name=\'<stdout>\' mode=\'w\' encoding=\'cp1252\'>\r\nOSError: [Errno 22] Invalid argument\r\n').stderr
 +    where IterableSubprocessError(120, b'Traceback (most recent call last):\r\n  File "<string>", line 3, in <module>\r\nOSError: [Errno 22] Invalid argument\r\nException ignored in: <_io.TextIOWrapper name=\'<stdout>\' mode=\'w\' encoding=\'cp1252\'>\r\nOSError: [Errno 22] Invalid argument\r\n') = <ExceptionInfo IterableSubprocessError(120, b'Traceback (most recent call last):\r\n  File "<string>", line 3, in <module>\r\nOSError... in: <_io.TextIOWrapper name=\'<stdout>\' mode=\'w\' encoding=\'cp1252\'>\r\nOSError: [Errno 22] Invalid argument\r\n') tblen=3>.value
```


mih:
This should fix the windows behavior
```diff
diff --git a/datalad_next/iterable_subprocess/iterable_subprocess.py b/datalad_next/iterable_subprocess/iterable_subprocess.py
index 2b61f0e..12abe95 100644
--- a/datalad_next/iterable_subprocess/iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/iterable_subprocess.py
@@ -75,11 +75,24 @@ def iterable_subprocess(program, input_chunks, chunk_size=65536):
                     stdin.write(chunk)
                 except BrokenPipeError:
                     raise _BrokenPipeError()
+                except OSError as e:
+                    if e.errno != 22:
+                        # Errno22 indicates an IO failure with a
+                        # file descriptor (maybe process is dead already)
+                        raise _BrokenPipeError()
+                    else:
+                        raise
         finally:
             try:
                 stdin.close()
             except BrokenPipeError:
                 raise _BrokenPipeError()
+            except OSError as e:
+                # silently ignore Errno22, which happens on
+                # windows when trying to interacted with file descriptors
+                # associated with a process that existed already
+                if e.errno != 22:
+                    raise

     def output_from(stdout):
         while True:
diff --git a/datalad_next/iterable_subprocess/test_iterable_subprocess.py b/datalad_next/iterable_subprocess/test_iterable_subprocess.py
index d27587d..f900377 100644
--- a/datalad_next/iterable_subprocess/test_iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/test_iterable_subprocess.py
@@ -213,7 +213,7 @@ def test_program_that_outputs_for_a_long_time_is_interrupted_on_context_exit():
     end = time.monotonic()

     assert excinfo.value.returncode != 0
-    assert b'BrokenPipeError' in excinfo.value.stderr
+    assert b'BrokenPipeError' in excinfo.value.stderr or b'Errno 22' in excinfo.value.stderr
     assert end - start < 10
```

mih:
This is ready. I will now port previous attempts to this new implementation -- in separate PRs.

mih:
#539 has now shown that this code is both fast and usable.

I will now add a thin wrapper around it, such that we can extend without changing its nature too much (possibly ease future merges).

mih:
OK, let's do this!

PR #539, opened 2023-11-21
`iterable_subprocess` based `annexworktree()`

mih:
This is moving PR https://github.com/mih/datalad-next/pull/3 here, to get it tested properly. Below is the original description by @christian-monch 

It includes (sits on top of) https://github.com/datalad/datalad-next/pull/538 (merged now)

TODO

- [x] Rebase on #552
- [x] Add documentation
- [x] Make `keep_ends=True str.strip()` combo obsolete (and ease debugging why @mih cannot do it, ideally)
- [x] `fp=True` currently only does meaningful things for annexed files, but it should also act properly for any other file. This aspect of the functionality is still undertested.

----

This PR adds an implementation of `iter_annexworktree` that is based on `iterable_subprocesses` and the ideas laid out in issue https://github.com/datalad/datalad-next/issues/537.

This PR included a collection of data processors that are basically generator-wrapper.

The PR also modifies `iter_gitworktree` to use `iterable_subprocesses` instead of the datalad-core runner.

The current implementation of `iter_annexworktree` iterates over a dataset with 33k annex files in less than 5 seconds on my machine.

### DIFF
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -67,11 +67,12 @@ class AnnexWorktreeFileSystemItem(GitWorktreeFileSystemItem):
 # iter_gitworktree() we should provide the same filtering for that one
 # too!
 def iter_annexworktree(
-        path: Path,
-        *,
-        untracked: str | None = 'all',
-        link_target: bool = False,
-        fp: bool = False,
+    path: Path,
+    *,
+    untracked: str | None = 'all',
+    link_target: bool = False,
+    fp: bool = False,
+    recursive: str = 'repository',
 ) -> Generator[AnnexWorktreeItem | AnnexWorktreeFileSystemItem, None, None]:
     """Companion to ``iter_gitworktree()`` for git-annex repositories
 
@@ -124,6 +125,10 @@ def iter_annexworktree(
       Moreover, each file-type item includes a file-like object
       to access the file's content. This file handle will be closed
       automatically when the next item is yielded.
+    recursive: {'repository', 'no'}, optional
+      Pass on to
+      :func:`~datalad_next.iter_collections.gitworktree.iter_gitworktree`,
+      thereby determining which items this iterator will yield.
 
     Yields
     ------
@@ -135,6 +140,7 @@ def iter_annexworktree(
         untracked=untracked,
         link_target=False,
         fp=False,
+        recursive=recursive,
     )
 
     git_fileinfo_store: list[Any] = list()
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -68,11 +68,20 @@ def content_path(item: AnnexWorktreeItem | AnnexWorktreeFileSystemItem,
     return item.link_target if link_target else item.name
 
 
-def get_annex_item(data):
+def get_annex_item(data, **kwargs):
+    # Reuse the work done in `iter_gitworktree`.
+    # TODO: this approach is a little wasteful because we have to pass `fp` and
+    #  `link_target` into `iter_gitworktree` to get the proper
+    #  `GitWorkTree*`-class. But that might lead `iter_gitworktree` to open a
+    #  file for us (if `fp` is `True`), which we do not use.
+    all_args = {
+        **data.__dict__,
+        **kwargs,
+    }
     if isinstance(data, GitWorktreeItem):
-        return AnnexWorktreeItem(**data.__dict__)
+        return AnnexWorktreeItem(**all_args)
     elif isinstance(data, GitWorktreeFileSystemItem):
-        return AnnexWorktreeFileSystemItem(**data.__dict__)
+        return AnnexWorktreeFileSystemItem(**all_args)
     else:
         raise TypeError(
             'Expected GitWorktreeItem or '
@@ -81,18 +90,19 @@ def get_annex_item(data):
 
 
 def join_annex_info(processed_data,
-                    stored_data: AnnexWorktreeItem | AnnexWorktreeFileSystemItem
+                    stored_data: GitWorktreeItem | GitWorktreeFileSystemItem,
                     ):
     if processed_data is StoreOnly:
-        return stored_data
+        return get_annex_item(stored_data)
     else:
-        if processed_data:
-            stored_data.annexkey = processed_data['key']
-            stored_data.annexsize = int(processed_data['bytesize'])
-            stored_data.annexobjpath = PurePath(
+        return get_annex_item(
+            stored_data,
+            annexkey=processed_data['key'],
+            annexsize=int(processed_data['bytesize']),
+            annexobjpath=PurePath(
                 PurePosixPath(str(processed_data['objectpath']))
-            )
-        return stored_data
+            ),
+        )
 
 
 def iter_annexworktree(
@@ -106,17 +116,17 @@ def iter_annexworktree(
     glsf = iter_gitworktree(
         path,
         untracked=untracked,
-        link_target=link_target,
-        # TODO: we have to pass `fp` to `iter_gitworktree` because we want to
-        #  get the attributes from `GitWorktreeFileSystemItem`. We only get those
-        #  if we specify `fp=True` or `link_target=True`. However, we cannot use
-        #  the file pointer that we receive from `iter_gitworktree`, because the
-        #  they are closed when the next item is consumed from
-        #  `iter_gitworktree`. This happens before we yield our results.
-        #  Therefore, we have to open the file again. This is not ideal, but
+        # TODO: we have to pass `fp` and `link_target` to `iter_gitworktree`
+        #  because we want to get the attributes from `GitWorktreeFileSystemItem`.
+        #  We only get those if we specify `fp=True` or `link_target=True`.
+        #  However, we cannot use the file pointer that we receive from
+        #  `iter_gitworktree`, because they are closed when the next item is
+        #  consumed from `iter_gitworktree`. This happens before we yield our
+        #  results. Therefore, we have to open the file again. This is not ideal, but
         #  it works for now. A better solution might be, for example, to factor
         #  the worktree code out and provide object-factories for
         #  `GitWorktreeFileSystemItem` and for `AnnexWorktreeFileSystemItem`.
+        link_target=link_target,
         fp=fp,
     )
 
@@ -142,10 +152,10 @@ def iter_annexworktree(
                         git_fileinfo_store,
                         lambda git_worktree_item: (
                                 str(git_worktree_item.name).encode(),
-                                get_annex_item(git_worktree_item),
+                                git_worktree_item
                         )
                     )
-                ),
+                )
             ) as gaf, \
             iter_subproc(
                 # get the key properties JSON-lines style
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -27,7 +27,7 @@ from datalad_next.itertools import (
     load_json,
     route_in,
     route_out,
-    dont_process,
+    StoreOnly,
 )
 from datalad_next.runners import iter_subproc
 
@@ -67,7 +67,7 @@ def get_annex_item(data):
 def join_annex_info(processed_data,
                     stored_data: AnnexWorktreeItem | AnnexWorktreeFileSystemItem
                     ):
-    if processed_data is dont_process:
+    if processed_data is StoreOnly:
         return stored_data
     else:
         if processed_data:
@@ -131,7 +131,7 @@ def iter_annexworktree(
                     # do not process empty key lines. Non-empty key lines
                     # are processed, but nothing needs to be stored because the
                     # processing result includes the key itself.
-                    lambda key: (dont_process, None)
+                    lambda key: (StoreOnly, None)
                                  if key == linesep_bytes
                                  else (key, None)
                 )
@@ -139,14 +139,14 @@ def iter_annexworktree(
 
         yield from route_in(
             # the following `route_in` yields processed keys for annexed
-            # files and `dont_process` for non-annexed files. Its
+            # files and `StoreOnly` for non-annexed files. Its
             # cardinality is the same as the cardinality of
             # `iter_gitworktree`, i.e. it produces data for each element
             # yielded by `iter_gitworktree`.
             route_in(
                 load_json(itemize(gek, sep=linesep_bytes)),
                 key_store,
-                # `processed` data is either `dont_process` or detailed
+                # `processed` data is either `StoreOnly` or detailed
                 # annex key information. we just return `process_data` as
                 # result, because `join_annex_info` knows how to incorporate
                 # it into an `AnnexWorktree*`-object.
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -115,33 +115,30 @@ def iter_annexworktree(
                 ['git', '-C', str(path),
                  'annex', 'examinekey', '--json', '--batch'],
                 # use only non-empty keys as input to `git annex examinekey`.
-                input=route_out(
-                    itemize(
-                        gaf,
-                        # git-annex changed its line-ending behavior, but we
-                        # should be safe, because we declare a specific format
-                        # for git-annex-find above
-                        # TODO this should be the following line only, not need
-                        # to put ends back
-                        # but MIH cannot get this adjusted
-                        #sep=b'\n',
-                        sep=None,
-                        keep_ends=True,
-                    ),
-                    # we need this route-out solely for the purpose
-                    # of maintaining a 1:1 relation ship of items reported
-                    # by git-ls-files and git-annex-find (merged again
-                    # in the route-in that gives `results` below`. The
-                    # "store" here does not actually store anything other than
-                    # `None`s
-                    _annex_git_align,
-                    # do not process empty key lines. Non-empty key lines
-                    # are processed, but nothing needs to be stored because the
-                    # processing result includes the key itself.
-                    # TODO when `keep_ends=True` is removed above, the strip()
-                    # should not be necessary anymore, but MIH cannot get
-                    # this adjusted
-                    lambda key: (key if key.strip() else StoreOnly, None)
+                input=intersperse(
+                    # Add line ending to submit the key to batch processing in
+                    # `git annex examinekey`.
+                    b'\n',
+                    route_out(
+                        itemize(
+                            gaf,
+                            # git-annex changed its line-ending behavior, but we
+                            # should be safe, because we declare a specific format
+                            # for git-annex-find above
+                            sep=b'\n',
+                        ),
+                        # we need this route-out solely for the purpose
+                        # of maintaining a 1:1 relation ship of items reported
+                        # by git-ls-files and git-annex-find (merged again
+                        # in the route-in that gives `results` below`. The
+                        # "store" here does not actually store anything other than
+                        # `None`s
+                        _annex_git_align,
+                        # do not process empty key lines. Non-empty key lines
+                        # are processed, but nothing needs to be stored because the
+                        # processing result includes the key itself.
+                        lambda key: (key if key else StoreOnly, None)
+                    )
                 )
             ) as gek:
 
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -19,6 +19,7 @@ from more_itertools import intersperse
 from .gitworktree import (
     GitWorktreeItem,
     GitWorktreeFileSystemItem,
+    FileSystemItemType,
     iter_gitworktree
 )
 from datalad_next.itertools import (
@@ -50,6 +51,7 @@ class AnnexWorktreeFileSystemItem(GitWorktreeFileSystemItem):
 
 def content_path(item: AnnexWorktreeItem | AnnexWorktreeFileSystemItem,
                  base_path: Path,
+                 link_target: bool,
                  ) -> PurePath | None:
 
     if item.annexobjpath:
@@ -57,7 +59,13 @@ def content_path(item: AnnexWorktreeItem | AnnexWorktreeFileSystemItem,
             item.annexobjpath \
             if (base_path / item.annexobjpath).exists() \
             else None
-    return item.name
+
+    # if this is not an annexed file, honor `link_target`-parameter.
+    if item.type == FileSystemItemType.file:
+        return item.name
+    elif item.type == FileSystemItemType.symlink:
+        return item.link_target if link_target is True else None
+    return item.link_target if link_target else item.name
 
 
 def get_annex_item(data):
@@ -180,7 +188,7 @@ def iter_annexworktree(
             # Ensure that `path` is a `Path` object
             path = Path(path)
             for item in results:
-                item_path = content_path(item, path)
+                item_path = content_path(item, path, link_target)
                 if item_path:
                     with (path / item_path).open('rb') as active_fp:
                         item.fp = active_fp
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -1,4 +1,6 @@
 """Report on the content of a Git-annex repository worktree
+
+The main functionality is provided by the :func:`iter_annexworktree()` function.
 """
 from __future__ import annotations
 
@@ -69,67 +71,63 @@ def iter_annexworktree(
         link_target: bool = False,
         fp: bool = False,
 ) -> Generator[AnnexWorktreeItem | AnnexWorktreeFileSystemItem, None, None]:
-    """Report work tree item of an annexed Git repository
+    """Companion to ``iter_gitworktree()`` for git-annex repositories
 
-    This iterator can be used to report on all tracked, and untracked
-    non-annexed content and on the annexed content of the work tree of an
-    annexed Git repository. This includes files that have been removed
-    from the work tree (deleted), unless their removal has already been staged.
+    This iterator wraps
+    :func:`~datalad_next.iter_collections.gitworktree.iter_gitworktree`.
+    For each item, it determines whether it is an annexed file. If so,
+    it amends the yielded item with information on the respective
+    annex key, the byte size of the key, and its (would-be) location
+    in the repository's annex.
 
-    For any tracked content, yielded items include type information, gitsha
-    as last known to Git, and annex information, if the file is annexed. Annex
-    information includes the key of the annexed item, the size of the annexed
-    item in bytes, and the path where the content of an annexed item will be
-    available, if it is present.
+    The basic semantics of all arguments are identical to
+    :func:`~datalad_next.iter_collections.gitworktree.iter_gitworktree`.
+    Importantly, with ``fp=True``, the annex object is opened directly,
+    if available. If not available, no attempt is made to open the associated
+    symlink or pointer file.
 
-    This iterator is based on
-    :func:`datalad_next.iter_collections.gitworktree.iter_gitworktree` and like
-    this, any yielded item reflects the last committed or staged content, not
-    the state of an unstaged modification in the work tree.
-
-    When no reporting of link targets or file-objects are requested, items of
+    With ``link_target`` and ``fp`` disabled items of
     type :class:`AnnexWorktreeItem` are yielded, otherwise
     :class:`AnnexWorktreeFileSystemItem` instances are yielded. In both cases,
-    ``gitsha``, ``gittype``, ``annexkey``, ``annexsize``, and ``annnexobjpath``
-    properties are provided. Either of ``gitsha`` and ``gittyoe`` being ``None``
-    indicates untracked work tree content. Either of ``annexkey``, ``annexsize``,
-    ``annexobjpath`` being ``None`` indicates non-annexed work tree content.
+    ``annexkey``, ``annexsize``, and ``annnexobjpath`` properties are provided.
 
     .. note::
-      The ``gitsha`` is not equivalent to a SHA1 hash of a file's content,
-      but is the SHA-type blob identifier as reported and used by Git.
-
-    .. note::
-      Although ``annexobjpath`` is always set for annexed content, that does not
-      imply that an object at this path actually exists. The latter will only
-      be the case if the annexed content is present in the work tree, typically
-      as a result of a `datalad get`- or `git annex get`-call.
+      Although ``annexobjpath`` is always set for annexed content, that does
+      not imply that an object at this path actually exists. The latter will
+      only be the case if the annexed content is present in the work tree,
+      typically as a result of a `datalad get`- or `git annex get`-call.
 
     Parameters
     ----------
     path: Path
-      Path of a directory in a Git repository to report on.
-      Please see
-      :func:`datalad_next.iter_collections.gitworktree.iter_gitworktree` for
-      details.
-    untracked: {'all', 'whole-dir', 'no-empty'} or `None`, optional
-      Please see
-      :func:`datalad_next.iter_collections.gitworktree.iter_gitworktree` for
-      details.
+      Path of a directory in a git-annex repository to report on. This
+      directory need not be the root directory of the repository, but
+      must be part of the repository's work tree.
+    untracked: {'all', 'whole-dir', 'no-empty'} or None, optional
+      If not ``None``, also reports on untracked work tree content.
+      ``all`` reports on any untracked file; ``whole-dir`` yields a single
+      report for a directory that is entirely untracked, and not individual
+      untracked files in it; ``no-empty-dir`` skips any reports on
+      untracked empty directories.
     link_target: bool, optional
-      If ``True`` and the item represents a symlink, the target of the symlink
-      is stored in the ``link_target`` attribute of the item.
+      If ``True``, information matching a
+      :class:`~datalad_next.iter_collections.utils.FileSystemItem`
+      will be included for each yielded item, and the targets of
+      any symlinks will be reported, too.
     fp: bool, optional
-      If ``True``, each file-type item includes a file-like object
-      to access the file's content, if the file is either: non-annexed, or if
-      the files is annexed and the content is locally available.
-      This file handle will be closed automatically when the next item is
-      yielded.
+      If ``True``, information matching a
+      :class:`~datalad_next.iter_collections.utils.FileSystemItem`
+      will be included for each yielded item, but without a
+      link target detection, unless ``link_target`` is given.
+      Moreover, each file-type item includes a file-like object
+      to access the file's content. This file handle will be closed
+      automatically when the next item is yielded.
 
     Yields
     ------
-    :class:`AnnexWorktreeItem` or `AnnexWorktreeFileSystemItem`
+    :class:`AnnexWorktreeItem` or :class:`AnnexWorktreeFileSystemItem`
     """
+
     glsf = iter_gitworktree(
         path,
         untracked=untracked,
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -8,7 +8,6 @@ from more_itertools import intersperse
 from pathlib import (
     Path,
     PurePath,
-    PurePosixPath,
 )
 from typing import (
     Any,
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -48,6 +48,18 @@ class AnnexWorktreeFileSystemItem(GitWorktreeFileSystemItem):
     annexobjpath: PurePath | None = None
 
 
+def content_path(item: AnnexWorktreeItem | AnnexWorktreeFileSystemItem,
+                 base_path: Path,
+                 ) -> PurePath | None:
+
+    if item.annexobjpath:
+        return \
+            item.annexobjpath \
+            if (base_path / item.annexobjpath).exists() \
+            else None
+    return item.name
+
+
 def get_annex_item(data):
     if isinstance(data, GitWorktreeItem):
         return AnnexWorktreeItem(**data.__dict__)
@@ -87,7 +99,17 @@ def iter_annexworktree(
         path,
         untracked=untracked,
         link_target=link_target,
-        fp=fp
+        # TODO: we have to pass `fp` to `iter_gitworktree` because we want to
+        #  get the attributes from `GitWorktreeFileSystemItem`. We only get those
+        #  if we specify `fp=True` or `link_target=True`. However, we cannot use
+        #  the file pointer that we receive from `iter_gitworktree`, because the
+        #  they are closed when the next item is consumed from
+        #  `iter_gitworktree`. This happens before we yield our results.
+        #  Therefore, we have to open the file again. This is not ideal, but
+        #  it works for now. A better solution might be, for example, to factor
+        #  the worktree code out and provide object-factories for
+        #  `GitWorktreeFileSystemItem` and for `AnnexWorktreeFileSystemItem`.
+        fp=fp,
     )
 
     git_fileinfo_store: list[Any] = list()
@@ -133,7 +155,7 @@ def iter_annexworktree(
                 )
             ) as gek:
 
-        yield from route_in(
+        results = route_in(
             # the following `route_in` yields processed keys for annexed
             # files and `StoreOnly` for non-annexed files. Its
             # cardinality is the same as the cardinality of
@@ -151,3 +173,18 @@ def iter_annexworktree(
             git_fileinfo_store,
             join_annex_info,
         )
+
+        if not fp:
+            yield from results
+        else:
+            # Ensure that `path` is a `Path` object
+            path = Path(path)
+            for item in results:
+                item_path = content_path(item, path)
+                if item_path:
+                    with (path / item_path).open('rb') as active_fp:
+                        item.fp = active_fp
+                        yield item
+                else:
+                    item.fp = None
+                    yield item
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -1,6 +1,7 @@
 """Report on the content of a Git-annex repository worktree
 
-The main functionality is provided by the :func:`iter_annexworktree()` function.
+The main functionality is provided by the :func:`iter_annexworktree()`
+function.
 """
 from __future__ import annotations
 
@@ -145,7 +146,8 @@ def iter_annexworktree(
                 # we get the annex key for any filename
                 # (or empty if not annexed)
                 ['git', '-C', str(path),
-                 'annex', 'find', '--anything', '--format=${key}\n', '--batch'],
+                 'annex', 'find', '--anything', '--format=${key}\n',
+                 '--batch'],
                 # intersperse items with newlines to trigger a batch run
                 # this avoids string operations to append newlines to items
                 input=intersperse(
@@ -178,9 +180,9 @@ def iter_annexworktree(
                     route_out(
                         itemize(
                             gaf,
-                            # git-annex changed its line-ending behavior, but we
-                            # should be safe, because we declare a specific format
-                            # for git-annex-find above
+                            # git-annex changed its line-ending behavior, but
+                            # we should be safe, because we declare a specific
+                            # format for git-annex-find above
                             sep=b'\n',
                         ),
                         # we need this route-out solely for the purpose
@@ -193,8 +195,8 @@ def iter_annexworktree(
                         # output of `git annex examinekey`).
                         _annex_git_align,
                         # do not process empty key lines. Non-empty key lines
-                        # are processed, but nothing needs to be stored because the
-                        # processing result includes the key itself.
+                        # are processed, but nothing needs to be stored because
+                        # the processing result includes the key itself.
                         lambda key: (key if key else StoreOnly, None)
                     )
                 )
--- a/datalad_next/iter_collections/annexworktree.py
+++ b/datalad_next/iter_collections/annexworktree.py
@@ -82,9 +82,10 @@ def iter_annexworktree(
     item in bytes, and the path where the content of an annexed item will be
     available, if it is present.
 
-    This iterator is based on :func:`iter_gitworktree` and like this, any
-    yielded item reflects the last committed or staged content, not the state
-    of an unstaged modification in the work tree.
+    This iterator is based on
+    :func:`datalad_next.iter_collections.gitworktree.iter_gitworktree` and like
+    this, any yielded item reflects the last committed or staged content, not
+    the state of an unstaged modification in the work tree.
 
     When no reporting of link targets or file-objects are requested, items of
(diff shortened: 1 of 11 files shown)


datalad_next/processors/__init__.py
```diff
@@ -0,0 +1,33 @@
+""" Iterators that support subprocess pipelining and output processing
```
mih, line 1: I am wondering if `processors` is a good name, and whether this framing is optimal.

It is true that some iterators in here are working with iterables of `bytes`, but most are agnostic.

To me this is really generics, pretty much like `itertools`.

datalad_next/processors/decode_processor.py
```diff
@@ -0,0 +1,75 @@
...
+                joined_data = b''
+            except UnicodeDecodeError as e:
+                # If an encoding error occurs, we first check whether it was
+                # in the middle of `joined_data` or whether it extends until the
+                # end of `joined_data`.
+                # If it occurred in the middle of
+                # `joined_data`, we replace it with backslash encoding or
+                # re-raise the decoding error.
+                # If it occurred at the end of `joined_data`, we wait for the
+                # next chunk, which might fix the problem.
```
mih, line 59: This cannot be a comment. It is essential for understanding what this tool does. It should be moved to the documentation.

In general, I wonder what the use case is for this helper? In yields decodeable chunks that are similiar, but not identical to the items of the input iterator.

Users may have similar concerns, and adding a clear recommendation when this would be useful should be in the documentation. I would think that it would be simpler to re-chunk a bytes "stream", and decode actual items -- rather than trial-and-error decoding.

datalad_next/processors/lines_processor.py
```diff
@@ -0,0 +1,123 @@
...
+                                ) -> Generator[bytes | str, None, None]:
+    assembled = None
+    for chunk in iterable:
+        if not assembled:
+            assembled = chunk
+        else:
+            assembled += chunk
+        lines = assembled.split(sep=separator)
+        if len(lines) == 1:
+            continue
```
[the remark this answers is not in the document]
mih: Ah, understood

datalad_next/processors/lines_processor.py
```diff
@@ -0,0 +1,123 @@
...
+    Parameters
+    ----------
+    iterable: Iterable[bytes | str]
+        The iterable that yields the input data
+    separator: str | bytes | None
+        The separator that is used to split the lines. If `None`, the lines are
+        split by the line-separators that are built into `splitlines`.
+    keep_ends: bool
+        If `True`, the line-separator will be contained in each yielded line. If
+        `False`, the line-separator will not be contained in the yielded lines.
```
mih, line 50: Maybe worth mentioning that this convenience implies a copy of each item?

datalad_next/processors/lines_processor.py
```diff
@@ -0,0 +1,123 @@
...
+from typing import (
+    Generator,
+    Iterable,
+)
+
+
+__all__ = ['lines_processor']
+
+
+def lines_processor(iterable: Iterable[bytes | str],
```
mih, line 14: I wonder what convinced you to reject the `finditer` based implementation from https://github.com/datalad/datalad-next/issues/537#issuecomment-1815895654?

I found this to be rather fast. Did you benchmark it? That implementation would also allow for a `keep_end` feature without an implied copy.

datalad_next/processors/pattern_processor.py
```diff
@@ -0,0 +1,77 @@
...
+
+    The processor guarantees:
+
+    1. All chunks have at minimum the size of the pattern
+    2. If a complete pattern exists, it will be contained completely within a
+       single chunk, i.e. it will NOT be the case that a prefix of the pattern
+       is at the end of a chunk, and the rest of the pattern in the beginning
+       of the next chunk
+
+    The pattern might be present multiple times in a data chunk.
```
mih, line 35: I think this docstring would benefit from an example. What is "pattern" exactly? Is it a regex? Is it a delimiter? If the latter, why would one want to have more than one in a yielded item, rather than using splitlines?

datalad_next/processors/lines_processor.py
christian-monch: I made a test, this implementation is about twice as fast

mih (commented):
Here are the performance metrics missing from the original PR:

This PR
```
>>> %timeit list(iter_annexworktree('.'))
4.33 s ± 587 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Sketch from https://github.com/datalad/datalad-next/issues/537 (include interpreter/import overhead)

```
1: python /home/mih/hacking/datalad/next/iter_subproc.py
real        4.096       0.325       3.812       3.864       4.565       
user        5.340       0.558       4.694       5.138       6.245       
sys         2.487       0.234       2.210       2.426       2.880       
```

Naked shell (excluding JSON decoding)

```
❯ multitime -n 5 git ls-files | git annex find --anything --format="\${key}\n" --batch | grep -v '^[[:space:]]*$' | git annex examinekey --json --batch > /dev/null
===> multitime results
1: git ls-files
real        3.473       0.222       3.270       3.420       3.903       
user        0.008       0.006       0.000       0.010       0.015       
sys         0.007       0.007       0.000       0.003       0.016       
```

Bottom line: This PR is 8% slower than the original sketch and 25% slower than the (rough, non-Python) baseline. We are getting there!

I am not sure, where the 8% slowdown is coming from.

mih:
Looking further into the timing:

Current `main`

```
>>> %timeit list(iter_gitworktree('.'))
220 ms ± 4.59 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This PR:

```
>>> %timeit list(iter_gitworktree('.'))
221 ms ± 907 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Same. And both 5x slower than the baseline:

```
multitime -n 5 git ls-files --stage --cached --exclude-standard --others' > /dev/null
real        0.038       0.007       0.029       0.043       0.046       
user        0.020       0.005       0.015       0.019       0.029       
sys         0.017       0.005       0.011       0.014       0.026       
```

This explains some of the 25% overall difference, but not the 8%.

datalad_next/processors/lines_processor.py
mih: I do not see this large speed-up:

```
❯ multitime -n10 python ~/hacking/datalad/next/bm_split_finditer.py
===> multitime results
1: python /home/mih/hacking/datalad/next/bm_split_finditer.py
real        0.215       0.021       0.196       0.209       0.268       
user        0.182       0.018       0.153       0.185       0.216       
sys         0.036       0.012       0.022       0.037       0.054       

❯ multitime -n10 python ~/hacking/datalad/next/bm_split_linesproc.py
===> multitime results
1: python /home/mih/hacking/datalad/next/bm_split_linesproc.py
real        0.212       0.018       0.192       0.210       0.244       
user        0.179       0.014       0.162       0.177       0.215       
sys         0.037       0.013       0.018       0.034       0.059       
```

Mean difference for me is below a standard deviation.
[bm_split_linesproc.py.txt](https://github.com/datalad/datalad-next/files/13436044/bm_split_linesproc.py.txt)
[bm_split_finditer.py.txt](https://github.com/datalad/datalad-next/files/13436045/bm_split_finditer.py.txt)

mih: I made a small modification that seems to be slightly faster:

```
❯ multitime -n10 python ~/hacking/datalad/next/bm_split_finditer.py
===> multitime results
1: python /home/mih/hacking/datalad/next/bm_split_finditer.py
real        0.205       0.011       0.197       0.199       0.227       
user        0.179       0.008       0.166       0.176       0.194       
sys         0.030       0.007       0.021       0.029       0.042       
```

[bm_split_finditer.py.txt](https://github.com/datalad/datalad-next/files/13437239/bm_split_finditer.py.txt)

datalad_next/processors/__init__.py
christian-monch: I also wondered about the name. Changed it to `datalad_next.itertools` in ff8344485ea111661f960f95d331326a318df37b

datalad_next/processors/lines_processor.py
mih: I had the hunch that the interpreter/import overhead is disproportionally larger that the implementation performance differences. So I changes the brenchmark scripts to contain an importable function that I could more easily benchmark after all imports are made. This should give a more relevant metric:

This is now much closer to the "about twice" assessment.

```
In [3]: %timeit run_iter()
13.4 ms ± 65.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit run_linesproc()
8.26 ms ± 87.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

mih:
After normalizing the benchmark conditions further, I have the following stats:

```
# this code
In [6]: %timeit list(iter_annexworktree('.'))
3.59 s ± 104 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# original sketch
In [7]: %timeit run_itersubproc()
4.17 s ± 317 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# baseline with iterable_subprocess
In [9]: %timeit list(iter_gitworktree('.'))
302 ms ± 8.02 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This compared to a shell pipe (using `jq` as consumer that added the JSON decoding load):

```
❯ multitime -n 5 git ls-files | git annex find --anything --format="\${key}\n" --batch | grep -v '^[[:space:]]*$' | git annex examinekey --json --batch | jq > /dev/null
===> multitime results
1: git ls-files
real        3.399       0.169       3.236       3.369       3.703       
user        0.006       0.005       0.000       0.003       0.012       
sys         0.006       0.006       0.000       0.008       0.015       
```

~8% difference, and nearly within the margin of error.

This works for me!

datalad_next/processors/decode_processor.py
christian-monch: > This cannot be a comment. It is essential for understanding what this tool does. It should be moved to the documentation.

It is documented in the interface documentation for the `backslash_replace`-argument:

```
    backslash_replace: bool (default: ``True``)
        If ``True``, backslash-escapes are used for undecodable bytes. If
        ``False``, a ``UnicodeDecodeError`` is raised if a byte sequence cannot
        be decoded.
```

> In general, I wonder what the use case is for this helper? In yields decodeable chunks that are similiar, but not identical to the items of the input iterator.

The use case is to properly decode bytes into unicode characters, even if the encoding crosses a chunk-border. The simple `map(decode, chunk_source)` would not work, if the encoded unicode character is split between two chunks.
The function `decode_processor` fulfills an equivalent function as `datalad.runner.utils.AssemblingDecoderMixIn`. Tha latter was introduced to fix datalad's issue https://github.com/datalad/datalad/issues/6526.

In my view, this should always be used if a user wants to decode streamed bytes into strings.

datalad_next/processors/json_processor.py
```diff
@@ -0,0 +1,72 @@
...
+    tuple[Any, bool]
+        A tuple containing the JSON-object that are generated from the data
+        yielded by ``iterable`` and a flag that indicates whether the JSON
+        decoding was successful
+    """
+    for json_string in iterable:
+        try:
+            yield json.loads(json_string), True
+        except json.decoder.JSONDecodeError as e:
+            yield e, False
```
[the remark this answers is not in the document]
christian-monch: The invalid string is available in `e.doc`.

Documentation and type annotations are updated in 5192629365a495cd4216dbbc727cf178eb64d312

datalad_next/processors/lines_processor.py
christian-monch: That is correct, appending the line-ends decreases performance. It is mentioned in the description, but I now mentioned it in the parameter description as well (in c779d60893b20c9c3800447c042a9f6be604af4b)

datalad_next/processors/reroute_processor.py
```diff
@@ -0,0 +1,163 @@
...
+    Iterable,
+)
+
+
+__all__ = ['route_in', 'route_out', 'join_with_list']
+
+_active_fifos: dict[str, list] = defaultdict(list)
+
+
+def route_out(iterable: Iterable,
```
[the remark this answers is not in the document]
christian-monch: It does not have to be an internal data structure. A passed-in `list` (or `Queue`, but `list` is faster) would work as well. The motivation for the "identifier"-string was to be able to swap the implementation internally if a faster FIFO implementation is found. The number of lines that are dedicated to the mapping of "identifier"-strings to lists is relatively small though: 4 lines.

The remaining lines are dedicated to managing the "route data around iterables". The main motivations for dedicated `route_out`- and `route_in`-functions were:

- carry additional data, e.g. filenames the correspond with keys, alongside data that is processed by iterators
- allow to skip processing by an iterator, if the data item is not applicable, e.g. an empty key and an `iterable_subprocess` that executes `git annex examinekey --batch` (I am aware that in this specific example a fake-key could be used and filtered out later, but such an approach might not always be possible).

The functions `route_out` and `route_in` support this in general cases.

datalad_next/processors/lines_processor.py
mih, line 14: Kept thinking about the name. Something active would be good. What about `itemize`?

datalad_next/processors/lines_processor.py
```diff
@@ -0,0 +1,123 @@
...
+                    separator: str | bytes | None = None,
```
mih, line 15: What about `separator` -> `sep`? This would align the names with `str.split()` and `str.splitlines()`

datalad_next/processors/lines_processor.py
```diff
@@ -0,0 +1,123 @@
...
+                                separator: str | bytes,
+                                keep_ends: bool = False,
+                                ) -> Generator[bytes | str, None, None]:
+    assembled = None
+    for chunk in iterable:
+        if not assembled:
+            assembled = chunk
+        else:
+            assembled += chunk
+        lines = assembled.split(sep=separator)
```
mih, line 78: I tried what happens of `assembled` is large and contains no separator. It does NOT double the memory demand.

datalad_next/processors/decode_processor.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import (
+    Generator,
+    Iterable,
+)
+
+
+__all__ = ['decode_processor']
+
+
+def decode_processor(iterable: Iterable[bytes],
```
mih, line 14: What about `decode_processor` -> `decode_bytes`? It is made for this input only, and give an "active" name that is similar to other itertools.

datalad_next/processors/json_processor.py
```diff
@@ -0,0 +1,72 @@
...
+    Any,
+    Generator,
+    Iterable,
+)
+
+
+__all__ = ['json_processor', 'json_processor_with_flag']
+
+
+def json_processor(iterable: Iterable[bytes | str],
```
mih, line 16: `json_processor` -> `load_json`?

christian-monch: will do

datalad_next/processors/decode_processor.py
christian-monch: will do

datalad_next/processors/lines_processor.py
christian-monch: Not sure. It does assembling and splitting to yield only complete lines. Maybe `complete_lines`? But that seems a little too long. How about `lines`?

mih: A line somehow needs to come with a lineend. What about zero-bytes. Not a line, but an item.

datalad_next/processors/decode_processor.py
mih: done in #538

datalad_next/processors/lines_processor.py
mih: #538 has a variant with a more generic description. works well, I think.

mih:
#538 is merged now. Please rebase on `main`. Thanks!

datalad_next/processors/lines_processor.py
christian-monch: Will do

mih: Thought about this again. Not a strong preference. It also makes sense to have it spelled out.

datalad_next/processors/pattern_processor.py
christian-monch: started with examples in `datalad_next.itertools`

datalad_next/processors/lines_processor.py
christian-monch: I adopted the name from your merge PR, i.e. `itemize`.

I am not too happy with it though, it seems too generic, especially because an `itemize(iterable)` without any further parameters will yield lines.

What about one of the following?

- `iter_lines`
- `build_lines`
- `line_builder`
- `linify`  (don't take this one too seriously)

christian-monch: > What about `separator` -> `sep`? This would align the names with `str.split()` and `str.splitlines()`

Done in f309fb3ebcbf670bb8ba9c00e316ba19bef68af0 (and used in 9468963e978af249ee1063e6166af261bacc6c63)

mih: My thinking is this: The only use case so far does not involved "lines" (zero-byte delimited items).

Actually, there is quite some complexity in the code to also mimic `splitlines()`. I considered ripping it out, but it clearly has value.

I would be OK (and even in favor of) not having a default argument for `separator`.

I am currently working on a 2nd use case for the newly merged code and it involved tab-delimited items (again no lines).

Let's remove the default argument?

datalad_next/processors/json_processor.py
christian-monch: done

datalad_next/processors/lines_processor.py
christian-monch: > My thinking is this: The only use case so far does not involved "lines" (zero-byte delimited items).
> 

I think that is not entirely correct. We process "proper" lines as result of `git annex examinekey --batch` output.

> [...]
>
> Let's remove the default argument?

Should we keep the option to pass `None`? I think it would make sense, because python's builtin `splitline` knows about Unix, OSX, and Windows line terminators. In addition, if operating on strings, it knows about a number of Unicode characters that terminate lines. The latter might not really be relevant, but an OS-agnostic splitlines-operation might come in handy.

mih: > Should we keep the option to pass `None`?

Yes, that is what I meant by "but it clearly has value"

christian-monch: Let's remove the default argument. I approved the PR

datalad_next/processors/reroute_processor.py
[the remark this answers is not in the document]
christian-monch: I changed the identifier string to a list-object in 256cdf970a922fb8564c014e20ae0cbe38bbd6e4

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,125 @@
...
+    route_in,
+    route_out,
+)
+from datalad_next.runners import iter_subproc
+
+lgr = logging.getLogger('datalad.ext.next.iter_collections.annexworktree')
+
+
+# TODO Could be `StrEnum`, came with PY3.11
+class AnnexTreeItemType(Enum):
```
mih, line 38: This is the same as `GitTreeItemType`. I guess it should derive from that, to get a dedicated type, but not change it or add to it. I think this would be better than an unrelated copy.

datalad_next/itertools/tests/test_reroute.py
```diff
@@ -0,0 +1,35 @@
...
+        join_with_list
+    )
+    # The result should be a list in which every odd element consists of a list
+    # with the elements `[n / 2.0, n]` and every even element consists of a list
+    # with `[None, 0]`, because the `0`s were routed around the consumer, i.e.
+    # around `lambda x: x / 2.0`.
+    assert list(r) == list(
+        intersperse([None, 0], map(lambda x: [x / 2.0, x], range(2, 20)))
+    )
+    print(list(r))
```
mih, line 35: The tests should not have output that is avoidable.

docs/source/pyutils.rst
```diff
@@ -25,6 +25,7 @@ packages.
    exceptions
    itertools
    iter_collections
+   itertools
```
mih, line 28: This is not needed. It already exists two lines above.

datalad_next/itertools/itemize.py
```diff
@@ -13,20 +13,21 @@
...
-    """ Generator that emits only complete items from chunks of an iterable
+    """ Function that only yields complete items, assembled from an iterable
 
-    This generator consumes chunks from an iterable and yields items defined by
+    This function consumes chunks from an iterable and yields items defined by
     a separator. An item might span multiple input chunks.
 
-    Items are defined by a ``separator``. If ``separator`` is ``None``, the
-    line-separators built into `str.plitlines` are used.
+    Items are defined by a separator given in ``sep``. If ``separator`` is
```
mih, line 24:
```suggestion
    Items are defined by a separator given in ``sep``. If ``sep`` is
```

datalad_next/itertools/__init__.py
```diff
@@ -1,13 +1,270 @@
 """Various iterators, e.g., for subprocess pipelining and output processing
 
+This module provides iterators that are useful when combining
+iterable_subprocesses and for processing of subprocess output.
```
mih, line 4: Style remark: I think this is superfluous. It is largely redundant with the line above, and the additional statement about `iterable_subprocess` artificially limits the scope. From my POV these are pretty generic iterators, we plan to use them with this tools, but by themselves they do not require that.

datalad_next/itertools/__init__.py
```diff
@@ -1,13 +1,270 @@
...
 
 
+from .align_pattern import align_pattern
 from .decode_bytes import decode_bytes
 from .itemize import itemize
+from .load_json import (
+    load_json,
+    load_json_with_flag,
+)
+from .decode_bytes import decode_bytes
```
mih, line 265: Duplicate of line 259

datalad_next/itertools/__init__.py
```diff
@@ -1,13 +1,270 @@
...
+The performance of ``itemize`` with a non-``None`` separator is slightly lower,
+if ``keep_ends`` is ``True``.
+
+Examples:
+
+.. code-block:: python
+
+    >>> from datalad_next.itertools import itemize
+    >>> with open ('/etc/passwd', 'rt') as f:
+    ...     print(tuple(itemize(iter(f.read, '')))[0:2])
```
mih, line 156: I think this would not work anymore with a mandatory `sep`

datalad_next/itertools/__init__.py
```diff
@@ -1,13 +1,270 @@
...
+the items would be terminated by Unix line-endings, i.e. each item would be a
+single line. The separator could also be, for example, ``b'\\x00'`` (or
+``'\\x00'``), in which case ``itemize`` can be used to separate output items of
+commands like ``git ls-files -z``, which terminate their output items with
+``b'\\x00'``.
+
+Separators can be longer than one byte or character, e.g. ``b'\\r\\n'``, or
+``b'\\n-------------------\\n'``.
+
+If ``sep`` is ``None`` (the default), ``itemize`` will use the line-separators
```
mih, line 138: It is no longer a default

datalad_next/iter_collections/annexworktree.py
christian-monch: That would be nice, but enumerations with members, i.e. `GitTreeItemType` cannot be subclassed (see https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing).

datalad_next/itertools/__init__.py
```diff
@@ -1,13 +1,270 @@
...
+    >>> tuple(decode_bytes([b'\\xc3']))
+    ('\\xc3',)
+    >>> tuple(decode_bytes([b'\\xc3'], backslash_replace=False))
+    Traceback (most recent call last):
+        ...
+    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 1: invalid continuation byte
+
+Note that backslash-replacement of undecodable bygtes does not result in an
+unambiguous mapping, because, for example, ``b'\\xc3'`` might
+be present in the input chunks.
```
mih, line 115: It is unclear to me what that means.

mih (commented):
`align_pattern()` is introduced here (with tests), but it unused. I would prefer to keep PRs self contained. Concretely, I would prefer to not introduce `align_pattern` until it is needed, and then introduce it in that context.

Which likely is https://github.com/datalad/datalad-next/pull/546

datalad_next/processors/pattern_processor.py
mih: In https://github.com/datalad/datalad-next/pull/548 i moved the docs into the docstrings of the respective functions. This helps to remove redundancies (which have substantial maintenance cost).

datalad_next/itertools/__init__.py
mih: Ok, so:

```
Note that backslash-replacement of undecodable bytes does not result in an
unambiguous mapping, because, for example, ``b'\\xc3'`` might
be present in the input chunks.
```

should become

```
Backslash-replacement of undecodable bytes is an ambiguous mapping, because, for example, ``b'\\xc3'`` can already be present in the input.
```

datalad_next/itertools/__init__.py
mih: Fixed in https://github.com/datalad/datalad-next/pull/548

datalad_next/itertools/__init__.py
christian-monch: Thanks @mih Much better!

datalad_next/iter_collections/annexworktree.py
mih: Ah, true. I wonder if we should even use a separate `enum` then.

datalad_next/itertools/__init__.py
mih: Fixed in https://github.com/datalad/datalad-next/pull/548

mih:
Ideally, this PR and https://github.com/datalad/datalad-next/pull/546 are now orthogonal and could be reviewed independently of each other.

christian-monch:
> `align_pattern()` is introduced here (with tests), but it unused. I would prefer to keep PRs self contained. Concretely, I would prefer to not introduce `align_pattern` until it is needed, and then introduce it in that context.

Ok, I will remove it from the PR

datalad_next/iter_collections/annexworktree.py
christian-monch: I removed it

datalad_next/itertools/__init__.py
christian-monch: Removed.

datalad_next/itertools/__init__.py
christian-monch: Thx, removed

datalad_next/itertools/reroute.py
```diff
@@ -0,0 +1,232 @@
...
+
+__all__ = ['dont_process', 'join_with_list', 'route_in', 'route_out']
+
+
+dont_process = object()
+
+
+def route_out(iterable: Iterable,
+              data_store: list,
+              splitter: Callable[[Any], tuple[Any, Any | None]],
```
mih, line 21: What is the meaning of the second item in the return tuple being allowed `None`? This seems undocumented.

datalad_next/itertools/reroute.py
```diff
@@ -0,0 +1,232 @@
...
+
+    To determine which data is to be yielded to the consumer and which data
+    should only be stored but not yielded, :func:`route_out` calls
+    :func:`splitter`. :func:`splitter` is called for each element of the input
+    iterable, with the element as sole argument. The function should return a
+    tuple of two elements. The first element is the data that is to be
+    yielded to the consumer. The second element is the data that is to be
+    stored in the list ``data_store``. If the first element of the tuple is
+    ``datalad_next.itertools.dont_process``, no data is yielded to the
+    consumer.
```
mih, line 39: I wonder if the `dont_process` being specific to the first item is more complex then needed and more limiting then necessary. What about making it `Ignore` and have it be applicable to yield and store independently (item 0 and 1 respectively).

Whenever `Ignore` is found, the corresponding item is ignored. Either not yielded, and/or not stored. I think a use case where all items should be yielded and only some routed around is not far-fetched.

PS: in the docstring `element` is used instead of `item`, which seems to be a more standard term.

datalad_next/itertools/reroute.py
```diff
@@ -0,0 +1,232 @@
...
+
+        def joiner(processed_data, stored_data):
+            #
+            return nan if processed_data is dont_process else processed_data
+
+        numbers = [0, 1, 0, 2, 0, 3, 0, 4]
+        store = list()
+        r = route_in(
+            map(
+                lambda x: x / 2.0,
```
mih, line 80: The example feels a bit weird. 0/2 is 0 not nan. I think the order of variable should be switched here.

datalad_next/itertools/reroute.py
```diff
@@ -0,0 +1,232 @@
...
+    .. code-block:: python
+
+        from math import nan
+        from datalad_next.itertools import route_out, route_in, dont_process
+
+        def splitter(data):
+            # if data == 0, return `dont_process` in the first element of the
+            # result tuple to indicate that route_out should not yield this
+            # element to its consumer
+            return (dont_process, [data]) if data == 0 else (data, [data])
```
mih, line 70: I am placing this comment here, but this is just one example.

It seems that this wrapping of `data` into a list for storage `[data]` is not only a pattern, but the pattern. However, I do not understand immediately which this complication is necessary.

The only motivation that I can see is that `join_with_list` requires it. However, in this example this particular joiner is not used.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,123 @@
...
+                                 else (data, [data])
+                )
+            ) as gek:
+
+        for item in route_in(
+                route_in(
+                    load_json(itemize(gek, sep=linesep_bytes)),
+                    key_store,
+                    join_with_list,
+                ),
```
mih, line 110: What is the purpose of this `route_in`? It seems to join the keyname with the output record of `examinekey` for that keyname. However, the keyname is already present in that record.

Maybe a comment can clarify the purpose.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,123 @@
...
+        ):
+            yield AnnexWorktreeItem(
+                name=item[2].name,
+                gitsha=item[2].gitsha,
+                gittype=item[2].gittype,
+                annexkey=item[1].decode().strip() if item[1] else None,
+                annexsize=int(item[0]['bytesize']) if item[0] else None,
+                annexobjpath=PurePath(PurePosixPath(str(item[0]['objectpath'])))
+                             if item[0]
+                             else None,
```
mih, line 122: I think the semantics of the items in `item` deserve a comment. `item[0]` would be the `examinekey` record. `item[1]` is the keyname, which is also in `item[0]['key']`. `item[2]` is the gitworktree item.

With a different joiner (than `join_with_list`), this could clearer. I am under the impression that the joiner could produce `AnnexWorktreeItem` instances directly, rather than producing an intermediate container.

mih (commented):
Thanks! This looks clean. My comments focus on a few aspects that may either be misunderstanding (hinting at documentation gaps), or could lead to a tuned design.

Re `iter_annexworktree()` itself: I believe it currently ignores the whole `fp=True` code branch entirely. It does the processing, but hides the outcome.

datalad_next/itertools/load_json.py
```diff
@@ -0,0 +1,112 @@
...
+
+def load_json(iterable: Iterable[bytes | str],
+              ) -> Generator[Any, None, None]:
+    """ Convert items yielded by ``iterable`` into JSON objects and yield them
+
+    ``load_json`` fetches items from the underlying
+    iterable. The items are expected to be ``bytes``, ``str``, or ``bytearry``,
+    and contain one JSON-encoded object. ``load_json``
+    will convert each item into a JSON-object, by feeding it into
+    ``json.loads``.
```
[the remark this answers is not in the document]
christian-monch: Good suggestion thanks

datalad_next/itertools/reroute.py
christian-monch: Yes, of course. Thanks for catching that. Done in 9edab7517295a2a8007b562933c5ecb6910fac41

datalad_next/iter_collections/annexworktree.py
christian-monch: The purpose is twofold:

1. inform the consumer whether a key was present for an item of `git_fileinfo_store` or not. (If it was present, processing through `examinekey` was peformed and the key record will be returned.  If it was not present, `examinekey` was not performed and `None` will be returned.)
2. "upscale" the results of gek (# annexed files) to the cardinality of `git_fileinfo_store` (# git files)

datalad_next/itertools/reroute.py
christian-monch: > I wonder if the `dont_process` being specific to the first item is more complex then needed and more limiting then necessary. What about making it `Ignore` and have it be applicable to yield and store independently (item 0 and 1 respectively).
> 

In order to keep the cardinality of the underlying iterator (so that it can be reproduced at the corresponding `route_in`-call), `route_out` will store information for every item from the underlying iterator. If a user does not need to store anything, he could pass `None` (or `b''` or a similar value that indicates "no data" to him) in element 1 of the result. It seems to me that applying `Ignore` to element 1 would be more restrictive 


> Whenever `Ignore` is found, the corresponding item is ignored. Either not yielded, and/or not stored. I think a use case where all items should be yielded and only some routed around is not far-fetched.

iter_annexworktree with a lot of annexed files and only a few non-annexed files is actually that use case. As mentioned above, the number of ignored and not-ignored elements has no influence on the maximum cardinality of the data store. Each item of the underlying iterable has an entry in the data store. 


> PS: in the docstring `element` is used instead of `item`, which seems to be a more standard term.

I changed the docstring to refer to `items` if we talk about members of an `iterable`, and to `elements` if we talk about members of a `tuple` (which is technically also iterable, but we use it here to pass results). Fixed in a9d56e4d2e7d929d1bf8bb52197690c4d716b99e

datalad_next/itertools/reroute.py
christian-monch: That is actually not necessary, since `Any` includes `None` and because the second element of the result-tuple is not inspected by `route_out`.

`None` is removed in eca443038f4a73edd52b4713d889636c55f8bfdf

datalad_next/itertools/reroute.py
christian-monch: Removed `join_with_list` from the code, it advertises a suboptimal paradigm. Done in 917b529b082f1dead545bdc0844f565ed227bf9c

christian-monch: `iter_annexworktree` adjusted in 1e1d3e57fab42ebd1b9b20c1e6e5d91ea69473ee

datalad_next/iter_collections/annexworktree.py
christian-monch: Comment updated in 1e1d3e57fab42ebd1b9b20c1e6e5d91ea69473ee

datalad_next/iter_collections/annexworktree.py
christian-monch: > With a different joiner (than join_with_list), this could clearer. I am under the impression that the joiner could produce AnnexWorktreeItem instances directly, rather than producing an intermediate container.

That is a very good idea. It is applied in 1e1d3e57fab42ebd1b9b20c1e6e5d91ea69473ee

christian-monch:
> Thanks! This looks clean. My comments focus on a few aspects that may either be misunderstanding (hinting at documentation gaps), or could lead to a tuned design.
> 
> Re `iter_annexworktree()` itself: I believe it currently ignores the whole `fp=True` code branch entirely. It does the processing, but hides the outcome.

Fixed in 1e1d3e57fab42ebd1b9b20c1e6e5d91ea69473ee

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,123 @@
...
+        untracked: str | None = 'all',
+        link_target: bool = False,
+        fp: bool = False,
+) -> Generator[AnnexWorktreeItem | AnnexWorktreeFileSystemItem, None, None]:
+
+    glsf = iter_gitworktree(
+        path,
+        untracked=untracked,
+        link_target=link_target,
+        fp=fp
```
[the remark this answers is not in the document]
christian-monch: > I wonder if these two can/should be delegated to `iter_worktree()`.

It seems sensible to factor out the common code from `iter_gitworktree` and `iter_annexworktree`.

> If something is actually an annexed file, but not locally available, the file pointer would be a symlink, or a pointer file. For someone using `iter_annexworktree`, this might come as a surprise and would also require platform-aware client code.

Currently, no file-pointer is returned for a symlink (file-pointers are just copied from the corresponding `iter_gitworktree`-result).

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,157 @@
...
+    route_out,
+    dont_process,
+)
+from datalad_next.runners import iter_subproc
+
+
+lgr = logging.getLogger('datalad.ext.next.iter_collections.annexworktree')
+
+
+linesep_bytes = linesep.encode()
```
mih, line 38: https://git-annex.branchable.com/news/version_10.20231129/

Git annex changes lineending behavior. I believe it would break this choice

datalad_next/itertools/reroute.py
```diff
@@ -0,0 +1,232 @@
...
+    Callable,
+    Generator,
+    Iterable,
+)
+
+
+__all__ = ['dont_process', 'join_with_list', 'route_in', 'route_out']
+
+
+dont_process = object()
```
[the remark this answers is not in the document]
christian-monch: I will move to the pattern of using custom types.
I agree that the name should be improved. I think `DontYield`, `Drop`, `Ignore`, or `Nothing` would fit, since they appear in the element of the result-tuple that is either yielded or not yielded. Looking at them:

Some examples in which `data.name` is stored and nothing should be yielded to the consumer:

```python
return (DontYield, data.name)

return (Drop, data.name)

return (Ignore, data.name)

return (Nothing, data.name)

```

Alternatively we could use a dataclass for the result of a splitter-function which would make the semantics clearer:

```python

return SplitterResult(yield_value=DontYield, store=data.name)

return SplitterResult(yield_value=Drop, store=data.name)

return SplitterResult(yield_value=Ignore, store=data.name)

return SplitterResult(yield_value=Nothing, store=data.name)
```

datalad_next/itertools/load_json.py
```diff
@@ -0,0 +1,112 @@
...
+    Generator,
+    Iterable,
+)
+
+
+__all__ = ['load_json', 'load_json_with_flag']
+
+
+def load_json(iterable: Iterable[bytes | str],
+              ) -> Generator[Any, None, None]:
```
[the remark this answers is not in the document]
christian-monch: I keep it as it is then.

datalad_next/iter_collections/annexworktree.py
christian-monch: Thanks for the pointer. I changed the `itemize`-calls to use `sep=None`. This should take care of all known variants of line-endings.

Done in 28be76c2a96518899ae9fc4353ab04c4f0f12257

mih: I assume that means we pay an unconditional cost on non-windows platforms for that now. Maybe `sep=None` would have been better as a windows-only setting?

mih (commented):
The `fp=True` related functionality is not working for me -- and I don't see how it could work:

```py
>>> from datalad_next.iter_collections.annexworktree import iter_annexworktree
>>> l=list(iter_annexworktree('.', fp=True))
...
File ~/hacking/datalad/next/datalad_next/iter_collections/utils.py:79, in FileSystemItem.from_path(cls, path, link_target)
     66 @classmethod
     67 def from_path(
     68     cls,
   (...)
     71     link_target: bool = True,
     72 ):
     73     """Populate item properties from a single `stat` and `readlink` call
     74 
     75     The given ``path`` must exist. The ``link_target`` flag indicates
     76     whether to report the result of ``readlink`` for a symlink-type
     77     path.
     78     """
---> 79     cstat = path.lstat()
     80     cmode = cstat.st_mode
     81     if stat.S_ISLNK(cmode):

```

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,208 @@
...
+
+        if not fp:
+            yield from results
+        else:
+            # Ensure that `path` is a `Path` object
+            path = Path(path)
+            for item in results:
+                item_path = content_path(item, path, link_target)
+                if item_path:
+                    with (path / item_path).open('rb') as active_fp:
```
mih, line 203: This does not work in general, because it does not anticipate directory records, which can come from untracked content.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,208 @@
...
+
+    # if this is not an annexed file, honor `link_target`-parameter.
+    if item.type == FileSystemItemType.file:
+        return item.name
+    elif item.type == FileSystemItemType.symlink:
+        return item.link_target if link_target is True else None
+    return item.link_target if link_target else item.name
+
+
+def get_annex_item(data, **kwargs):
```
mih, line 71: I think it would help to rename this to something like `get_worktree_item`. it is true that and `Annex***Item` comes back, but one can have the misconception that any item refers to an annexed file. This is not the case.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,153 @@
...
+        return stored_data
+
+
+def iter_annexworktree(
+        path: Path,
+        *,
+        untracked: str | None = 'all',
+        link_target: bool = False,
+        fp: bool = False,
+) -> Generator[AnnexWorktreeItem | AnnexWorktreeFileSystemItem, None, None]:
```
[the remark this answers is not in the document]
christian-monch: Yes.

datalad_next/iter_collections/annexworktree.py
christian-monch: That is a good point, I will change it.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,208 @@
...
+def content_path(item: AnnexWorktreeItem | AnnexWorktreeFileSystemItem,
+                 base_path: Path,
+                 link_target: bool,
+                 ) -> PurePath | None:
+
+    if item.annexobjpath:
+        return \
+            item.annexobjpath \
+            if (base_path / item.annexobjpath).exists() \
+            else None
```
mih, line 61: I believe this does not work. There is no guarantee that `base_path` is the root of the repository.

datalad_next/iter_collections/annexworktree.py
christian-monch: Thanks for catching that. There is a bug in `content_path`, which should only return existing annex-object paths, a path to an existing file, or a symlink target.

Will merge the fix after you pushed your changes

datalad_next/iter_collections/annexworktree.py
christian-monch: I think it will work, because we use `-C base_path` when calling `git annex examinekey` (if `base_path` is not the root of the repository `git -C base_path annex examinekey` will yield an object-path so that `base_path / item.annexobjpath` will point to the annexed content).

But I might not be aware of all possible pitfalls here.

mih: You are corrrect

datalad_next/iter_collections/annexworktree.py
mih: This can be removed again. We declare the format.

datalad_next/iter_collections/annexworktree.py
```diff
@@ -0,0 +1,317 @@
...
+    # annex object path, relative to the item
+    annexobjpath: PurePath | None = None
+
+
+# TODO this iterator should get a filter mechanism to limit it to a single
+# directory (non-recursive). This will be needed for gooey.
+# unlike iter_gitworktree() we pay a larger dedicated per item cost.
+# Given that the switch to iterative processing is also made for
+# iter_gitworktree() we should provide the same filtering for that one
+# too!
```
mih, line 64: #552 is ready for inspection and would address this TODO here.

mih:
Current state of implementation is between 5% (Python) and 15% (CLI) slower than a rough shell equivalent:

```
❯ multitime -n 5 datalad -f disabled ls-file-collection annexworktree .
===> multitime results
1: datalad -f disabled ls-file-collection annexworktree .
real        3.767       0.042       3.697       3.778       3.823       
user        5.569       0.143       5.320       5.586       5.765       
sys         2.264       0.104       2.165       2.194       2.427       

❯ multitime -n 5 git ls-files | git annex find --anything --format="\${key}\n" --batch | grep -v '^[[:space:]]*$' | git annex examinekey --json --batch | jq > /dev/null
===> multitime results
1: git ls-files
real        3.289       0.062       3.207       3.275       3.392       
user        0.009       0.003       0.006       0.007       0.013       
sys         0.004       0.003       0.000       0.005       0.007       

In [3]: %timeit list(iter_annexworktree(Path('.')))
3.48 s ± 39.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

datalad_next/iter_collections/tests/test_iterannexworktree.py
```diff
@@ -0,0 +1,101 @@
...
+    ds = _mkds(tmp_path_factory, monkeypatch, {})
+    for i in range(3):
+        (ds.pathobj / f'file_{i}').write_text(f'content: #ö file_{i}\n')
+    ds.save(result_renderer='disabled')
+    ds.drop(
+        ds.pathobj / 'file_1',
+        reckless='availability',
+        result_renderer='disabled'
+    )
+    for ai in filter(lambda i: str(i.name).startswith('file_'), iter_annexworktree(ds.pathobj, fp=True)):
```
mih, line 97: The missing coverage warnings already point it out: the code never runs. The reason is that paths/names are absolute and never start with `file_`

mih: Test adjusted.

datalad_next/iter_collections/annexworktree.py
mih: Done

mih:
```
FAILED ..\iter_collections\tests\test_iterannexworktree.py::test_iter_annexworktree_basic_fp - UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf6 in position 10: invalid start byte
```

Something is still broken. The test condition existed for some time, but the test never actually ran.

mih:
#555 brings helpers that can be used to remove duplication of `fp=True` handling.

PR #544, opened 2023-11-23
Replace regex approach to `ls-files` output parsing

### LINKED ISSUE #540: `iter_gitworktree()` too slow
See https://github.com/datalad/datalad-next/pull/539#issuecomment-1822331093 for the benchmark.

mih:
In #540 I found about ~10% of the runtime to be attributable to the regex-matching. This patch replaces the regex with two split() calls. In local benchmarks I see a ~10% speedup.

One major other source of slow-down was/is the construction of `PurePosixPath` objects. It seems this is being taken care of by https://github.com/python/cpython/issues/101362 and will be resolved eventually.

Closes #540

### DIFF
--- a/datalad_next/iter_collections/gitworktree.py
+++ b/datalad_next/iter_collections/gitworktree.py
@@ -13,7 +13,6 @@ from pathlib import (
     PurePath,
     PurePosixPath,
 )
-import re
 from typing import (
     Dict,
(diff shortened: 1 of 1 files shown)


christian-monch (approved):
LGTM

mih:
Thanks for the review!

PR #545, opened 2023-11-23
Remove default separator from `itemize()`

mih:
This addresses an issues brought up in
https://github.com/datalad/datalad-next/pull/539#discussion_r1403427421

This changeset removes the default argument to avoid the impression that "line-processing" is the main target. The code does not imply that, and the existing usage also not.

The possibility to do line-splitting is not touched (or removed). The documentation needs no adjustment.

### DIFF
--- a/datalad_next/itertools/itemize.py
+++ b/datalad_next/itertools/itemize.py
@@ -13,7 +13,8 @@ __all__ = ['itemize']
 
 def itemize(
     iterable: Iterable[bytes | str],
-    separator: str | bytes | None = None,
+    separator: str | bytes | None,
+    *,
     keep_ends: bool = False,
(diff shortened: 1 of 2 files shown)


christian-monch:
I think we are actually mostly line-based, even if the line terminator is often `'\0'`. For example, the `git ls-files` manpage says:
```
       -z
           \0 line termination on output and do not quote filenames. See OUTPUT below for more information.
```


But I also think it makes sense to let the user provide a separator, especially because not providing a separator will reduce performance. But we should probably highlight the fact that passing `None` as a separator will result in using all standard line splitting patterns that are built into python.

christian-monch (approved):

mih:
Thanks for the feedback!

PR #546, opened 2023-11-24
Use `iter_subproc` in `datalad_next.url_operations.ssh` instead of datalad-core runner code

christian-monch:
This PR uses `iter_subproc` instead of the datalad-core based `run` context-manager to implement `SshUrlOperations`.

The code changes are based on `iter-annexworktree-subproc`-branch (PR #539) and contain only one commit beyond the branch `iter-annexworktree-subproc`: 64e82d761e38f27d4aad73757c359570d45c2b19. This branch should be rebased on `main` once PR #539 is merged.

TODO:

- [x] Run a benchmark

### DIFF
--- a/datalad_next/iterable_subprocess/test_iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/test_iterable_subprocess.py
@@ -128,13 +128,12 @@ def test_exception_from_not_found_process_propagated():
 
 
 def test_exception_from_return_code():
-    with pytest.raises(IterableSubprocessError) as excinfo:
+    with pytest.raises(IterableSubprocessError, match='No such file or directory') as excinfo:
         with iterable_subprocess(['ls', 'does-not-exist'], ()) as output:
             a = b''.join(output)
 
     assert excinfo.value.returncode > 0
-    assert (b'No such file or directory' in excinfo.value.stderr) \
-           or (b'Datei oder Verzeichnis nicht gefunden' in excinfo.value.stderr)
+    assert b'No such file or directory' in excinfo.value.stderr
 
 
 def test_exception_from_context_even_though_return_code_with_long_standard_error():
--- a/datalad_next/iterable_subprocess/test_iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/test_iterable_subprocess.py
@@ -128,12 +128,13 @@ def test_exception_from_not_found_process_propagated():
 
 
 def test_exception_from_return_code():
-    with pytest.raises(IterableSubprocessError, match='No such file or directory') as excinfo:
+    with pytest.raises(IterableSubprocessError) as excinfo:
         with iterable_subprocess(['ls', 'does-not-exist'], ()) as output:
             a = b''.join(output)
 
     assert excinfo.value.returncode > 0
-    assert b'No such file or directory' in excinfo.value.stderr
+    assert (b'No such file or directory' in excinfo.value.stderr) \
+           or (b'Datei oder Verzeichnis nicht gefunden' in excinfo.value.stderr)
 
 
 def test_exception_from_context_even_though_return_code_with_long_standard_error():
--- a/datalad_next/itertools/align_pattern.py
+++ b/datalad_next/itertools/align_pattern.py
@@ -1,0 +1,94 @@
+""" Function to ensure that a pattern is completely contained in single chunks
+"""
+
+from __future__ import annotations
+
+from typing import (
+    Generator,
+    Iterable,
+)
+
+
+def align_pattern(iterable: Iterable[str | bytes | bytearray],
+                  pattern: str | bytes | bytearray
+                  ) -> Generator[str | bytes | bytearray, None, None]:
+    """ Yield data chunks that contain a complete pattern, if it is present
+
+    ``align_pattern`` makes it easy to find a pattern (``str``, ``bytes``,
+    or ``bytearray``) in data chunks. It joins data-chunks in such a way,
+    that a simple containment-check (e.g. ``pattern in chunk``) on the chunks
+    that ``align_pattern`` yields will suffice to determine whether the pattern
+    is present in the stream yielded by the underlying iterable or not.
+
+    To achieve this, ``align_pattern`` will join consecutive chunks to ensures
+    that the following two assertions hold:
+
+    1. Each chunk that is yielded by ``align_pattern`` has at least the length
+       of the pattern (unless the underlying iterable is exhausted before the
+       length of the pattern is reached).
+
+    2. The pattern is not split between two chunks, i.e. no chunk that is
+       yielded by ``align_pattern`` ends with a prefix of the pattern (unless
+       it is the last chunk that the underlying iterable yield).
+
+    The pattern might be present multiple times in a yielded data chunk.
+
+    Note: the ``pattern`` is compared verbatim to the content in the data
+    chunks, i.e. no parsing of the ``pattern`` is performed and no regular
+    expressions or wildcards are supported.
+
+    .. code-block:: python
+
+        >>> from datalad_next.itertools import align_pattern
+        >>> tuple(align_pattern([b'abcd', b'e', b'fghi'], pattern=b'def'))
+        (b'abcdefghi',)
+        >>> # The pattern can be present multiple times in a yielded chunk
+        >>> tuple(align_pattern([b'abcd', b'e', b'fdefghi'], pattern=b'def'))
+        (b'abcdefdefghi',)
+
+    Use this function if you want to locate a pattern in an input stream. It
+    allows to use a simple ``in``-check to determine whether the pattern is
+    present in the yielded result chunks.
+
+    Parameters
+    ----------
+    iterable: Iterable
+        An iterable that yields data chunks.
+    pattern: str | bytes | bytearray
+        The pattern that should be contained in the chunks. Its type must be
+        compatible to the type of the elements in ``iterable``.
+
+    Yields
+    -------
+    str | bytes | bytearray
+        data chunks that have at least the size of the pattern and do not end
+        with a prefix of the pattern. Note that a data chunk might contain the
+        pattern multiple times.
+    """
+
+    def ends_with_pattern_prefix(data: str | bytes | bytearray,
+                                 pattern: str | bytes | bytearray,
+                                 ) -> bool:
+        """ Check whether the chunk ends with a prefix of the pattern """
+        for index in range(len(pattern) - 1, 0, -1):
+            if data[-index:] == pattern[:index]:
+                return True
+        return False
+
+    # Join data chunks until they are sufficiently long to contain the pattern,
+    # i.e. have at least size: `len(pattern)`. Continue joining, if the chunk
+    # ends with a prefix of the pattern.
+    current_chunk = None
+    for data_chunk in iterable:
+        # get the type of current_chunk from the type of this data_chunk
+        if current_chunk is None:
+            current_chunk = data_chunk
+        else:
+            current_chunk += data_chunk
+        if len(current_chunk) >= len(pattern) \
+                and not ends_with_pattern_prefix(current_chunk, pattern):
+            yield current_chunk
+            current_chunk = None
+
+    if current_chunk is not None:
+        yield current_chunk
--- a/datalad_next/itertools/align_pattern.py
+++ b/datalad_next/itertools/align_pattern.py
@@ -50,6 +50,14 @@ def align_pattern(iterable: Iterable[str | bytes | bytearray],
     allows to use a simple ``in``-check to determine whether the pattern is
     present in the yielded result chunks.
 
+    The function always yields everything it has fetched from the underlying
+    iterable. So after a yield it does not cache any data from the underlying
+    iterable. That means, if the functionality of
+    ``align_pattern`` is no longer required, the underlying iterator can be
+    used, when ``align_pattern`` has yielded a data chunk.
+    This allows more efficient  processing of the data that remains in the
+    underlying iterable.
+
     Parameters
     ----------
     iterable: Iterable
--- a/datalad_next/url_operations/ssh.py
+++ b/datalad_next/url_operations/ssh.py
@@ -19,10 +19,9 @@ from typing import (
     Dict,
     Generator,
     IO,
+    cast,
 )
 from urllib.parse import urlparse
-
-from more_itertools import side_effect
 
 from datalad_next.consts import COPY_BUFSIZE
 from datalad_next.runners import (
@@ -105,7 +104,7 @@ class SshUrlOperations(UrlOperations):
             raise UrlOperationsResourceUnknown(url)
         return {k: v for k, v in props.items() if not k.startswith('_')}
 
-    def _get_props(self, url, stream: Generator) -> dict | None:
+    def _get_props(self, url, stream: Generator) -> dict:
         # Any stream must start with this magic marker, or we do not
         # recognize what is happening
         # after this marker, the server will send the size of the
@@ -119,7 +118,9 @@ class SshUrlOperations(UrlOperations):
 
         # Because the stream should start with the pattern, the first chunk of
         # the aligned stream must contain it.
-        chunk = next(aligned_stream)
+        # We know that the stream will deliver bytes, cast the result
+        # accordingly.
+        chunk = cast(bytes, next(aligned_stream))
         if chunk[:len(magic_marker)] != magic_marker:
             raise RuntimeError("Protocol error: report header not received")
         chunk = chunk[len(magic_marker):]
@@ -152,7 +153,7 @@ class SshUrlOperations(UrlOperations):
                  # obtain escalated/different privileges on a system
                  # to gain file access
                  credential: str | None = None,
-                 hash: str | None = None,
+                 hash: list[str] | None = None,
                  timeout: float | None = None) -> Dict:
         """Download a file by streaming it through an SSH connection.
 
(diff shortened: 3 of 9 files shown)


datalad_next/url_operations/ssh.py
```diff
@@ -16,23 +16,19 @@
...
-    Protocol as RunnerProtocol,
-    StdOutCaptureGeneratorProtocol,
-    ThreadedRunner,
-    CommandError,
-)
+from more_itertools import side_effect
+
+from datalad_next.consts import COPY_BUFSIZE
+from datalad_next.runners.iter_subproc import iter_subproc
+from datalad_next.iterable_subprocess.iterable_subprocess import IterableSubprocessError
```
mih, line 29: This is an indication that we should import `IterableSubprocessError` in `datalad_next.runners.iter_subproc`, such that client code can grab these pieces, that belong together, from the same spot.

mih:
I created a rebased version of this PR that sits on the recent state of https://github.com/datalad/datalad-next/pull/539

I would push that here, once that PR is merged.

datalad_next/itertools/align_pattern.py
```diff
@@ -0,0 +1,94 @@
...
+        >>> from datalad_next.itertools import align_pattern
+        >>> tuple(align_pattern([b'abcd', b'e', b'fghi'], pattern=b'def'))
+        (b'abcdefghi',)
+        >>> # The pattern can be present multiple times in a yielded chunk
+        >>> tuple(align_pattern([b'abcd', b'e', b'fdefghi'], pattern=b'def'))
+        (b'abcdefdefghi',)
+
+    Use this function if you want to locate a pattern in an input stream. It
+    allows to use a simple ``in``-check to determine whether the pattern is
+    present in the yielded result chunks.
```
mih, line 51: Thanks much for this updated description. Very clear now!

datalad_next/url_operations/ssh.py
```diff
@@ -43,7 +39,11 @@
 lgr = logging.getLogger('datalad.ext.next.ssh_url_operations')
 
 
-__all__ = ['SshUrlOperations']
+__all__ = [
+    'SshUrlOperations',
+    'UrlOperationsRemoteError',
+    'UrlOperationsResourceUnknown'
+]
```
mih, line 46: Not clear why this is needed

datalad_next/url_operations/ssh.py
```diff
@@ -81,63 +93,56 @@ def stat(self,
...
+
+        marker_index = chunk.index(b'\x01')
+        expected_size = int(chunk[:marker_index])
+        chunk = chunk[marker_index + 1:]
+        props = {
+            'content-length': expected_size,
+            '_stream':
+                chain([chunk], aligned_stream)
+                if chunk
+                else aligned_stream
```
mih, line 143: It there anything that would speak against the following:

```suggestion
                chain([chunk], aligned_stream)
                if chunk
                else aligned_stream
```

The `align_patter` iterator advanced the `stream` enough to get to the file content. The will be no other marker to look for. We should be able to detach and go back to the original iterable.

There is a comment a few lines above:

> # TODO: the align-pattern iterator stays "on" the stream, which might
> #  slow down processing upstream. We could consider to change
> #  align_pattern to an iterator-class and support detaching.

This is talking about the same issue, but suggests a different,
seemingly more sophisticated approach.

@christian-monch can you comment on the adequacy of my proposal?

datalad_next/url_operations/ssh.py
```diff
@@ -160,49 +165,55 @@ def download(self,
...
+                            len(chunk)
+                        ),
+                        download_stream,
+                        before=partial(
+                            self._progress_report_start,
+                            progress_id,
+                            ('Download %s to %s', from_url, to_path),
+                            'downloading',
+                            expected_size
+                        )
```
mih, line 202: Nice pattern!

It appears to me that we should add a `stream_progress` helper to the base class that provides this side-effect without needing the verbosity of the setup. We just need a `progress_id` the texts and an `expected_size` (if available).

This particular implementation here seems to be missing the end of the progress report.

mih (changes requested):
This looks nice! It is great to see that the new paradigm is smoothly applicable to this use case too!

I made a few comments. Some low hanging fruit, only one actual bug (progress end reporting).

datalad_next/url_operations/ssh.py
mih: This is done now.

datalad_next/url_operations/ssh.py
mih: I have undone it.

datalad_next/url_operations/ssh.py
mih: Tests remain happy.

mih:
The codecov patch report seems bogus. Here is what I get locally:

```
DATALAD_TESTS_SSH=1 python -m pytest --cov datalad_next.url_operations.ssh -s -v datalad_next/url_operations/tests/test_ssh.py
...
---------- coverage: platform linux, python 3.11.2-final-0 -----------
Name                                 Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------------
datalad_next/url_operations/ssh.py     106      7     34      4    89%   75-79, 100, 207, 315, 317->321
--------------------------------------------------------------------------------
TOTAL                                  106      7     34      4    89%
```

Not stellar, but not bad either.

mih:
Benchmark:

I downloaded a 230M file over an SSH connection. I see

- ~50+ seconds (4.5MB/s) for `scp`
- ~45 seconds (5.5MB/s) for `datalad download`

I checked that the md5sums for all downloads are the same, because it could not easily believe it.

Given only N=4, I would not call it "faster", but certainly also not slower. Well done!

datalad_next/url_operations/ssh.py
```diff
@@ -160,50 +162,59 @@ def download(self,
...
+                    hasher.update(chunk)
+
+        except IterableSubprocessError as e:
+            self._check_return_code(e.returncode, from_url)
+        except StopIteration:
+            raise UrlOperationsResourceUnknown(from_url)
         finally:
-            if dst_fp and to_path is not None:
-                dst_fp.close()
             self._progress_report_stop(progress_id, ('Finished download',))
```
mih, line 211: How confident are you that it is OK to skip closing `dst_fp`?

It would eventually be closed on garbage collection, I guess. But the timing of that is relatively unpredictable. What made you remove it?

datalad_next/url_operations/ssh.py
christian-monch: Thanka

datalad_next/url_operations/ssh.py
christian-monch: Thanks

datalad_next/url_operations/ssh.py
christian-monch: Moved progress report end into the `after`-function of  `side_effect` (which is executed in a `finally`-branch, so it will also run in case of errors). (Done in 06f3ed56805fd9d537eea0b681a39b2f6608937f)

christian-monch: > ...
> It appears to me that we should add a `stream_progress` helper to the base class that provides this side-effect without needing the verbosity of the setup. We just need a `progress_id` the texts and an `expected_size` (if available).
> ...

Nice idea. Done in 585525007caf0fb70b8336af9d2f0d4802e62624

datalad_next/iterable_subprocess/test_iterable_subprocess.py
```diff
@@ -128,12 +128,13 @@ def test_exception_from_not_found_process_propagated():
...
 def test_exception_from_return_code():
-    with pytest.raises(IterableSubprocessError, match='No such file or directory') as excinfo:
+    with pytest.raises(IterableSubprocessError) as excinfo:
         with iterable_subprocess(['ls', 'does-not-exist'], ()) as output:
             a = b''.join(output)
 
     assert excinfo.value.returncode > 0
-    assert b'No such file or directory' in excinfo.value.stderr
+    assert (b'No such file or directory' in excinfo.value.stderr) \
+           or (b'Datei oder Verzeichnis nicht gefunden' in excinfo.value.stderr)
```
mih, line 137: Let's not do this. Instead, patch the ENV to set `LANG=C`. This is a safe pattern. The solution above is waiting for 195 other locales to be added.

In general, this is hinting at a possible extension of `iter_subproc()` (not for this test, though), to accept ENV overrides. Parsing of program output is always fragile, and should only happen as a last resort. We would postponed such an addition to a later point in time.

datalad_next/url_operations/ssh.py
```diff
@@ -81,63 +90,63 @@ def stat(self,
...
+            with iter_subproc(cmd) as stream:
+                # any exception that is raised in this context and not caught
+                # will prevent the creation of `IterableSubprocessError`. But
+                # we rely on the return code of the ssh-process to signal
+                # specific errors. Therefore, we catch the expected
+                # `StopIteration` here.
+                try:
+                    props = self._get_props(url, stream)
+                except StopIteration:
+                    pass
```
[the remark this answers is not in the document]
christian-monch: An example without an exception in the `iter_subproc`-context:

```python
>>> from datalad_next.runners.iter_subproc import iter_subproc
>>> with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
...     tuple(sub)
... 
()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/home/cristian/Develop/datalad-next/datalad_next/iterable_subprocess/iterable_subprocess.py", line 153, in iterable_subprocess
    raise IterableSubprocessError(proc.returncode, b''.join(stderr_deque)[-chunk_size:])
datalad_next.iterable_subprocess.iterable_subprocess.IterableSubprocessError: (2, b"ls: Zugriff auf '/THIS_DOES_NOT_EXIST' nicht m\xc3\xb6glich: Datei oder Verzeichnis nicht gefunden\n")
```

This raises `IterableSubprocessError`.

The following example raises `StopIteration`:

```python
>>> from datalad_next.runners.iter_subproc import iter_subproc
>>> with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
...     next(sub)
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
StopIteration
```

Because we want to get the return code of our subprocesses,  we need to catch `StopIteration` (in this case it is `StopIteration`, but generally all exceptions that are raised in the `iter_subproc`-context need to be caught to get to an `IterableSubprocessError`).

We could of course write code that does not raise `StopIteration`, but that requires to pass error values around, which is not too nice either.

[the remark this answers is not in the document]
mih: But how do the examples relate to the usage pattern in this PR? The iteration is done within the for-loop. How would `StopIteration` get out? My last example raises inside the for-loop, and we see `RuntimeError`, the actual exception.

If I understand you correctly, `IterableSubprocessError` would never come out, unless all other exceptions are caught. But I do not see that (and it would be a bit weird):

```py
In [1]: from datalad_next.runners import iter_subproc
In [2]: with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
   ...:     for i in sub:
   ...:         print('some')
   ...: 
---------------------------------------------------------------------------
IterableSubprocessError                   Traceback (most recent call last)
Cell In[2], line 1
----> 1 with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
      2     for i in sub:
      3         print('some')

File /usr/lib/python3.11/contextlib.py:144, in _GeneratorContextManager.__exit__(self, typ, value, traceback)
    142 if typ is None:
    143     try:
--> 144         next(self.gen)
    145     except StopIteration:
    146         return False

File ~/hacking/datalad/next/datalad_next/iterable_subprocess/iterable_subprocess.py:153, in iterable_subprocess(program, input_chunks, chunk_size)
    150         raise e.__context__ from None
    152 if proc.returncode:
--> 153     raise IterableSubprocessError(proc.returncode, b''.join(stderr_deque)[-chunk_size:])

IterableSubprocessError: (2, b"ls: cannot access '/THIS_DOES_NOT_EXIST': No such file or directory\n")
```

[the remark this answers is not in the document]
christian-monch: > But how do the examples relate to the usage pattern in this PR? The iteration is done within the for-loop. How would `StopIteration` get out? My last example raises inside the for-loop, and we see `RuntimeError`, the actual exception.
> [...]

We also use `next()` in `self._get_props()`. We can use a for-loop for those as well, but then we have to pass an error-result, e.g. `None` to the caller and handle that in the the caller.

> If I understand you correctly, `IterableSubprocessError` would never come out, unless all other exceptions are caught. [,,,]

Yes! `IterableSubprocessError` will never come out, if any exception is raised in the `iter_subproc`-context (there is one exception from that rule, the internal `_BrokenPipeError`). E.g:

```python
>>> from datalad_next.runners.iter_subproc import iter_subproc
>>> with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
...     raise(ValueError)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError
```
So we have to prevent exception-raising, if we want to get access to the return-code of the subprocess.
 

> [...] But I do not see that (and it would be a bit weird):
> 
> ```python
> In [1]: from datalad_next.runners import iter_subproc
> In [2]: with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
>    ...:     for i in sub:
>    ...:         print('some')
>    ...: 
> ---------------------------------------------------------------------------
> IterableSubprocessError                   Traceback (most recent call last)
> Cell In[2], line 1
> ----> 1 with iter_subproc(['ls', '/THIS_DOES_NOT_EXIST']) as sub:
>       2     for i in sub:
>       3         print('some')
> 
> File /usr/lib/python3.11/contextlib.py:144, in _GeneratorContextManager.__exit__(self, typ, value, traceback)
>     142 if typ is None:
>     143     try:
> --> 144         next(self.gen)
>     145     except StopIteration:
>     146         return False
> 
> File ~/hacking/datalad/next/datalad_next/iterable_subprocess/iterable_subprocess.py:153, in iterable_subprocess(program, input_chunks, chunk_size)
>     150         raise e.__context__ from None
>     152 if proc.returncode:
> --> 153     raise IterableSubprocessError(proc.returncode, b''.join(stderr_deque)[-chunk_size:])
> 
> IterableSubprocessError: (2, b"ls: cannot access '/THIS_DOES_NOT_EXIST': No such file or directory\n")
> ```
> 
> Take a look at the traceback ^^^

IIANM, the traceback does not involve an exception raised in the context (the value of `typ` in `_GeneratorContextManager.__exit__` is `None`), therefore lines 152 - 153 are executed.

Any exception in the `with iter_subproc`-context would trigger execution of lines 138 - 139 of `iterable_subprocess.py`:
```python
   137              except BaseException:
   138                  proc.terminate()
   139                  raise
```
And lines 152 - 153 are never executed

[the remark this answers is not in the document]
mih: OK, I think I am slowly getting what you are saying. I believe the key is:

> We also use next() in self._get_props().

But to me all the wrapping in try/except seems to be *solely*, because that is the case, and not "also".

If that is true, it would seem to be more sensible to prevent `_get_props()` from ever raising `StopIteration`. It is an internal function that always needs this protection, and that need is neither obvious, nor documented.

But that seems to require a rather complex logic, because, again if I understand this right, we would want to leave the `iter_subproc` context immediately, and have an outer shell catch the then generated `IterableSubprocessError` -- which is only generated, if there is no exception already.

If that is all correct, we need to discuss two things:

- this should be documented in `iter_subproc()` -- it is all but obvious
- can this be simplified (avoid having to wrap any and all context body like this) by also handling `StopIteration` where `_BrokenPipeError` is handled in `iterable_subprocess()`? Basically:

```py

    except _BrokenPipeError as e:
        if proc.returncode == 0:
            raise e.__context__ from None
    # add this ----------------
    except StopIteration:
       # something try to read output, but there is none
       # if that is a problem the next check will tell
    # until here ---------------

    if proc.returncode:
        raise IterableSubprocessError(proc.returncode, b''.join(stderr_deque)[-chunk_size:])
```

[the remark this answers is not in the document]
mih: I think that is a good approach (dedicated PR first, and this one waiting).

I will keep trying to think of a reason why we would not want to change `iterable_subprocesses`. Obviously, fewer changes are better, but this issues seems to be impossible to fix from the outside.

Conceptually, `_BrokenPipeError` and `StopIteration` appear to be to variants of the same problem: premature death. It continues to appear sensible to me to also deal with them in the same way.

You patch is a minimal as it can get. What I cannot anticipate are sideeffect of raising the `StopIteration.__context__`. Never seen that being done.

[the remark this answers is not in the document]
christian-monch: > [...] You[r] patch is a minimal as it can get. What I cannot anticipate are sideeffect of raising the StopIteration.__context__. Never seen that being done.

That is a good point because we should not do that. So the patch will have to be a little more intrusive, something like:
```
    except _BrokenPipeError as e:
        if proc.returncode == 0:
            raise e.__context__ from None
    except StopIteration:
        if proc.returncode == 0:
            raise
```

> [...]
> Conceptually, `_BrokenPipeError` and `StopIteration` appear to be to variants of the same problem: premature death. It continues to appear sensible to me to also deal with them in the same way.
> [...]

That is a nice thought. I was struggeling to find sound reasoning for the patch.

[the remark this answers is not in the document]
christian-monch: > [...] 
> Preparing the PR now.

I came across another idea on how to make return codes accessible in the presence of exceptions: PR #565.

(I had this idea after opening PR #562, which implements what is outlined above. So I put PR #562 into draft-mode)

mih:
Yhus PR is replaced by #566

PR #548, opened 2023-11-27
Improve error handling, documentation and tests of `itertools`

mih:
Extracted and modified from https://github.com/datalad/datalad-next/pull/539


datalad_next/itertools/itemize.py
```diff
@@ -13,64 +13,92 @@
...
-        determined by the line-separators that are built into `splitlines`.
+        determined by the line-separators that are built into
+        ``str.splitlines()``.
     keep_ends: bool
-        If `True`, the item-separator will be present at the end of a
-        yielded item line. If `False`, items will not contain the
-        separator. Preserving separators an additional implies a runtime cost.
+        If `True`, the item-separator will remain at the end of a
+        yielded item. If `False`, items will not contain the
+        separator. Preserving separators implies a runtime cost.
```
christian-monch, line 63:
```suggestion
        separator. Preserving separators implies a runtime cost, unless the separator is ``None``.
```

christian-monch (approved):
Thanks for including the documentation here.

I just have a few remarks regarding the runtime of `itemize` when `keep_ends=True` is specified.  The runtime is only depending on the value of `keep_ends` if `sep` is not `None`.

PR #552, opened 2023-12-05
Add a mode and test for non-recursive `iter_gitworktree()`, plus fixes

### LINKED ISSUE #463: `iter_gitworktree` docs misleading
The docs for the `link_target` and `fp` parameters are identical. This is correct, because of #462, but this is neither intentional not desirable.

A test must be added and the docs rectified.

### LINKED ISSUE #468: `iter_gitworktree` stumbles over dead symlink
Looking into #463 it seems that the implementation is not able to work with dangling symlinks (e.g., absent annex objects).

```
---> 79     cstat = path.lstat()
     80     cmode = cstat.st_mode
     81     if stat.S_ISLNK(cmode):

AttributeError: 'PurePosixPath' object has no attribute 'lstat'
```

### LINKED ISSUE #551: Robustify `iter_gitworktree()`
It wants a `Path` instance as the main argument. It also works with a plain path string (although that is undocumented), as long as `fp=False`. With `fp=True` it crashes:

```py
File ~/hacking/datalad/next/datalad_next/iter_collections/utils.py:79, in FileSystemItem.from_path(cls, path, link_target)
     66 @classmethod
     67 def from_path(
     68     cls,
   (...)
     71     link_target: bool = True,
     72 ):
     73     """Populate item properties from a single `stat` and `readlink` call
     74 
     75     The given ``path`` must exist. The ``link_target`` flag indicates
     76     whether to report the result of ``readlink`` for a symlink-type
     77     path.
     78     """
---> 79     cstat = path.lstat()
     80     cmode = cstat.st_mode
     81     if stat.S_ISLNK(cmode):

AttributeError: 'PurePosixPath' object has no attribute 'lstat'
```

I think that is not good. Having an initial conversion of str to path at the beginning would avoid this, somewhat hidden, UX problem.

mih:
This adds support for a standard query that Gooey would be making. Therefore, this is a significant step towards a resolution of https://github.com/datalad/datalad-next/issues/323.

When interfaced with https://github.com/datalad/datalad-next/pull/539, it replaces `gooey-lsdir` and half of `gooey-status-light`, with a much more efficient/convenient implementation.

Also:

- Closes #463 
- Closes #468 
- Closes #551

### DIFF
--- a/datalad_next/iter_collections/gitworktree.py
+++ b/datalad_next/iter_collections/gitworktree.py
@@ -127,14 +127,18 @@ def iter_gitworktree(
       ``all`` reports on any untracked file; ``whole-dir`` yields a single
       report for a directory that is entirely untracked, and not individual
       untracked files in it; ``no-empty-dir`` skips any reports on
-      untracked empty directories. Any untracked content is yielded as
-      a ``PurePosixPath``.
+      untracked empty directories.
     link_target: bool, optional
-      If ``True``, each file-type item includes a file-like object
-      to access the file's content. This file handle will be closed
-      automatically when the next item is yielded.
+      If ``True``, information matching a
+      :class:`~datalad_next.iter_collections.utils.FileSystemItem`
+      will be included for each yielded item, and the targets of
+      any symlinks will be reported, too.
     fp: bool, optional
-      If ``True``, each file-type item includes a file-like object
+      If ``True``, information matching a
+      :class:`~datalad_next.iter_collections.utils.FileSystemItem`
+      will be included for each yielded item, but without a
+      link target detection, unless ``link_target`` is given.
+      Moreover, each file-type item includes a file-like object
       to access the file's content. This file handle will be closed
       automatically when the next item is yielded.
     recursive: {'repository', 'no'}, optional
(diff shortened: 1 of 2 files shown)


mih:
OK, let's go.

datalad_next/iter_collections/gitworktree.py
```diff
@@ -93,6 +93,7 @@ def iter_gitworktree(
     untracked: str | None = 'all',
     link_target: bool = False,
     fp: bool = False,
+    recursive: str = 'repository',
```
christian-monch, line 96: What is the rationale for using the string `'repository'`? It is not validated, anything apart from `'no'` would lead to identical behavior. So it seems more like a `'no'` and not-`'no'` value. Wouldn't a `bool` fit better? Or are we expecting additional recursion behaviors? In the latter case, we might want to check for known behaviors, i.e. for `'repository'`

christian-monch (commented):
Looks good, just one remark

datalad_next/iter_collections/gitworktree.py
mih: Yes, there will be subdataset recursion. The corresponding test already has a hint at that.

PR #562, opened 2023-12-08
Handle `StopIteration` in `iterable_subprocess`

christian-monch:
[This has been put into draft-mode in favor of an alternative approach to make return codes available in case of exceptions in PR #565]

This PR implements the result of this [discussion](https://github.com/datalad/datalad-next/pull/546#discussion_r1418976060) in PR #546

It modifies `iterable_subprocess` and therefore `iter_subproc` to prioritize `IterableSubprocessError`-exceptions over `StopIteration`-exception. The rationale for that is, that `StopIteration` and a non-zero exit value of a subprocess indicate that there was a short-read due to an error in the subprocess.

Because we require access to the return code of the subprocess, prioritizing `IterableSubprocessError`-exceptions over `StopIteration`-exceptions alleviates the need to catch `StopIteration` in the `iter_subproc`-context.

### DIFF
--- a/datalad_next/iterable_subprocess/iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/iterable_subprocess.py
@@ -148,6 +148,17 @@ def iterable_subprocess(program, input_chunks, chunk_size=65536):
     except _BrokenPipeError as e:
         if proc.returncode == 0:
             raise e.__context__ from None
+    except StopIteration:
+        # This except-clause prioritizes `IterableSubprocessError` over
+        # `StopIteration` to avoid swallowing the former.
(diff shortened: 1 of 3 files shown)


christian-monch:
I am wondering whether it would be better to change `iterable_subprocess` into a class that implements the context-manager interface?

The instances of this class could then store the return code of the subprocess and it would still be available to us, independent from the exceptions that were raised.

christian-monch:
Ok, I will create another PR that turns the `as`-variable into a generator that can hold the return ode

mih:
With #565 merged, we can close this one.

PR #565, opened 2023-12-10
Use `OutputFrom` class as `iter_subproc` result, which keeps returncodes on exception

christian-monch:
This PR implements a solution for the problem of not-easily available return codes that was discussed [here](https://github.com/datalad/datalad-next/pull/546#discussion_r1418976060) in PR #546.

This is an alternative (and in my view superior) approach to the discussed solution, which is implemented in PR #562.

The reason why this second approach exists is:

1. It seems much more pythonic to me
2. It is more general and does not introduce special handling of different exceptions (in this case, the handling of `StopIteration`in PR #562)
3. It makes for cleaner code when using `iter_subproc`, i.e. in `SshUrlOperations`.

### DIFF
--- a/datalad_next/iterable_subprocess/iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/iterable_subprocess.py
@@ -1,4 +1,5 @@
 from collections import deque
+from collections.abc import Generator
 from contextlib import contextmanager
 from subprocess import PIPE, SubprocessError, Popen
 from threading import Thread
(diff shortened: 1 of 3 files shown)


christian-monch:
The change that this PR provides are applied in `SshUrlOperations` in the PR #566 (in commit e2937d810de701f8469cfa4e5c3098fa9fb826f4)

mih (approved):
This works for me. It feels much better than the alternative handling, indeed. Thx!

PR #566, opened 2023-12-11
Use `iter_subproc` instead of `ThreadedRunner` in `SshUrlOperations`

christian-monch:
This PR uses `iter_subproc` instead of `ThreadedRunner` from datalad-core to implement `SshUrlOperations`. It also introduces the `align_pattern` iterator, which is used by `SshUrlOperations` to reliably detect marker in the output stream of a ssh-shell.


christian-monch:
Merged main again to fetch code from PR #570

datalad_next/url_operations/ssh.py
```diff
@@ -165,49 +169,56 @@ def download(self,
...
+                    # write data
+                    dst_fp_write(chunk)
+                    # compute hash simultaneously
+                    hasher.update(chunk)
+        except IterableSubprocessError:
+            self._check_return_code(stream.returncode, from_url)
+        except StopIteration:
+            # an unexpected `StopIteration` occurred, this indicates a
+            # non-existing resource.
+            self._check_return_code(244, from_url)
```
mih, line 213: Why not use the actual exit code, rather than declaring one that may or may not be correct?

christian-monch: If an exception is raised and the `iter_subproc`-content is left, the subprocess is terminated. A terminated `ssh` returns 255 as exit code. This code does not lead to an `UrlOperationsResourceUnknown`-exception, and would therefore not trigger the correct code-path, i.e. the "resource-not-available" code-path.

The alternative (and we can discuss whether that would be more correct) approach would be to catch `StopIteration`-exceptions (or ---if we do use `for` instead of `next()`--- detect short iterator results) and return an error indicator instead of the `props`-dictionary, e.g. `None`. That would then lead to an `IterableSubprocessError` with return code 244 (in case of non-existing files).

mih: OK, thanks for the explanation.

mih:
Merged as #573

PR #570, opened 2023-12-13
SSH implementation and test upgrade

### LINKED ISSUE #563: Enable SSH testing on appveyor
https://github.com/datalad/datalad-ria/blob/main/.appveyor.yml shows how a container-based SSH server can be deployed, and the tests in that repository show how it can be used.

The containers are provided by https://github.com/datalad/datalad-ci-docker-containers

### LINKED ISSUE #571: `SshUrlOperations` cannot handle non-default port/user/id
As discovered in #570, the `_SshCat` helper is only considering the HOST when compiling an `ssh` command call.

This is a blocker for resolving #563, which relies on more complex setups, and is also generally limiting the applicability of this tooling.

mih:
Closes #563
Closes #571 

TODO:

- [x] The current SSHUrlOperation implementation neither deals with users nor ports in URL. Given `ssh://sshuser@datalad-test-sshd:2222//usr/local/apache2/htdocs/myfile` it runs `ssh datalad-test-sshd`. The lack of support comes from `_SshCat`.


christian-monch (approved):
Assuming that the type errors would be fixed in  #566, this looks good to me.

mih:
Thanks!

PR #576, opened 2023-12-16
Enhance `iter_subproc` with a `cwd` parameter

mih:
Running in a particular working directory is a very common paradigm (think in a Git repository). So far, we used Git's `-C` option to perform the directory changes. This is fine, fast, and works universally.
    
However, recent discussion asked whether it would make sense to call annex commands via `git-annex`, rather than `git annex`. The former does not provide `-C`. This change readies the code for such a move.
    
Another use case for this is the passing of (many) file paths. With an option to run in a particular directory, the file names can be just that (not full paths). This can help to mitigate situations, where many file paths potentially exhaust the max command length.

### DIFF
--- a/datalad_next/runners/iter_subproc.py
+++ b/datalad_next/runners/iter_subproc.py
@@ -1,4 +1,6 @@
 from __future__ import annotations
+
+from pathlib import Path
 from typing import (
     Iterable,
     List,
@@ -20,6 +22,7 @@ def iter_subproc(
     *,
     input: Iterable[bytes] | None = None,
     chunk_size: int = COPY_BUFSIZE,
+    cwd: Path | None = None,
 ):
     """Context manager to communicate with a subprocess using iterables
 
@@ -87,6 +90,8 @@ def iter_subproc(
       subprocess's ``stdin``.
     chunk_size: int, optional
       Size of chunks to read from the subprocess's stdout/stderr in bytes.
+    cwd: Path
+      Working directory for the subprocess, passed to ``subprocess.Popen``.
 
     Returns
     -------
@@ -96,4 +101,5 @@ def iter_subproc(
         args,
         tuple() if input is None else input,
         chunk_size=chunk_size,
+        cwd=cwd,
     )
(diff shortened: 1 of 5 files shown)


datalad_next/runners/iter_subproc.py
```diff
@@ -20,6 +22,7 @@ def iter_subproc(
     *,
     input: Iterable[bytes] | None = None,
     chunk_size: int = COPY_BUFSIZE,
+    cwd: Path | None = None,
```
christian-monch, line 25: Since `Popen` supports `Path`, `bytes`, and `str` from Python 3.8 on Posix and Windows, let's put them in here.
```suggestion
    cwd: Path | bytes | str | None = None,
```

datalad_next/runners/iter_subproc.py
```diff
@@ -87,6 +90,8 @@ def iter_subproc(
       subprocess's ``stdin``.
     chunk_size: int, optional
       Size of chunks to read from the subprocess's stdout/stderr in bytes.
+    cwd: Path
+      Working directory for the subprocess, passed to ``subprocess.Popen``.
```
christian-monch, line 94:
```suggestion
    cwd: Path | bytes  | str,  optional
      If given, working directory for the subprocess, passed to ``subprocess.Popen``.
```

christian-monch (approved):
LGTM, I only had some type-related nitpicking.

mih:
Thanks for the weekend review! Much appreciated!

My ratio al for not putting it all types there was (and is) that I would really like to discourage any use of non `Path` paths representations. The misinterpretation of paths (in particular across platforms) is one of the great pains of the datalad dev history. So you are technically correct, of course. But I am thinking of it as a disciplinary instrument.

christian-monch:
> Thanks for the weekend review! Much appreciated!
> 
> My rational for not putting all types there was (and is) that I would really like to discourage any use of non `Path` path representations. The misinterpretation of paths (in particular across platforms) is one of the great pains of the datalad dev history. So you are technically correct, of course. But I am thinking of it as a disciplinary instrument.

That appeals to my authoritarian side. Jawoll! ;-)

Issue #349, opened 2023-05-10
Implement `itergittree()` for `ls-file-collection`


mih:
At the command level this should be exposed in multiple ways that also enable the specification of a commit or branch.

It would then become a standard helper wherever `lstree` is used right now.

This will require an extension of the command API to include a standard `dataset` parameter.

Ping #343

Issue #541, opened 2023-11-23
Discontinue (direct) usage of any runner implementations


mih:
With #538 we made the start of a new paradigm: iterator-based interaction with (concurrent) subprocesses.

It is our (@mih and @christian-monch) understanding that this basic approach would work for most (if not all) use cases that involve the execution of subprocesses. Respective experiments have been made and the results look promising. The "ultimate" challenge will be the implementation of a queue-based remote shell.

In order to bring down the complexity of implementations after the introduction of this new paradigm, all subprocess execution in datalad-next should be switch to it.

Afterwards all runner-related imports from datalad-core should be discontinued (which involves updating any extensions that make use of it).

This plan answers the questions of #519. `GitRunner` has no future.

mih:
Remaining usage (already discounting #546):

- `datalad_next/constraints/git.py`
- `datalad_next/gitremotes/datalad_annex.py`
- `datalad_next/patches/annexrepo.py`

mih:
`datalad_next/constraints/git.py` has been replaced in #592

`datalad_next/patches/annexrepo.py` does not actually use any runner functionality provided through datalad-next.

Only `datalad_next/gitremotes/datalad_annex.py` is TODO.

PR #580, opened 2024-01-04
Implement `iter_gittree()` and support it in `ls_file_collections`

mih:
This is the first iterator that is not simply reporting on some file system that is identifiable via a single location parameter.  Instead, it requires a repository location and a second "tree-ish" identifier.

I decided to skip the implementation of a file-pointer inclusion into the returned items. Two reasons: (1) to make this efficient it would need a lazy-execution of `git cat-file`, and (2) I cannot think of a use case for content hashing beyond identification, and that use case is perfectly addressed by the reported `gitsha` already.

For the integration with `ls-file-collection` I decided to stay close to the behavior of `git ls-tree` itself. The collection identifier parameter is used for the tree-ish, while the repository selection is done via the working directory. This includes constraining reports to subdirectories (in non-bare repos), like `git ls-tree` does too.

Closes: #349

TODO:

- [x] Add tests

### DIFF
--- a/datalad_next/iter_collections/tests/test_itergittree.py
+++ b/datalad_next/iter_collections/tests/test_itergittree.py
@@ -15,6 +15,7 @@ from ..gittree import (
 
 def test_iter_gittree(existing_dataset, no_result_rendering):
     ds = existing_dataset
+    is_crippled_fs = ds.repo.is_crippled_fs()
 
     tracked_items = list(iter_gittree(ds.pathobj, 'HEAD'))
     # without untracked's and no link resolution this is plain and fast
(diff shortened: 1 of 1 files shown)


mih:
Some performance figures (again a Git repo with 36k tracked files):

```
❯ multitime -q -n 10 git ls-tree -r HEAD
===> multitime results
1: -q git ls-tree -r HEAD
real        0.015       0.002       0.013       0.014       0.022       
user        0.013       0.002       0.009       0.014       0.014       
sys         0.002       0.003       0.000       0.000       0.010       
```

```
In [1]: from datalad_next.iter_collections.gittree import iter_gittree
In [2]: %timeit list(iter_gittree('.', 'HEAD', recursive='repository'))
104 ms ± 9.7 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

~6x slower than the plain Git command at this point.

mih:
Given the underwhelming performance I did some profiling and found that half the runtime it spent on converting path strings to `PurePath` instances. That seems a bit out of proportion...

mih:
I think this is good now. I will merge it shortly in order to then address #581 and #541 jointly for all collection types.

PR #583, opened 2024-01-05
Change semantics of `PathBasedItem` names

### LINKED ISSUE #554: Double check and document (if needed) `.name` conventions for `iter_collections`
I cannot put my finger on it, but I feel the nature of the value of `.name` for a reported item is non-uniform.

I am not aware of a documentation for the convention, if how iterators should behave.

The `.name` should always be a `PurePath|Path` objects.

However, should it carry the base path (key argument to the iterator) in it? Or not?

It would make sense to me to never have it. Rational:

There is no reason while a (file) collection should be on a file system. It could have a very non-Path base address and no hierarchical organization inside (think object store: some URL, plus some object ID).

We need to go through and verify that we did not overfit to a filesystem use case in an implementation.

### LINKED ISSUE #581: Performance impact of `PathBasedItem`
This dataclass is used in pretty much all iterators of `iter_collections`. In https://github.com/datalad/datalad-next/pull/580 I saw that the conversion of path strings to `PurePath` instances can represent as much as half of the runtime of such an iterator.

It seems worthwhile to consider optimizations in this regard.

One possibility would be to change `PathBasedItem` to leave its `name` property untouched (`Any`), and add an accessor property (`@cached_property`) `path` that performs the conversion lazily. With this approach, any conversion performance cost would only need to be paid when actually necessary. Moreover, consumer would be given the choice of format of an item `name`.

mih:
Previously, the `name` attribute was some kind of `PurePath` instance. The precise nature varied across item types and corresponding iterators.

This led to two problems:

- Performance (see #581): the creation of a Path instance is not particularly cheap. If a consumer does not need such an instance, and unconditional creation is a waste of resources
- Abstraction (see #430): the Path representation of a path-like name does not always match the exact semantics of the original name (it might have a trailing slash that has a purpose, etc)

Both problems are now addressed by relaxing the type of `.name` of a `PathBasedItem`. It can now be anything. Consumers that require an actual Path instance can use the `.path` attribute. An analog access is also implemented for `.link_target` (which now remains literal), that is accompanied by `.link_target_path`, where it was necessary.

This change in approach removed the need for the `_ZipFileDirPath` workaround that was used to re-establish compatibility of a Path-based `.name` with the conventions in a `ZipFile` collection/container. With the `.name` attribute retaining its native semantics, the workaround is no longer needed as was removed.

In order to document the (lack of) implemented homogeneous conventions for path-based items, the docstrings of all iterators have been amended with information on the nature of the `.name` attribute yielded by them. The corresponding data classes have received docstrings for the newly added properties for Path access.

These properties uniformly use the `cached_property` decorator. For the lifetime of an item, the nature of such a path should never change, and caching it automatically addresses the significant creation cost on accessing a path representation repeatedly.

Closes #554
Closes #581

### DIFF
--- a/datalad_next/commands/ls_file_collection.py
+++ b/datalad_next/commands/ls_file_collection.py
@@ -57,7 +57,6 @@ from datalad_next.iter_collections.utils import (
 )
 from datalad_next.iter_collections.gittree import (
     GitTreeItemType,
-    GitTreeItem,
     iter_gittree,
 )
 from datalad_next.iter_collections.gitworktree import (
--- a/datalad_next/commands/tests/test_ls_file_collection.py
+++ b/datalad_next/commands/tests/test_ls_file_collection.py
@@ -40,9 +40,9 @@ def _check_archive_member_result(r, collection):
     # a collection identifier, here the tar location
     assert 'collection' in r
     assert r['collection'] == collection
-    # an item identifier, here a path of an archive member
+    # an item identifier, here a str-path of an archive member
     assert 'item' in r
-    assert isinstance(r['item'], PurePath)
+    assert isinstance(r['item'], str)
     # item type info, here some filesystem-related category
     assert 'type' in r
     assert r['type'] in ('file', 'directory', 'symlink', 'hardlink')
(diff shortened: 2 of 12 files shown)


mih:
Ok, let's go without review

Issue #587, opened 2024-01-08
Add `GitPathSpec` support class and constraint


mih:
Looking into #586 it seems attractive to start support Git's pathspecs properly https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec

The main development challenge would be support for mangling pathspecs for subdataset recursions implemented outside Git. Something like stripping away a path prefix when a pathspec is passed on to another call on a subdataset -- or detecting that a particular pathspec would no longer apply to a subdataset.

The specification suggests great flexibility. The key statement/property seems to be

> the pathspec up to the last slash represents a directory prefix. The scope of that pathspec is limited to that subtree.

This would imply that we need to implement pathspec parsing enough to be able to extract and manipulate that directory prefix.

Such a manipulation might be as simple as removing segments of the `/` delimited directory prefix or `glob`.

For operation on Windows we might also want to convert platform paths to POSIX paths for simplicity. But I have not yet tested how Git behaves there.

mih:
Turns out that the mangling of path specs for submodules (subdirectories) is not trivial. Supporting all possible combinations of magics is laborious, but seems doable. The main challenge is the deal with wildcards. When translating a pathspec for a subdirectory we do have a full path to match against (only the subdirectory). I cannot come up with a rule when and how to decide where to strip the longest or the shortest match from the pathspec. The best I can come up with is to multiply the pathspecs whenever they contain `*`, and use shortest *and* longest match for a subdirectory.

mih:
#592 provides code and a test environment to continue the development of #588.

PR #588, opened 2024-01-09
Add support subpackage for Git's pathspecs

mih:
The main (if not only) purpose of this functionality is pathspec mangling/translation for handing them over to analog Git command calls on submodules -- for any Git command that supports pathspecs, but not recursion.

A simple example for such a command is `git ls-files --other`. It accepts pathspecs, but does not implement `--recurse-submodules` for listing untracked files.

The goal of this functionality is to be able to take pathspecs that is valid in the context of a top-level repository, and translate it such that the set of paths specs given to the same command running on/in a submodule/subdirectory gives the same results, as if the initial top-level invocation reported them (if it even could).

The included sketch of a testbattery uses ``git ls-files --other` for testing, rather than a formal description -- because the behavior of the implementation is more elaborate than the documentation at https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec suggests.

All testing is (for now) performed within a single repository, and with translation for execution in subdirectories.

The implementation is a rough sketch for exploring the problem, rather than anything polished.

Ping #587

Also see https://github.com/datalad/datalad/issues/6933

TODO:

- [x] ready tests for case-insensitive file systems
- [ ] support `:(glob)` pathspecs via `PurePath.match()`
- [ ] test `:(exclude)` pathspec handling
- [ ] test `:(attr:...)` pathspec handling
- [x] have another go at figuring out why the `icase` magic does not work as documented
- [ ] add optimization to bypass expensive translations when a pathspec is just a path with no wildcards
- [ ] Add constraint to validate pathspecs
- [ ] Add pathspec support to `iter_submodules()`. Alternatively, add pathspec support to `iter_gitworktree()`. However, the former could be less complex. We could take any reported submodule, and try to port any given pathspec to it, and report only submodule that has any match. In `iter_gitworktree()` we'd have the problem that `git ls-files` is totally silent if given a pathspec that points (exclusively) into a submodule.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -53,7 +53,7 @@ build: off
 
 environment:
   # unless indicated otherwise, we test datalad_next
-  DTS: datalad_next
+  DTS: datalad_next.gitpathspecs datalad_next.iter_collections
   # SSH testing is done via a side-loaded container that provides a POSIX/SSHable
   # server environment
   DATALAD_TESTS_DOCKER_SSHD_SECKEY_DOWNLOADURL: https://ci.appveyor.com/api/projects/mih/datalad-ci-docker-containers/artifacts/recipes/sshd/id_rsa?job=sshd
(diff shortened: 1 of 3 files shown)


mih:
The additions related to `next-status` now provide a nice test ground for getting this done.

mih:
Leaving a note here re adoption of pathspecs.

I found that one implementation is missing: Recursive discovery of untracked files that honors pathspecs.

A use case would be a recursive save of a pathspec-defined subset.

However, for consistency and simplicity of behavior, even a status report should be able to select or ignore untracked elements by pathspec.

A `git status` does not do that. A submodule is considered "untracked content" whenever there is any untracked content. It does not pass on pathspecs that may filter out all untracked content, and would make a submodule "clean considering constraints".

mih:
Closing this PR, because #714 has an advanced implementation.

Issue #586, opened 2024-01-07
Implement replacement for `datalad status`


mih:
With https://github.com/datalad/datalad-next/pull/580 and related iterators I quickly hacked together an iterator for changes in git repos (see https://github.com/datalad/datalad-next/issues/323 for the background). This essentially comprises all pieces necessary to reimplement `datalad status`.

This implementation would differ significantly, both in code base and also in approach.

- it would not be able to "list" the full content of a repository -- this is possible via `ls-file-collection (git|annex)worktree`, and it is a lot faster with that approach
- it would be "change-driven", ie. only operate and report on changes in a repo (the underlying implementation could still deliver a "status" and a "diff", like the one in datalad-core
- it would not run any datalad-core repo-class-level code, but exclusively rely on `datalad_next.iter_collections`

Based on initial performance tests of an unpublished draft, we can expect up to 100x speed increase for sizable repos (status report in <20ms instead of 1.5s (for a test repo with >30k files).

### `datalad status` behavior summary

Here is what present `datalad status` in core does.

```json
{
  "action": "status",
  "bytesize": 63,
  "gitshasum": "30e9ebb10380d9460522d063e53d39adecef4d94",
  "parentds": "/tmp/some",
  "path": "/tmp/some/.datalad/config",
  "prev_gitshasum": "30e9ebb10380d9460522d063e53d39adecef4d94",
  "refds": "/tmp/some",
  "state": "clean",
  "status": "ok",
  "type": "file"
}
```

`bytesize` is only present for files in Git (not for annexed files, unless `--annex` is given).

`state` can be the expected `clean`, `added`, `deleted`, `modified`, ...

### `datalad status --annex` behavior summary

```json
{
  "action": "status",
  "backend": "MD5E",
  "bytesize": 4,
  "gitshasum": "670d89de1eaf38bd9740fdc274ef214e0f63408f",
  "hashdirlower": "ff4/c57/",
  "hashdirmixed": "pF/Zf/",
  "humansize": "4 B",
  "key": "MD5E-s4--ba1f2511fc30423bdbb183fe33f3dd0f",
  "keyname": "ba1f2511fc30423bdbb183fe33f3dd0f",
  "mtime": "unknown",
  "parentds": "/tmp/some",
  "path": "/tmp/some/testme",
  "prev_gitshasum": "670d89de1eaf38bd9740fdc274ef214e0f63408f",
  "refds": "/tmp/some",
  "state": "clean",
  "status": "ok",
[...]

PR #589, opened 2024-01-10
New `iter_gitdiff`

mih:
This is a step towards https://github.com/datalad/datalad-next/issues/586 -- it implements comparison of Git tree-ishes with each other, or with *tracked* content in the worktree (ie., so support for untracked content).

Noteworthy items:

- includes test to verify that it does not suffer from https://github.com/datalad/datalad/issues/6791
- support for finding copies and renames (useful for https://github.com/datalad/datalad/issues/2160)
- uses `prev_gittype` for deletions (addresses https://github.com/datalad/datalad/issues/6877)

Performance notes

Comparing with `status` from datalad-core, on the studyforrest superdataset (11 subdatasets, ~4k files):

```
In [4]: %timeit status('.', result_renderer='disabled', untracked='no', recursive=True)
368 ms ± 10 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit list(iter_gitdiff(".", "HEAD", None, recursive='submodules'))
11.2 ms ± 41.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

3% of the original runtime. The result being "all-clean" in both cases.

```
❯ multitime -n5 -q git status -uno
===> multitime results
1: -q git status -uno
real        0.013       0.001       0.012       0.013       0.014       
user        0.006       0.003       0.002       0.006       0.012       
sys         0.009       0.004       0.004       0.008       0.013       
```

### DIFF
--- a/datalad_next/iter_collections/gitdiff.py
+++ b/datalad_next/iter_collections/gitdiff.py
@@ -72,11 +72,12 @@ class GitDiffItem(GitTreeItem):
     for modifications."""
 
     @cached_property
-    def prev_path(self) -> PurePosixPath:
+    def prev_path(self) -> PurePosixPath | None:
         """Returns the item ``prev_name`` as a ``PurePosixPath``
         instance"""
-        if self.prev_name:
-            return PurePosixPath(self.prev_name)
+        if self.prev_name is None:
+            return None
+        return PurePosixPath(self.prev_name)
 
 
 def iter_gitdiff(
@@ -88,7 +89,7 @@ def iter_gitdiff(
     find_renames: int | None = None,
     find_copies: int | None = None,
     yield_tree_items: str | None = None,
-) -> Generator[GitTreeItem, None, None]:
+) -> Generator[GitDiffItem, None, None]:
     """Report differences between Git tree-ishes or tracked worktree content
 
     This function is a wrapper around the Git command ``diff-tree`` and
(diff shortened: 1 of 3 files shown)


datalad_next/iter_collections/gitdiff.py
```diff
@@ -0,0 +1,430 @@
...
+def iter_gitdiff(
+    path: Path,
+    from_treeish: str | None,
+    to_treeish: str | None,
+    *,
+    recursive: str = 'repository',
+    find_renames: int | None = None,
+    find_copies: int | None = None,
+    yield_tree_items: str | None = None,
+) -> Generator[GitTreeItem, None, None]:
```
mih, line 92:
```suggestion
) -> Generator[GitDiffItem, None, None]:
```

mih:
I will leave this here for a bit, but will stop working on this PR and continue to build on it in another. It is ready from my POV.

adswa:
I played with it at least briefly, and once I figured out what to import and how to call it, it really felt lightning fast and smooth. I have to admit that the very Git-centric terminology and required Git familiarity made it a bit harder to understand, but that's on me - I think can fully appreciate it when I see it in use in front-facing commands or Gooey etc.

mih:
As a clarification: This is for the Git-layer, hence all Git terminology. The terminology would be different for everything that (eventually) comes out of a `datalad status`.

PR #592, opened 2024-01-11
Initial draft of `iter_gitstatus()` and `datalad next-status`

### LINKED ISSUE #323: Consolidate "traverser" from gooey and metalad into a common implementation

### LINKED ISSUE #418: Switch from result record dicts to dataclass
To be done post 1.0 release.

https://github.com/datalad/datalad/pull/7403 reminded me on the age-old mistake to use dicts for the results, instead of a more defined yet still flexible data structure.

With dataclasses now being available universally, this seems to be the goto.

Two patch targets seems sensible/necessary:

- `get_result_dict()` where most result dicts are created. However, much code does not use it, but creates dicts by hand.
- `eval_results` where each and every result passes through. Here legacy results can be intercepted and converted (without having to change any and all command implementations). This would allow for an instant switch after the initial yield in the original command.

Such a change enables binding of closely related functionality to such a class. The cide base needs to be scanned for common procedures that can be made obsolete. Candidates:

- is the result on something that went wrong (error, impossible) or something appreciated (ok, notneeded)
- does it match a particular path pattern
- formating of paths for a specific context
- reliably work with Path objects and str paths
- generic result rendering

If the machinery can be implemented sensibly, we could have different, derived result record types that each have their own renderer. This would finally address the problem of rendering results at a more global level, by a command that may not understand a particular result will. The renderer would be a property of a result, not a command.

For a transition, the result record class must implement enough of the dict interface, to make it work as a drop-in replacement. However, I think it would be better to limit the time legacy results are passed around in dict form, rather than making it a better dict.

It is worth thinking about JSON serialization/deserialization. I see value in support subtypes directly with a dedicated implementation.

### LINKED ISSUE #586: Implement replacement for `datalad status`

### LINKED ISSUE #597: Establish `repo_utils` subpackage
A design goal of datalad-next is to shed the dependency on the ancient `Repo` interface in datalad-core. It was established when DataLad used a more stateful representation of a Dataset/Repo, and when Git and git-annex where less powerful. Today it is largely an obstacle.

Nevertheless, there continues to be a need to have implementations for common query and manipulation tasks for repositories. We should establish a place for such implementation to promote their reuse, over reimplementations of near-copies.

#592 brings a whole range of internal helper that should be promoted to the public, and equipped with tests.

It would be worthwhile to *not* require a standard data structure (like a `Dataset`) for such helpers, but work with the bare necessities (often just a path) to promote wide applicability.

Moreover, it would be helpful to aim for generic implementations that work with git and git-annex repos alike, and also in adjusted mode repos -- unless there is a speed difference that justifies alternative implementations, or if there is a substantial complexity cost to pay.

mih:
[Although this PR is big, it does not break API (only additions). The only "breaking" change is a relocation of `iter_git_subproc()`, which was never released so far.]

- Closes #323
- Closes #418
- Closes #586
- Closes #597

It can eventually replace `datalad status`.

Feature and differences of this new implementation (compared to datalad-core):

- Does not use any of the datalad-core tooling for diffing (`Repo.diffstatus`, etc.)
- all generator-based implementation
- command results are `dataclass` now, no longer result-dict -- first of this kind
- more closely aligned with Git, for example, no listings, just change reports
- capable of reporting modification types for container items (submodules, directories): new commits, modified/untracked content
- proper typechange detection
- strict separation of datalad command implementation from underlying tooling. The command implementation is solely a frontend API with no dedicated functionality beyond input validation, result rendering -- and largely semantic glue for backward compatibility for datalad-core's `status`.

It is substantially faster on dataset hierarchies (11 datasets, 4k
files):

```
In [3]: %timeit list(iter_gitstatus(".", recursive='submodules', untracked='all'))
111 ms ± 506 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit status('.', result_renderer='disabled', untracked='all', recursive=True)
199 ms ± 3.9 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

and on large individual datasets (36k tracked and 2k untracked files):

```
In [3]: %timeit status('.', result_renderer='disabled', untracked='all', recursive=True)
1.44 s ± 97 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit list(iter_gitstatus(".", recursive='submodules', untracked='all'))
233 ms ± 31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

It also supports the "single-dir" or no-recursion mode, where the report is contrained to the items in a single directory (needed by Gooey for lean population of the file tree).

Here is a comparison for a convoluted situation. First the report by a standard `git-status`, and then the outputs for each recursion mode (as implemented right now):

```
❯ git status
On branch dl-test-branch
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        deleted:    dir_d/file_d
        deleted:    dir_m/file_d
        modified:   dir_m/file_m
        deleted:    dir_sm/sm_d
        modified:   dir_sm/sm_mu (modified content, untracked content)
        modified:   dir_sm/sm_n (new commits)
        modified:   dir_sm/sm_nm (new commits, modified content)
        modified:   dir_sm/sm_nmu (new commits, modified content, untracked content)
        modified:   dir_sm/sm_u (untracked content)
        deleted:    file_d
        modified:   file_m

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        dir_m/dir_u/
        dir_m/file_u
        dir_u/
        file_u
```

`recursive='no'`

```
❯ datalad next-status -r no
  deleted: dir_d (directory)
 modified: dir_m (directory) [modified content, untracked content]
 modified: dir_sm (directory) [modified content, untracked content]
  deleted: file_d (file)
 modified: file_m (file)
untracked: dir_u
untracked: file_u
```

`recursive='repository'` (standard)

```
❯ datalad next-status
  deleted: dir_d/file_d (file)
  deleted: dir_m/file_d (file)
 modified: dir_m/file_m (file)
  deleted: dir_sm/sm_d (submodule)
 modified: dir_sm/sm_n (submodule) [new commits]
 modified: dir_sm/sm_nm (submodule) [new commits, modified content]
 modified: dir_sm/sm_nmu (submodule) [new commits, modified content, untracked content]
  deleted: file_d (file)
 modified: file_m (file)
 modified: dir_sm/sm_mu (submodule) [modified content, untracked content]
 modified: dir_sm/sm_u (submodule) [untracked content]
untracked: dir_m/dir_u
untracked: dir_m/file_u
untracked: dir_u
untracked: file_u
```

`recursive='datasets'`

```
❯ datalad next-status -r
  deleted: dir_d/file_d (file)
  deleted: dir_m/file_d (file)
 modified: dir_m/file_m (file)
  deleted: dir_sm/sm_d (submodule)
 modified: dir_sm/sm_nm/file_m (file)
 modified: dir_sm/sm_nmu/file_m (file)
untracked: dir_sm/sm_nmu/dir_u
untracked: dir_sm/sm_nmu/file_u
  deleted: file_d (file)
 modified: file_m (file)
 modified: dir_sm/sm_mu/file_m (file)
untracked: dir_sm/sm_mu/dir_u
untracked: dir_sm/sm_mu/file_u
untracked: dir_sm/sm_u/dir_u
untracked: dir_sm/sm_u/file_u
untracked: dir_m/dir_u
untracked: dir_m/file_u
untracked: dir_u
untracked: file_u
```

TODO:
- [x] Tests for `recursive='no'`
- [x] Tests for `recursive='submodules'`
- [x] Tests for handling staged content
- [x] Basic tests for `CommandResult` class
- [x] Tests for `next-status` command implementation
- [x] Implement a *mono* (monolitic) recursive mode where submodules are not checked against their own HEAD, but against the commit that is recorded in the parent repository. This mode would make it unnecessary to emit results on submodule records, because any and all modification with respect to the parent would be discovered.

### DIFF
--- a/datalad_next/commands/status.py
+++ b/datalad_next/commands/status.py
@@ -1,0 +1,375 @@
+"""
+"""
+from __future__ import annotations
+
+__docformat__ = 'restructuredtext'
+
+from dataclasses import dataclass
+from enum import Enum
+from logging import getLogger
+from pathlib import Path
+from typing import Generator
+
+from datalad_next.commands import (
+    CommandResult,
+    CommandResultStatus,
+    EnsureCommandParameterization,
+    ValidatedInterface,
+    Parameter,
+    ParameterConstraintContext,
+    build_doc,
+    datasetmethod,
+    eval_results,
+)
+from datalad_next.constraints import (
+    EnsureChoice,
+    WithDescription,
+)
+from datalad_next.constraints.dataset import EnsureDataset
+
+from datalad_next.iter_collections.gitdiff import (
+    GitDiffStatus,
+    GitTreeItemType,
+    GitContainerModificationType,
+)
+from datalad_next.iter_collections.gitstatus import (
+    iter_gitstatus,
+)
+from datalad_next.uis import (
+    ui_switcher as ui,
+    ansi_colors as ac,
+)
+
+lgr = getLogger('datalad.core.local.status')
+
+
+# TODO Could be `StrEnum`, came with PY3.11
+class StatusState(Enum):
+    """Enumeration of possible states of a status command result
+
+    The "state" is the condition of the dataset item being reported
+    on.
+    """
+    clean = 'clean'
+    added = 'added'
+    modified = 'modified'
+    deleted = 'deleted'
+    untracked = 'untracked'
+    unknown = 'unknown'
+
+
+STATE_COLOR_MAP = {
+    StatusState.added: ac.GREEN,
+    StatusState.modified: ac.RED,
+    StatusState.deleted: ac.RED,
+    StatusState.untracked: ac.RED,
+    StatusState.unknown: ac.YELLOW,
+}
+
+
+diffstatus2resultstate_map = {
+    GitDiffStatus.addition: StatusState.added,
+    GitDiffStatus.copy: StatusState.added,
+    GitDiffStatus.deletion: StatusState.deleted,
+    GitDiffStatus.modification: StatusState.modified,
+    GitDiffStatus.rename: StatusState.added,
+    GitDiffStatus.typechange: StatusState.modified,
+    GitDiffStatus.unmerged: StatusState.unknown,
+    GitDiffStatus.unknown: StatusState.unknown,
+    GitDiffStatus.other: StatusState.untracked,
+}
+
+
+# see base class decorator comment for why this is commented out
+#@dataclass(kw_only=True)
+@dataclass
+class StatusResult(CommandResult):
+    # TODO any of the following property are not actually optional
+    # we only have to declare them such for limitations of dataclasses
+    # prior PY3.10 (see kw_only command in base class
+
+    diff_state: GitDiffStatus | None = None
+    """The ``status`` of the underlying ``GitDiffItem``. It is named
+    "_state" to emphasize the conceptual similarity with the legacy
+    property 'state'
+    """
+    gittype: GitTreeItemType | None = None
+    """The ``gittype`` of the underlying ``GitDiffItem``."""
+    prev_gittype: GitTreeItemType | None = None
+    """The ``prev_gittype`` of the underlying ``GitDiffItem``."""
+    modification_types: tuple[GitContainerModificationType] | None = None
+    """Qualifiers for modification types of container-type
+    items (directories, submodules)."""
+
+    @property
+    def state(self) -> StatusState:
+        """A (more or less legacy) simplified representation of the subject
+        state. For a more accurate classification use the ``diff_status``
+        property.
+        """
+        return diffstatus2resultstate_map[self.diff_state]
+
+    # the previous status-implementation did not report plain git-types
+    # we establish a getter to perform this kind of inference/mangling,
+    # when it is needed
+    @property
+    def type(self) -> str | None:
+        """
+        """
+        # TODO this is just a placeholder
+        return self.gittype.value if self.gittype else None
+
+    # we need a setter for this `type`-override stunt
+    @type.setter
+    def type(self, value):
+        self.gittype = value
+
+    @property
+    def prev_type(self) -> str:
+        """
+        """
+        return self.prev_gittype.value if self.prev_gittype else None
+
+    @property
+    def type_src(self) -> str | None:
+        """Backward-compatibility adaptor"""
+        return self.prev_type
+
+
+opt_untracked_values = ('no', 'whole-dir', 'no-empty-dir', 'normal', 'all')
+opt_recursive_values = ('no', 'repository', 'datasets', 'mono')
+opt_eval_subdataset_state_values = ('no', 'commit', 'full')
+
+
+class StatusParamValidator(EnsureCommandParameterization):
+    def __init__(self):
+        super().__init__(
+            param_constraints=dict(
+                # if given, it must also exist
+                dataset=EnsureDataset(installed=True),
+                untracked=EnsureChoice(*opt_untracked_values),
+                recursive=EnsureChoice(*opt_recursive_values),
+                eval_subdataset_state=EnsureChoice(
+                    *opt_eval_subdataset_state_values)
+            ),
+            validate_defaults=('dataset',),
+            joint_constraints={
+                ParameterConstraintContext(('untracked', 'recursive'),
+                                           'option normalization'):
+                self.normalize_options,
+            },
+        )
+
+    def normalize_options(self, **kwargs):
+        if kwargs['untracked'] == 'no':
+            kwargs['untracked'] = None
+        if kwargs['untracked'] == 'normal':
+            kwargs['untracked'] = 'no-empty-dir'
+        if kwargs['recursive'] == 'datasets':
+            kwargs['recursive'] = 'submodules'
+        if kwargs['recursive'] == 'mono':
+            kwargs['recursive'] = 'monolithic'
+        return kwargs
+
+
+@build_doc
+class Status(ValidatedInterface):
+    """Report on the (modification) status of a dataset
+
+    .. note::
+
+        This is a preview of an command implementation aiming to replace
+        the DataLad ``status`` command.
+
+        For now, expect anything here to change again.
+
+    This command provides a report that is roughly identical to that of
+    ``git status``. Running with default parameters yields a report that
+    should look familiar to Git and DataLad users alike, and contain
+    the same information as offered by ``git status``.
+
+    The main difference to ``git status`` are:
+
+    - Support for recursion into submodule. ``git status`` does that too,
+      but the report is limited to the global state of an entire submodule,
+      whereas this command can issue detailed reports in changes inside
+      a submodule (any nesting depth).
+
+    - Support for directory-constrained reporting. Much like ``git status``
+      limits its report to a single repository, this command can optionally
+      limit its report to a single directory and its direct children. In this
+      report subdirectories are considered containers (much like) submodules,
+      and a change summary is provided for them.
+
+    - Support for a "mono" (monolithic repository) report. Unlike a standard
+      recursion into submodules, and checking each of them for changes with
+      respect to the HEAD commit of the worktree, this report compares a
+      submodule with respect to the state recorded in its parent repository.
+      This provides an equally comprehensive status report from the point of
+      view of a queried repository, but does not include a dedicated item on
+      the global state of a submodule. This makes nested hierarchy of
+      repositories appear like a single (mono) repository.
+
+    - Support for "adjusted mode" git-annex repositories. These utilize a
+      managed branch that is repeatedly rewritten, hence is not suitable
+      for tracking within a parent repository. Instead, the underlying
+      "corresponding branch" is used, which contains the equivalent content
+      in an un-adjusted form, persistently. This command detects this condition
+      and automatically check a repositories state against the corresponding
+      branch state.
+
+    *Presently missing/planned features*
+
+    - There is no support for specifying paths (or pathspecs) for constraining
+      the operation to specific dataset parts. This will be added in the
+      future.
+
+    - There is no reporting of git-annex properties, such as tracked file size.
+      It is undetermined whether this will be added in the future. However,
+      even without a dedicated switch, this command has support for
+      datasets (and their submodules) in git-annex's "adjusted mode".
+
+    *Differences to the ``status`` command implementation prior DataLad v2*
+
+    - Like ``git status`` this implementation reports on dataset modification,
+      whereas the previous ``status`` also provided a listing of unchanged
+      dataset content. This is no longer done. Equivalent functionality for
+      listing dataset content is provided by the ``ls_file_collection``
+      command.
+    - The implementation is substantially faster. Depending on the context
+      the speed-up is typically somewhere between 2x and 100x.
+    - The implementation does not suffer from the limitation re type change
+      detection.
+    - Python and CLI API of the command use uniform parameter validation.
+    """
+    # Interface.validate_args() will inspect this dict for the presence of a
+    # validator for particular parameters
+    _validator_ = StatusParamValidator()
+
+    # this is largely here for documentation and CLI parser building
+    _params_ = dict(
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""Dataset to be used as a configuration source. Beyond
+            reading configuration items, this command does not interact with
+            the dataset."""),
+        untracked=Parameter(
+            args=('--untracked',),
+            choices=opt_untracked_values,
+            doc="""Determine how untracked content is considered and reported
+            when comparing a revision to the state of the working tree.
+            'no': no untracked content is considered as a change;
+            'normal': untracked files and entire untracked directories are
+            reported as such;
+            'all': report individual files even in fully untracked directories.
+            In addition to these git-status modes,
+            'whole-dir' (like normal, but include empty directories), and
+            'no-empty-dir' (alias for 'normal') are understood."""),
+        recursive=Parameter(
+            args=('-r', '--recursive'),
+            nargs='?',
+            const='datasets',
+            choices=opt_recursive_values,
+            doc="""Mode of recursion for status reporting.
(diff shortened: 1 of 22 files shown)


adswa (commented):
Is this new status on purpose not exposed in the command suite?

datalad_next/commands/status.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @property
+    def type_src(self) -> str | None:
+        """Backward-compatibility adaptor"""
+        return self.prev_type
+
+
+@build_doc
+class Status(ValidatedInterface):
+    """The is a previous of an upcoming command implementation to replace
```
adswa, line 101:
```suggestion
    """The is a preview of an upcoming command implementation to replace
```
?

datalad_next/commands/status.py
```diff
@@ -0,0 +1,200 @@
...
+    @datasetmethod(name="next_status")
+    @eval_results
+    def __call__(
+        # TODO later
+        #path=None,
+        *,
+        dataset=None,
+        # TODO later
+        #annex=None,
+        untracked='whole-dir',
```
adswa, line 154: This default mode isn't documented in the parameter docstring

datalad_next/commands/status.py
```diff
@@ -0,0 +1,200 @@
...
+            doc="""If and how untracked content is reported when comparing
+            a revision to the state of the working tree. 'no': no untracked
+            content is reported; 'normal': untracked files and entire
+            untracked directories are reported as such; 'all': report
+            individual files even in fully untracked directories."""),
+        recursive=Parameter(
+            args=('-r', '--recursive'),
+            nargs='?',
+            const='datasets',
+            doc="some"),
```
adswa, line 138: doc can probably be copied from ``iter_gitstatus``' recursive parameter?

datalad_next/commands/status.py
```diff
@@ -0,0 +1,200 @@
...
+            the dataset."""),
+        untracked=Parameter(
+            args=('--untracked',),
+            nargs='?',
+            const='no-empty-dir',
+            doc="""If and how untracked content is reported when comparing
+            a revision to the state of the working tree. 'no': no untracked
+            content is reported; 'normal': untracked files and entire
+            untracked directories are reported as such; 'all': report
+            individual files even in fully untracked directories."""),
```
adswa, line 133: I tried this, and `no` and `normal` aren't yet implemented, and produce a plain ``[ERROR]``. I assume you know, but wanted to leave a comment nevertheless

```
(handbook-upgrade) adina@muninn in ~/repos/datalad-next on git:gitstatus
❱ datalad next-status --untracked no
[ERROR  ] 'no' 
(handbook-upgrade) adina@muninn in ~/repos/datalad-next on git:gitstatus
❱ datalad next-status --untracked normal                                    1 !
[ERROR  ] 'normal' 
(handbook-upgrade) adina@muninn in ~/repos/datalad-next on git:gitstatus
❱ datalad next-status --untracked all                                       1 !
untracked: .idea/datalad-next.iml
untracked: .idea/inspectionProfiles/profiles_settings.xml
untracked: .idea/modules.xml
untracked: .idea/vcs.xml
untracked: .idea/workspace.xml
untracked: some
(handbook-upgrade) adina@muninn in ~/repos/datalad-next on git:gitstatus
❱ datalad next-status --untracked whole-dir
untracked: .idea
untracked: .mypy_cache
untracked: .pytest_cache
untracked: __pycache__
```

datalad_next/commands/status.py
```diff
@@ -0,0 +1,212 @@
...
+        max_len = len('untracked')
+        # message format is same as for previous command implementation
+        ui.message(u'{fill}{state}: {path}{type_}'.format(
+            fill=' ' * max(0, max_len - len(res.state.value)),
+            state=ac.color_word(
+                res.state.value,
+                STATE_COLOR_MAP.get(res.state)),
+            path=path,
+            type_=' ({})'.format(ac.color_word(type_str, ac.MAGENTA))
+            if type_str else ''),
```
adswa, line 211: I noticed that unlike ``datalad status``, ``next-status`` does not report "nothing to save, working tree clean" when run in a clean dataset. Is that intentional?

adswa (commented):
It may well be faulty on ``datalad status``' side (I didn't yet look close), but there is a difference in the file type reported for an unstaged typechange from symlink to file between ``status`` and ``next-status``:

```
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master
❱ git annex unlock .
unlock anotherfile ok
(recording state in git...)
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master+
❱ datalad status
 modified: anotherfile (file)
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master+
❱ datalad next-status
 modified: anotherfile (file)
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master+
❱ git status
On branch master
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	typechange: anotherfile

(handbook-upgrade) adina@muninn in /tmp/teststate on git:master+
❱ git restore --staged .
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master!
❱ git status
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	typechange: anotherfile

no changes added to commit (use "git add" and/or "git commit -a")
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master!
❱ datalad status
 modified: anotherfile (symlink)
(handbook-upgrade) adina@muninn in /tmp/teststate on git:master!
❱ datalad next-status
 modified: anotherfile (file)
```

datalad_next/commands/status.py
mih: I have not ported the result summary renderer yet. No other reason.

datalad_next/commands/status.py
mih: Thanks. Applied.

datalad_next/commands/status.py
mih: Done now. Thx!

datalad_next/commands/status.py
mih: Almost. I am working on this. There are some glitches.

datalad_next/commands/status.py
mih: Addressed locally now. Thx!

mih:
> It may well be faulty on `datalad status`' side (I didn't yet look close), but there is a difference in the file type reported for an unstaged typechange

Yes, this is a difference, and it needs to be figured out, how it should be addresses.

`datalad status` reported "would-be" types (only). If a symlink is to an annexed file then it decides that a user would want to not-know that this is actually a symlink (because it is, in some sense, an implementation detail).

`datalad next-status` reports the type as known to Git.

I will likely make that Git-type an explicit property of the result, and keep `type` for backward compatibility. However, the way `datalad status` infers the difference of a symlink from annex vs a symlink for other reasons is expensive. Not sure if I am willing to maintain this behavior entirely.

datalad_next/commands/status.py
mih: Done now

mih:
Adjusted mode is properly supported now.

datalad_next/commands/status.py
mih: I am resolving this (for now). All command docs are TODO.

datalad_next/iter_collections/gitstatus.py
```diff
@@ -0,0 +1,521 @@
...
+        from_treeish='HEAD',
+        # to the worktree
+        to_treeish=None,
+        recursive='repository',
+        # TODO check if we can go cheaper. We need to redo some check for
+        # adjusted mode, and other cases anyways
+        eval_submodule_state='commit'
+        if eval_submodule_state == 'full' else eval_submodule_state,
+        #eval_submodule_state=eval_submodule_state,
+    ):
```
mslw, line 262: I found an edge case: when `datalad next-status` is called in an empty git repo, this happens:

```
❱ git init
Initialized empty Git repository in /tmp/foo/.git/
❱ datalad next-status
CommandError: 'git diff-index --no-rename-empty --relative --raw -z --ignore-submodules=dirty HEAD' failed with exitcode 128 under /tmp/foo
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
```

There are no commits, so HEAD points to a `refs/heads/main` which does not exist yet. Both `git status` and `datalad status` work:

```
❱ git status
On branch main

No commits yet

nothing to commit (create/copy files and use "git add" to track)

❱ datalad status
nothing to save, working tree clean
```

(same applies to other `_yield*` functions)

That being said, it won't be relevant to DataLad datasets, because AFAIK `datalad create` always commits.

mih: Thanks! I will fix this!

mih: This was a bit more complex than anticipated, but it should work now.

I took the opportunity to restructure the implementations a bit and started to push towards #597

adswa:
I toyed with the ``mono`` recursive mode, and in general, I find that it works nicely as the name indicates - thanks! Unrelated to this specific mode, I noticed that untracked content (regardless of whether its files, directories, or datasets) doesn't get a type annotation (ie no ``(directory)``, ``(symlink)``, ``(file)`` appendix to the path). If this is intentional, we should add this behavior to the docs (its not a problem, just caught my eye)

mih:
> I toyed with the `mono` recursive mode, and in general, I find that it works nicely as the name indicates - thanks! Unrelated to this specific mode, I noticed that untracked content (regardless of whether its files, directories, or datasets) doesn't get a type annotation (ie no `(directory)`, `(symlink)`, `(file)` appendix to the path). If this is intentional, we should add this behavior to the docs (its not a problem, just caught my eye)

I looked at this, and we can get a cheap (near-zero) cost discrimination of directory-type items from other types (file, symlink). So I went for this. I did not add a type investigation to discriminate between files and symlinks. This would require touching each file, and make things slow.

For a user this should be "good enough", because they can see what is a directory and what is not. This should provide the main property of interest, i.e. "can there be something underneath?".

mih:
I think for this first round this is good.

Issue #590, opened 2024-01-11
Unreliable: `FAILED ..\patches\tests\test_push.py::test_force_checkdatapresent`


mih:
This test has an unacceptable failure-rate. I guesstimate 1/10.

PR #595, opened 2024-01-15
Ensure that the return-value of ssh-subprocess is used when `StopIteration` is raised.

### LINKED ISSUE #594: Return value of remote `ssh`-commands is not processed if `StopIteration` is raised locally
In the current `SshUrlOperations`-implementation a `StopIteration`-exception is raised on a short-read, i.e. the client expects more data but the server does not send it, due to an error condition.

These `StopIteration`-exceptions are not caught inside the `iterable_subprocess`-contexts.  As a result, `iterable_subprocesses` sends a `termination`-signal to the ssh-client. Depending on the state of the ssh-client, it will then either return `255` or the exit value of the remote command. In the `SshUrlOperations` the result is usually `255`, because the ssh-client usually has not terminated when the `StopIteration`-error is processed. That means, the ssh-client will be terminated before it has the chance to exit normally.

Currently, we map all `StopIteration`-triggered ssh-client terminations onto the return value `244`, assuming they all indicate an unknown remote resource. This is not a sound approach and will fail if the remote command signals different errors.

The `StopIteration` exceptions should not be allowed to exit the `iterable_subprocess`-context, in order to be able to fetch the exit code of the remote commandfrom the ssh-client.

The change should also be done to demonstrate the "correct" programing paradigm when dealing with expected exceptions in `iterable_subprocess`-contexts.

christian-monch:
This PR fixes #594

It modifies the implementation of `SshUrlOperations` so that it catches expected exceptions before they leave the `iterable_subprocess`-context. This leads to a proper shutdown of the ssh-subprocess, via closing its stdin, and to a reliable signaling of the return code of the remote command.

The latter would not always be the case if the `iterable_subprocess`-context is left with an exception, because that triggers a `terminate()` on the shell process, and the exit code of the ssh-subprocess will reflect this termination.

### DIFF
--- a/datalad_next/url_operations/ssh.py
+++ b/datalad_next/url_operations/ssh.py
@@ -99,13 +99,17 @@ class SshUrlOperations(UrlOperations):
         cmd = ssh_cat.get_cmd(SshUrlOperations._stat_cmd)
         try:
             with iter_subproc(cmd) as stream:
-                props = self._get_props(url, stream)
+                try:
+                    props = self._get_props(url, stream)
+                except StopIteration:
+                    # we did not receive all data that should be sent, if a
+                    # remote file exists. This indicates a non-existing
+                    # resource or some other problem. The remotely executed
+                    # command should signal the error via a non-zero exit code.
+                    # That will trigger a `CommandError` below.
+                    pass
         except CommandError:
             self._check_return_code(stream.returncode, url)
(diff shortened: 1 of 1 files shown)


mih:
Test failure is #590

mih (approved):
Makes sense to me. Thanks!

Issue #572, opened 2023-12-15
`PersistentSubShell` feature


mih:
Let's start with the thing itself only, and not a full set of abstractions.

- [X] focus on POSIX shell clients 
- [X] methods to execute a command with support for iterator for stdin and reading stdout/err, and report the return code
- [ ] need not work with any shell, but should be explicit what shells with which params are compatible
- [X] should work with such a shell executed via ssh on a remote system
- [X] must be "persistent" such that additional instructions/commands can be sent into the same session, bases on prior output

- have additional(?) convenience class with basic helpers to:
  - [X] write/upload,
  - [x] read/download,
  - [x] delete files,
  - [ ] create directories,
  - [ ] remove directories

__
This feature should avoid the need for ssh multiplexing (not available for windows) for any non-parallel processing.

EDIT: work is ongoing in branch https://github.com/christian-monch/datalad-next/tree/ssh-shell; tasks have been flagged as done according to the progress in that branch.

christian-monch:
I have a question w.r.t. point 2:

> - methods to execute a command with support for iterator for stdin and reading stdout/err, and report the return code

How is the iterator for stdin supposed to be used? I am currently working on a ssh-shell command which supports the following control flow:

```python
    with ssh_connection('host.example.com') as ssh:
        # execute 'ls' on the remote host
        results = ssh.execute_command(b'ls -l /etc'):
        for line in results:
            print(line.decode())
        # check the return code of remote 'ls'-command
        if results.code == 0:
            # do something else on the remote host
            results = ssh.execute_command(b'cat /etc/fstab'):
            for line in results:
                print(line.decode())
            assert results.code == 0
```

Is that along the lines of point 2?

mih:
Your example would be matching my "output". An example of an input iterator would be a file upload, by running `cat > somepath` on the remote end, and `cat inpath` locally via `iter_subproc` to get its "input iterator".

Does that make sense?

christian-monch:
That makes sense.

I added the possibility to feed stdin from an iterable, e.g. an `iter_subproc`-result. This can be done after an arbitrary number of commands without stdin-redirection have been executed on the `ssh_connection`.

```python
    with \
            ssh_connection(['ssh', 'localhost']) as ssh, \
            iter_subproc(['cat', '/etc/fstab']) as fstab_cat:

        results = ssh.execute_command(b'ls -l /etc/fstab')
        for line in results:
            print('ssh:', line.decode())
        print(results.code)
        
[...]

PR #596, opened 2024-01-15
Add `PersistentSubShell`-feature

christian-monch:
This PR addresses issue #572 

It provides a `shell_connection`-context manager that will start a shell, for example, a local ssh-client, and allows interaction with the shell-process. Interactions are started by sending a command, which will be executed in the shell. The user will then receive the command's `stdout`. When the command terminates, the user can retrieve the exit code of the command.

Usage example:
```python
>>> from datalad_next.shell import shell
>>> with shell(['ssh', 'localhost']) as ssh:
...     r = ssh(b'ls -l /etc/passwd')
...     print(tuple(r))
...     print(r.returncode)
... 
(b'-rw-r--r-- 1 root root 2773 14. Nov 10:05 /etc/passwd\n',)
0
```

### DIFF
--- a/datalad_next/shell/__init__.py
+++ b/datalad_next/shell/__init__.py
@@ -136,7 +136,7 @@ per-call basis by providing a different response generator class in the
    VariableLengthResponseGeneratorPowerShell
    FixedLengthResponseGenerator
    FixedLengthResponseGeneratorPosix
-   FixedLengthResponseGeneratorPowershell
+   FixedLengthResponseGeneratorPowerShell
    DownloadResponseGenerator
    DownloadResponseGeneratorPosix
    DownloadResponseGeneratorOSX
@@ -163,11 +163,11 @@ from .operations.posix import (
     DownloadResponseGeneratorPosix,
 )
 from .response_generators import (
+    FixedLengthResponseGenerator,
+    FixedLengthResponseGeneratorPosix,
+    FixedLengthResponseGeneratorPowerShell,
     ShellCommandResponseGenerator,
     VariableLengthResponseGenerator,
     VariableLengthResponseGeneratorPosix,
     VariableLengthResponseGeneratorPowerShell,
-    FixedLengthResponseGenerator,
-    FixedLengthResponseGeneratorPosix,
-    FixedLengthResponseGeneratorPowershell,
 )
--- a/datalad_next/shell/__init__.py
+++ b/datalad_next/shell/__init__.py
@@ -139,7 +139,6 @@ per-call basis by providing a different response generator class in the
    FixedLengthResponseGeneratorPowerShell
    DownloadResponseGenerator
    DownloadResponseGeneratorPosix
-   DownloadResponseGeneratorOSX
    operations.posix.upload
    operations.posix.download
    operations.posix.delete
@@ -159,7 +158,6 @@ from .shell import (
 from .operations import posix
 from .operations.posix import (
     DownloadResponseGenerator,
-    DownloadResponseGeneratorOSX,
     DownloadResponseGeneratorPosix,
 )
 from .response_generators import (
--- a/datalad_next/shell/__init__.py
+++ b/datalad_next/shell/__init__.py
@@ -1,0 +1,173 @@
+"""A persistent shell connection
+
+This module provides a context manager that establishes a connection to a shell
+and can be used to execute multiple commands in that shell. Shells are usually
+remote shells, e.g. connected via an ``ssh``-client, but local shells like
+``zsh``, ``bash`` or ``PowerShell`` can also be used.
+
+The context manager returns an instance of :class:`ShellCommandExecutor` that
+can be used to execute commands in the shell via the method
+:meth:`ShellCommandExecutor.__call__`. The method will return an instance of
+a subclass of :class:`ShellCommandResponseGenerator` that can be used to
+retrieve the output of the command, the result code of the command, and the
+stderr-output of the command.
+
+Every response generator expects a certain output structure. It is responsible
+for ensuring that the output structure is generated. To this end every
+response generator provides a method
+:meth:`ShellCommandResponseGenerator.get_command_list`. The method
+:class:`ShellCommandExecutor.__call__` will pass the user-provided command to
+:meth:`ShellCommandResponseGenerator.get_command_list` and receive a list of
+final commands that should be executed in the connected shell and that will
+generate the expected output structure. Instances of
+:class:`ShellCommandResponseGenerator` have therefore four tasks:
+
+    1. Create a final command list that is used to execute the user provided
+       command. This could, for example, execute the command, print an
+       end marker, and print the return code of the command.
+
+    2. Parse the output of the command, yield it to the user.
+
+    3. Read the return code and provide it to the user.
+
+    4. Provide stderr-output to the user.
+
+A very versatile example of a response generator is the class
+:class:`VariableLengthResponseGenerator`. It can be used to execute a command
+that will result in an output of unknown length, e.g. ``ls``, and will yield
+the output of the command to the user. It does that by using a random
+*end marker* to detect the end of the output and read the trailing return code.
+This is suitable for almost all commands.
+
+If :class:`VariableLengthResponseGenerator` is so versatile, why not just
+implement its functionality in :class:`ShellCommandExecutor`? There are two
+major reasons for that:
+
+1. Although the :class:`VariableLengthResponseGenerator` is very versatile,
+   it is not the most efficient implementation for commands that produce large
+   amounts of output. In addition, there is also a minimal risk that the end
+   marker is part of the output of the command, which would trip up the response
+   generator. Putting response generation into a separate class allows to
+   implement specific operations more efficiently and more safely.
+   For example,
+   :class:`DownloadResponseGenerator` implements the download of files. It
+   takes a remote file name as user "command" and creates a final command list
+   that emits the length of the file, a newline, the file content, a return
+   code, and a newline. This allows :class:`DownloadResponseGenerator`
+   to parse the output without relying on an end marker, thus increasing
+   efficiency and safety
+
+2. Factoring out the response generation creates an interface that can be used
+   to support the syntax of different shells and the difference in command
+   names and options in different operating systems. For example, the response
+   generator class :class:`VariableLengthResponseGeneratorPowerShell` supports
+   the invocation of commands with variable length output in a ``PowerShell``.
+
+In short, parser generator classes encapsulate details of shell-syntax and
+operation implementation. That allows support of different shell syntax, and
+the efficient implementation of specific higher level operations, e.g.
+``download``. It also allows users to extend the functionality of
+:class:`ShellCommandExecutor` by providing their own response generator
+classes.
+
+The module :mod:`datalad_next.shell.response_generators` provides two generally
+applicable abstract response generator classes:
+
+    - :class:`VariableLengthResponseGenerator`
+
+    - :class:`FixedLengthResponseGenerator`
+
+The functionality of the former is described above. The latter can be used to
+execute a command that will result in output of known
+length, e.g. ``echo -n 012345``. It reads the specified number of bytes and a
+trailing return code. This is more performant than the variable length response
+generator (because it does not have to search for the end marker). In addition,
+it does not rely on the uniqueness of the end marker. It is most useful for
+operation like ``download``, where the length of the output can be known in
+advance.
+
+As mentioned above, the classes :class:`VariableLengthResponseGenerator` and
+:class:`FixedLengthResponseGenerator` are abstract. The module
+:mod:`datalad_next.shell.response_generators` provides the following concrete
+implementations for them:
+
+    - :class:`VariableLengthResponseGeneratorPosix`
+
+    - :class:`VariableLengthResponseGeneratorPowerShell`
+
+    - :class:`FixedLengthResponseGeneratorPosix`
+
+    - :class:`FixedLengthResponseGeneratorPowerShell`
+
+When :func:`shell` is executed it will use a
+:class:`VariableLengthResponseClass` to skip the login message of the shell.
+This is done by executing a *zero command* (a command that will possibly
+generate some output, and successfully return) in the shell. The zero command is
+provided by the concrete implementation of class
+:class:`VariableLengthResponseGenerator`. For example, the zero command for
+POSIX shells is ``test 0 -eq 0``, for PowerShell it is ``Write-Host hello``.
+
+Because there is no way for func:`shell` to determine the kind of shell it
+connects to, the user can provide an alternative response generator class, in
+the ``zero_command_rg_class``-parameter. Instance of that class
+will then be used to execute the zero command. Currently, the following two
+response generator classes are available:
+
+    - :class:`VariableLengthResponseGeneratorPosix`: works with POSIX-compliant
+      shells, e.g. ``sh`` or ``bash``. This is the default.
+    - :class:`VariableLengthResponseGeneratorPowerShell`: works with PowerShell.
+
+Whenever a command is executed via :meth:`ShellCommandExecutor.__call__`, the
+class identified by ``zero_command_rg_class`` will be used by default to create
+the final command list and to parse the result. Users can override this on a
+per-call basis by providing a different response generator class in the
+``response_generator``-parameter of :meth:`ShellCommandExecutor.__call__`.
+
+.. currentmodule:: datalad_next.shell
+
+.. autosummary::
+   :toctree: generated
+   :recursive:
+
+   ShellCommandExecutor
+   ShellCommandResponseGenerator
+   VariableLengthResponseGenerator
+   VariableLengthResponseGeneratorPosix
+   VariableLengthResponseGeneratorPowerShell
+   FixedLengthResponseGenerator
+   FixedLengthResponseGeneratorPosix
+   FixedLengthResponseGeneratorPowershell
+   DownloadResponseGenerator
+   DownloadResponseGeneratorPosix
+   DownloadResponseGeneratorOSX
+   operations.posix.upload
+   operations.posix.download
+   operations.posix.delete
+"""
+
+
+__all__ = [
+    'shell',
+    'posix',
+]
+
+from .shell import (
+    shell,
+    ShellCommandExecutor,
+)
+
+from .operations import posix
+from .operations.posix import (
+    DownloadResponseGenerator,
+    DownloadResponseGeneratorOSX,
+    DownloadResponseGeneratorPosix,
+)
+from .response_generators import (
+    ShellCommandResponseGenerator,
+    VariableLengthResponseGenerator,
+    VariableLengthResponseGeneratorPosix,
+    VariableLengthResponseGeneratorPowerShell,
+    FixedLengthResponseGenerator,
+    FixedLengthResponseGeneratorPosix,
+    FixedLengthResponseGeneratorPowershell,
+)
--- a/.coveragerc
+++ b/.coveragerc
@@ -14,3 +14,5 @@ source =
 [report]
 # show lines missing coverage in output
 show_missing = True
+exclude_also =
+    raise NotImplementedError
--- a/datalad_next/iterable_subprocess/iterable_subprocess.py
+++ b/datalad_next/iterable_subprocess/iterable_subprocess.py
@@ -8,12 +8,30 @@ from threading import Thread
 from datalad_next.exceptions import CommandError
 
 
+class OutputFrom(Generator):
+    def __init__(self, stdout, stderr_deque, chunk_size=65536):
+        self.stdout = stdout
+        self.stderr_deque = stderr_deque
+        self.chunk_size = chunk_size
+        self.returncode = None
+
+    def send(self, _):
+        chunk = self.stdout.read(self.chunk_size)
+        if not chunk:
+            raise StopIteration
+        return chunk
+
+    def throw(self, typ, value=None, traceback=None):
+        return super().throw(typ, value, traceback)
+
+
 @contextmanager
 def iterable_subprocess(
     program,
     input_chunks,
     chunk_size=65536,
     cwd=None,
+    bufsize=-1,
 ):
     # This context starts a thread that populates the subprocess's standard input. It
     # also starts a threads that reads the process's standard error. Otherwise we risk
@@ -105,20 +123,6 @@ def iterable_subprocess(
                 if e.errno != 22:
                     raise
 
-    class OutputFrom(Generator):
-        def __init__(self, stdout):
-            self.stdout = stdout
-            self.returncode = None
-
-        def send(self, _):
-            chunk = self.stdout.read(chunk_size)
-            if not chunk:
-                raise StopIteration
-            return chunk
-
-        def throw(self, typ, value=None, traceback=None):
-            return super().throw(typ, value, traceback)
-
     def keep_only_most_recent(stderr, stderr_deque):
         total_length = 0
         while True:
@@ -144,12 +148,13 @@ def iterable_subprocess(
     try:
 
         with \
-                Popen(
+                Popen(  # nosec - all arguments are controlled by the caller
                     program,
                     stdin=PIPE,
                     stdout=PIPE,
                     stderr=PIPE,
                     cwd=cwd,
+                    bufsize=bufsize,
                 ) as proc, \
                 thread(
                     keep_only_most_recent,
@@ -164,7 +169,11 @@ def iterable_subprocess(
             try:
                 start_t_stderr()
                 start_t_stdin()
-                chunk_generator = OutputFrom(proc.stdout)
+                chunk_generator = OutputFrom(
+                    proc.stdout,
+                    stderr_deque,
+                    chunk_size
+                )
                 yield chunk_generator
             except BaseException:
                 proc.terminate()
--- a/datalad_next/itertools/align_pattern.py
+++ b/datalad_next/itertools/align_pattern.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import re
 from typing import (
     Generator,
     Iterable,
@@ -74,15 +75,19 @@ def align_pattern(iterable: Iterable[str | bytes | bytearray],
         pattern multiple times.
     """
 
-    def ends_with_pattern_prefix(data: str | bytes | bytearray,
-                                 pattern: str | bytes | bytearray,
-                                 ) -> bool:
-        """ Check whether the chunk ends with a prefix of the pattern """
-        for index in range(len(pattern) - 1, 0, -1):
-            if data[-index:] == pattern[:index]:
-                return True
-        return False
-
+    # Create pattern matcher for all
+    if isinstance(pattern, str):
+        regex: str | bytes | bytearray = '(' + '|'.join(
+            '.' * (len(pattern) - index - 1) + re.escape(pattern[:index]) + '$'
+            for index in range(1, len(pattern))
+        ) + ')'
+    else:
+        regex = b'(' + b'|'.join(
+            b'.' * (len(pattern) - index - 1) + re.escape(pattern[:index]) + b'$'
+            for index in range(1, len(pattern))
+        ) + b')'
+    pattern_matcher = re.compile(regex, re.DOTALL)
+    pattern_sub = len(pattern) - 1
     # Join data chunks until they are sufficiently long to contain the pattern,
     # i.e. have at least size: `len(pattern)`. Continue joining, if the chunk
     # ends with a prefix of the pattern.
@@ -94,7 +99,9 @@ def align_pattern(iterable: Iterable[str | bytes | bytearray],
         else:
             current_chunk += data_chunk
         if len(current_chunk) >= len(pattern) \
-                and not ends_with_pattern_prefix(current_chunk, pattern):
+                and not (
+                    current_chunk[-1] in pattern
+                    and pattern_matcher.match(current_chunk, len(current_chunk) - pattern_sub)):
             yield current_chunk
             current_chunk = None
 
--- a/datalad_next/itertools/tests/test_align_pattern.py
+++ b/datalad_next/itertools/tests/test_align_pattern.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import timeit
+
 import pytest
 
 from ..align_pattern import align_pattern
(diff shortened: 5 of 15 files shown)


christian-monch:
Multiple commands can be executed in the same `shell_connection` context, e.g.

```python
>>> from datalad_next.utils.shell_connection import shell_connection
>>> with shell_connection(['ssh', 'localhost']) as ssh:
...     r = ssh.execute_command(b'ls -l /etc/passwd')
...     print(tuple(r))
...     print(r.returncode)
...     r = ssh.execute_command(b'ls NO_SUCH_FILE')
...     print(tuple(r))
...     print(r.returncode)
... 
(b'-rw-r--r-- 1 root root 2773 14. Nov 10:05 /etc/passwd\n',)
0
()
2
```

The commands that are executed in the shell can read their `stdin` from a local iterator:

```python
>>> with shell_connection(['ssh', 'localhost']) as ssh:
...     r = ssh.execute_command(b'hexdump -n 10 -C', stdin=iter([b'0123456789']))
...     print(tuple(r))
...     r = ssh.execute_command(b'hexdump -n 5 -C', stdin=iter([b'abcde']))
...     print(tuple(r))
... 
(b'00000000  30 31 32 33 34 35 36 37  38 39                    |0123456789|\n0000000a\n',)
(b'00000000  61 62 63 64 65                                    |abcde|\n00000005\n',)
```

A `shell_connection` supports `download` of remote files, `upload` of local files, and `delete` of remote files.

datalad_next/utils/__init__.py
```diff
@@ -28,6 +40,12 @@
 
 from datalad_next.credman import CredentialManager
 from .log import log_progress
+from .shell_connection import (
+    shell_connection,
+    ShellCommandExecutor,
+    ShellCommandResponseGenerator,
```
mih, line 46: Note to self: Do we need these exposed?

datalad_next/utils/__init__.py
```diff
@@ -28,6 +40,12 @@
 
 from datalad_next.credman import CredentialManager
 from .log import log_progress
+from .shell_connection import (
+    shell_connection,
```
mih, line 44: I wonder if it could simply be a `shell`. A meaningful use case is also a persistent shell subprocess running on a local machine without an intermediate network transport. Seems to work fine:

```py
In [3]: with shell_connection(['bash']) as shell:
   ...:     r = shell.execute_command(b'uname')
   ...:     print(list(r))
   ...:     print(r.returncode)
   ...: 
[b'Linux\n']
0
```

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+        finally:
+            # Ensure that ssh it terminated if an exception is raised by code
+            # that uses `ssh_connection`. This is necessary because
+            # `terminate`-call that is invoked by leaving the
+            # `iterable_subprocess`-context will not end the ssh-process. It
+            # will only terminate if stdin is closed, or if it is killed.
+            subprocess_inputs.put(None)
+
+
+class ShellCommandExecutor:
```
mih, line 90: Should get a `__repr__`. Presently, it is not possible to say what shell one is dealing with when debugging:

```
(Pdb) p shell
<datalad_next.utils.shell_connection.ShellCommandExecutor object at 0x7f22825f2290>
```

Might be worth adding the actual "connection" args to the class for that, or support a dedicated label generated from them.

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+                 ) -> None:
+        self.process_inputs = process_inputs
+        self.stdout = stdout
+        self.end_marker, self.command_postfix = self._get_marker()
+
+    def execute_command(self,
+                        command: bytes,
+                        *,
+                        stdin: Iterable[bytes] | None = None,
+                        ) -> ShellCommandResponseGenerator:
```
mih, line 120: I feel like this should be `__call__`. Would turn

```py
>>> with shell_connection(['bash']) as shell:
...     r = shell.execute_command(b'uname')
```

into

```py
>>> with shell_connection(['bash']) as shell:
...     r = shell(b'uname')
```

It would be nice (and convenient) to support `str`-type commands too. Not sure though about the implications (even now). Wouldn't they be interpreted given the target encoding used in the connected shell (on a possibly remote systems).

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+        self.process_inputs.put([command + self.command_postfix])
+        if stdin is not None:
+            self.process_inputs.put(stdin)
+        return ShellCommandResponseGenerator(self.stdout, self.end_marker)
+
+    def upload(self,
+               local_path: Path,
+               remote_path: PurePosixPath,
+               ) -> tuple[int, bytes]:
+        """Upload a local file to a named file in the connected shell
```
mih, line 158: Should say what the requirement for the upload target are: Must be a POSIX system and must have `dd`.

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+            The file that should be uploaded.
+        remote_path : PurePosixPath
+            The name of the file on the connected shell that will contain the
+            uploaded content.
+
+        Returns
+        -------
+        tuple[int, bytes]
+            A tuple containing the return code of the command that was executed
+            to upload the content and the last 64k bytes of stderr output
```
mih, line 172: What is the rationale for "64k"? It is not set here, it just happens to be 64k because of the setup of the underlying tooling. Would be better to make that connection in the docs too.

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+        local_path : Path
+            The name of the local file that will contain the downloaded content.
+
+        Returns
+        -------
+        tuple[int, bytes]
+            A tuple containing the return code of the command that was executed
+            to upload the content and the last 64k bytes of stderr output
+        """
+        # We use 7z in gzip-mode to compress the data that is sent over the
```
mih, line 203: `gzip` does not seem to be "a" mode, but "the" mode. I did not see a switch to turn this off.

Maybe I am not understanding it properly, but what this compression feature not be redundant with `ssh -C`?

mih (commented):
Thanks! I played with it and it works for me.

Here are some first thoughts.

I think we should carefully decide what features go into the base class implementation. From my POV there is nothing inherently "SSH" about this functionality and also nothing inherently "POSIX".

The present up/download/delete implementation assumes a POSIX environment, but this is certainly not without alternative. This suggests that this feature should be in a separate class.

datalad_next/utils/shell_connection.py
```diff
@@ -0,0 +1,305 @@
...
+                if chunk:
+                    return chunk
+            else:
+                return chunk
+
+        if self.state == 'returncode':
+            # the
+            while b'\n' not in self.returncode_chunk:
+                self.returncode_chunk += next(self.stdout)
+            returncode, trailing = self.returncode_chunk.split(b'\n', 1)
```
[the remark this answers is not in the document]
christian-monch: That's a bug, thanks

datalad_next/utils/__init__.py
christian-monch: I like `shell`, but it is quite generic and I was not sure whether we have another concept that is "closer" to `shell`. I don't mind renaming it.

datalad_next/utils/shell_connection.py
christian-monch: > `gzip` does not seem to be "a" mode, but "the" mode. I did not see a switch to turn this off.

I will update/remove the comment.

> 
> Maybe I am not understanding it properly, but what this compression feature not be redundant with `ssh -C`?
We use the compression feature mainly to ensure that the end-marker is not present in the data stream that is received from the server, for example, if the downloaded file contains an end-marker.

datalad_next/utils/shell_connection.py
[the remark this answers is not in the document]
christian-monch: Fixed in 6919bc099a8c3dde871d8055e5274c41a4643852

christian-monch:
> Thanks! I played with it and it works for me.
> 
> Here are some first thoughts.
> 
> I think we should carefully decide what features go into the base class implementation. From my POV there is nothing inherently "SSH" about this functionality and also nothing inherently "POSIX".
> 

I see it the same way.

> The present up/download/delete implementation assumes a POSIX environment, but this is certainly not without alternative. [...]

I interpreted issue #572 as assuming a POSIX environment on the server.  

> [...] This suggests that this feature should be in a separate class.

Yes, I agree. Independent of the server environment, it would be better to factor out individual operation implementations and allow for flexible "mix and match". Will do that now.

datalad_next/utils/shell_connection.py
christian-monch: Good point, I forgot that `chunk_size` is exposed at `iter_subproc`. Will change it.

christian-monch: done in 1026ae678db87c2dba729ec4598147d90fafdafe

datalad_next/utils/__init__.py
christian-monch: Renamed `datalad_next.utils.shell_connection.shell_connection` to `datalad_next.utils.shell.shell` in b42e435b4e711013bc1ae385e7ad18fdf3593b26

datalad_next/utils/shell_connection.py
christian-monch: Done in a869e0e96b6dec081e5d5018479239dc7fe09f39

christian-monch:
> [...]
> I think we should carefully decide what features go into the base class implementation. From my POV there is nothing inherently "SSH" about this functionality and also nothing inherently "POSIX".
> 
> The present up/download/delete implementation assumes a POSIX environment, but this is certainly not without alternative. This suggests that this feature should be in a separate class.

Just to ensure that I interpret this correctly. We would like to support `shell`-context objects on Posix, OSX, and Windows clients (which is the case because no OS-specific code is used in the implementation)!? 

### Supported client environments:

- [x] Posix
- [x] OSX
- [x] Windows

## Server side

"Server side" refers to the environment in which the connected shell is executed. This environment mainly impacts two things. Firstly, how commands are augmented to allow output and return code detection. Secondly, which commands are sent to perform certain operations. My understanding was that we focus on Posix-server side, i.e. we connect to shells in a Posix environment (in fact the command-augmentation and the initial command that is used to skip login-messages are Posix-specific). So the following is the current state of support:

### Supported server environments:

- [x] Posix
- [ ] OSX
- [ ] Windows


I interpret the comment in such a way that we want to support (or at least create the possibility to support) also  suppor Windows and OSX server environments in the future. Is that correct?

datalad_next/utils/shell_connection.py
christian-monch: Done in 3f4d884d1531140d90f8890697336160a2696aba

datalad_next/utils/shell_connection.py
christian-monch: We now use base64 to encode uploaded and downloaded content. This keeps the end-marker significant and prevents the sending of control-characters

christian-monch:
The `shell_operations`-module supports operations like:

- download
- upload
- delete

For example:
```python
>>> with shell(['bash']) as bash:
...     r, stderr = download(bash, PurePosixPath('/etc/passwd'), Path('/tmp/passwd.download')).
... print(r, stderr)
0 b''
>>> print(Path('/tmp/passwd.download').read_text())
root:x:0:0:root:/root:/bin/bash
systemd-timesync:x:497:497:systemd Time Synchronization:/:/usr/sbin/nologin
messagebus:x:484:484:User for D-Bus:/run/dbus:/usr/sbin/nologin
mysql:x:60:60:MySQL database admin:/var/lib/mysql:/usr/sbin/nologin
polkitd:x:483:483:User for polkitd:/var/lib/polkit:/usr/sbin/nologin
scard:x:482:482:Smart Card Reader:/run/pcscd:/usr/sbin/nologin
    ...
```

datalad_next/utils/shell_connection.py
christian-monch: The requiremets are new POSIX system, `head`, and `base64`. All documented in the doc-string

datalad_next/utils/shell_connection.py
christian-monch: We don't use `7z` anymore. Instead we use `base64` to keep the end-marker significant and to prevent sending of control sequences.

mih (commented):
Thank you for putting this together. Overall this look very clean and well thought-through.

I made a bunch of inline commands, but a have a few key aspects that I want to bring to the table re organization and envisioned usage (patterns).

1. I believe this should not be "hidden" in `utils` -- at least I see no advantages in doing so. Given the plan outlines in https://github.com/datalad/datalad-next/issues/613#issuecomment-1923489248 I'd suggest to add one or two top-level sub-packages.  The actual number depends on the outcome of scope considerations below. Only "public" components should be exposed via the `__init__.py` of these new package(s).

2. The implementation comprises platform-agnostic code and POSIX-only pieces. It is not immediately clear to me what is what. Some class documentations mention windows terms and paths, but also POSIX requirements. This should be made clear.

One way of doing this would be to have a dedicated code component for generic functionality (e.g., `shell`), and additional, swapable units for particular platforms. It appears to me that the client-side implementations are uniformly platform agnostic already. So any swapping would only need to be done, based on user-input or autodetection re target platform capabilities.

It would be absolutely fine for me to only support POSIX targets (for now), but names and organization should make that clear and leave the necessary room and APIs for future alternatives.

3. Separation of shell feature and shell operations. This is already the case. it would make sense to me to make this even stronger by a split into separate top-level sub-packages. A library of operation implementation will potentially grow perpetually. At the same time, a strict separation of the underlying shell feature and its stability will be maximally helpful. It would not be unthinkable to have the `shell` feature eventually live in a core package, whereas the operations library is/remains an add-on.

4. Benchmarks. We still need some information on how fast this code is. Of particular relevance are up/download throughput. The chosen approach should have the capacity to be fast -- but the devil is always in the details. Some comparison to `scp` (or something equally basic) should be given.

5. Internal adoption. I think this new code should replace all non-shell implementations that already exist in datalad-next. This is, of course, `datalad_next/url_operations/ssh.py`, but also the test helper `assert_ssh_access`.

It may be, that for some implementations the use of this feature does not make sense. For those cases, the respective implementation should receive some annotation with the rationale (to avoid such attempts in the future). Seeing such arguments will be rather insightful re scope-definition/suggestion for developers re usage.

mih:
Reporting more observations. I ran this (interactively):

```py
>>> from datalad_next.utils.shell import shell
>>> from datalad_next.utils.shell_operations import download
>>> with shell(['ssh', 'localhost']) as sh:
...     download(sh, Path('~/probe'), Path('/tmp/down'))
```

It returns quickly, but I end up with an empty file (missing 800MB).

datalad_next/utils/shell.py
```diff
@@ -0,0 +1,283 @@
...
+                 stdin: Iterable[bytes] | None = None,
+                 ) -> ShellCommandResponseGenerator:
+        """Execute a command in the connected shell
+
+        Parameters
+        ----------
+        command : bytes
+            The command to execute. Command can be a single POSIX pipeline list.
+            (Internally the command is extended to mark the end of the output
+            by printing an end marker and the return code.)
```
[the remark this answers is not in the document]
christian-monch: Simple commands are supported, pipelines of simple commands are supported, and lists of pipelines that are constructed via `||`, or `&&`, are supported. I will clarify that.

(The terminology is used by bash in its help text)

datalad_next/utils/shell.py
```diff
@@ -0,0 +1,283 @@
...
+            self.process_inputs.put(stdin)
+        return ShellCommandResponseGenerator(self.stdout, self.end_marker)
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self.shell_cmd!r})'
+
+    def close(self):
+        self.process_inputs.put(None)
+
+    def get_command_list(self, command: bytes) -> list[bytes]:
```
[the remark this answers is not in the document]
christian-monch: If we implement support for different platforms, this method would be overwritten in a subclass. Therefore I would leave it as it is for now.

datalad_next/utils/shell.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Returns
+        -------
+        list[bytes]
+            A list of commands that execute ``command`` and send the
+            end-marker and the return code tp ``stdout``.
+        """
+        return [command + self.command_postfix]
+
+    def command_zero(self):
```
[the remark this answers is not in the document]
christian-monch: Same as for `get_command_list`.

datalad_next/utils/shell.py
```diff
@@ -0,0 +1,283 @@
...
+        result_zero = self(b'test')
+        for line in result_zero:
+            lgr.debug('skipped login message line: %s', line)
+        if result_zero.returncode != 1:
+            raise RuntimeError(
+                'unexpected return code from command'
+                f'`test`: {result_zero.returncode}, '
+                f'stderr: {b"".join(result_zero.stderr_deque).decode()}')
+
+    def get_marker(self) -> Any:
```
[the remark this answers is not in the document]
christian-monch: Will be moved to a private method

datalad_next/utils/shell.py
```diff
@@ -0,0 +1,283 @@
...
+        finally:
+            # Ensure that ssh it terminated if an exception is raised by code
+            # that uses `ssh_connection`. This is necessary because
+            # `terminate`-call that is invoked by leaving the
+            # `iterable_subprocess`-context will not end the ssh-process. It
+            # will only terminate if stdin is closed, or if it is killed.
+            subprocess_inputs.put(None)
+
+
+class ShellCommandExecutor:
```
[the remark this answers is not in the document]
christian-monch: I pushed a modified version that brings back the support for multiple shell-flavors, i.e. for POSIX and Microsoft PowerShell. The code existed, I took it out earlier to keep the PR simple and to see what would work best. But I think it makes more sense to bring it back now, so we can discuss the approach here already.

The shell code has now moved to `datalad_next.shell` with the operations in `datalad_next.shell.operations.posix`. As of now there is no `datalad_next.shell.operations.powershell` or similar.


`ssh.py` now contains a class named `ShellCommandExecutorPosix` that houses the POSIX-specific elements of the code. A respective `ShellCommandExecutorPowerShell` exists as well.

christian-monch:
> 
> About 10x slower. I have not checked where the bottleneck could be.

That is very slow. I think we would expect some slowdown, but not that much. I will check this

christian-monch:
> Reporting more observations. I ran this (interactively):
> 
> ```python
> >>> from datalad_next.utils.shell import shell
> >>> from datalad_next.utils.shell_operations import download
> >>> with shell(['ssh', 'localhost']) as sh:
> ...     download(sh, Path('~/probe'), Path('/tmp/down'))
> ```
> 
> It returns quickly, but I end up with an empty file (missing 800MB).
> 
> The return value is:
> 
> ```
> (1,
>  b"Pseudo-terminal will not be allocated because stdin is not a terminal.\r\nbase64: '~/probe': No such file or directory\n")
> ```
> 
> It feels weird that there was no exception. At the same time, an `scp` with that source path, or an absolute path, or a relative path (in this case no directory) work fine.
> 

Thanks for testing that. It turns out that resolving `~`  did not work in `delete`, `download`, or `upload` due to additional (unnecessary) quoting of the argument. Fixed in 675fbefb095b0a5af936a121d19528f12185f8b4

christian-monch:
> [...]
> ### Benchmark
> Using the approach above, but with absolute paths, I ran a first download benchmark. 800MB file from localhost via SSH.
> 
> This implementation
> 
> ```
> 12.1 s ± 187 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> ```
> 
> `scp`
> 
> ```
>             Mean        Std.Dev.    Min         Median      Max
> real        1.329       0.027       1.287       1.333       1.367       
> user        0.919       0.036       0.851       0.925       0.950       
> sys         0.425       0.039       0.382       0.408       0.493       
> ```
> 
> About 10x slower. I have not checked where the bottleneck could be.


I have improved the performance of `align_pattern`, reduced the number of `align_pattern`-calls in `ShellResponseGenerator`, and improved the arguments to `itemize` in `ReadLineFileFromIterator`. The following are the results for downloading a 200 MB file from locahost:

#### scp
```
real        0.993       0.063       0.932       0.969       1.114       
user        0.523       0.024       0.485       0.526       0.560       
sys         0.219       0.030       0.165       0.217       0.264       
```

### posix.download
```
6.14 s ± 85.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
This is still a 6 times slow-down compared `scp`. This slowdown is determined by the following processes:
- about 25% (1.5 x) due to reading base 64 encoded data.
- about 40% (2.5 x) due to line splitting and end-marker detection
- about 35% (2 x) due to base 64 decoding

There is also a general penalty because the base 64 encoding increases the size of transferred by 1/3 (plus some more bytes for newlines).

I will also check the alternative `7z`, encoding and post the results here

If the transfer-rate is high compared to the processing speed in the programs (I had about 440MB/s locally), the performance difference is maximized, on lower transfer-rates the impact decreases. I am looking into the numbers there to see, whether the slowdown is acceptable.

mih:
I am not confident that I understand the approach entirely, when I last parsed the implementation, there was no description. So maybe some of my comments below are not applicable:

AFAICT base64 encoding (or an even more problematic compression) is done solely for avoiding the detection of a false end-marker. These end markers are already random (or have been when I saw the code last), so encountering such invalid marker is unlikely, yet possible.

If the low chance of such an event would be accepted as a shortcoming of the implementation, we should see a substantial performance boost.

I did not find whether a known download/upload size is used as additional evidence. Unless we are looking at (infinite) streams (may just be another intentional limitation), we can know the amount of data to be transferred in advance. There seems to be no need to look for end-markers in such blobs.

It may be possible to implement download/upload end marker handling to merely inspect that last few bytes obtained from the remote process, solely for deciding whether there was an orderly completion (end marker present), or not (but process is dead).

christian-monch:
> I am not confident that I understand the approach entirely, when I last parsed the implementation, there was no description. So maybe some of my comments below are not applicable:
> 
> AFAICT base64 encoding (or an even more problematic compression) is done solely for avoiding the detection of a false end-marker. These end markers are already random (or have been when I saw the code last), so encountering such invalid marker is unlikely, yet possible.
> 
> If the low chance of such an event would be accepted as a shortcoming of the implementation, we should see a substantial performance boost.
> [...]

I removed any encoding in download, this yielded the 3x slower performance (see above). Now I also removed a logging call (commit 4adc4c157fccf073f3c8c5be61667a5abdd3d70f) , which seemed unsuspicious at first, but it made all the difference. Performance is now on par with `scp`. More precisely:

##### scp
```
real        0.930       0.020       0.904       0.929       0.954       
user        0.477       0.007       0.469       0.474       0.488       
sys         0.218       0.014       0.199       0.217       0.240       
```

##### shell download-operation
```
989 ms ± 72.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
This is with pattern detection

> [...]
> I did not find whether a known download/upload size is used as additional evidence. Unless we are looking at (infinite) streams (may just be another intentional limitation), we can know the amount of data to be transferred in advance. There seems to be no need to look for end markers in such blobs.
> 
> It may be possible to implement download/upload end marker handling to merely inspect the last few bytes obtained from the remote process, solely for deciding whether there was an orderly completion (end marker present), or not (but process is dead).

It is a good idea to support the download of known-length objects. That removes the need to search for an end-marker and alleviates the possibility of incorrect end-marker detection. I have an implementation of that in the branch . It is slightly faster than the end marker-based download:

```
972 ms ± 9.42 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
``` 

I think it makes sense to support length-based result interpretation. Will change the code accordingly.

christian-monch:
I have added support for known length operations and a response generator for fixed-length downloads (that allows for downloads with a single `shell.__call__`.

I am currently using this to implement an ssh-backend in `datalad-ria`, but there is a bug in `upload` that escaped my hunt so far. For now, the bug shows up non-deterministically when executing `git annex testremote`, it seems to be concurrency-related and is hard to track down. Searching since last Thursday.

christian-monch:
I finally nailed the bug. It was located in the `upload`-code.

The cause for it is that `BytesIO.read()` will always return **EOF** in an extra call and that the `iterable_subprocess.<locals>.input_to`-thread executes `read()` at "random" points in time. This can lead to a situation where the complete file is uploaded and the shell has sent a complete result, but the final, EOF-triggering, `read()`-call has not yet been done.
We use the availability of the result as an indicator of the completion of the upload operation and leave the context that held the handle of the upload-file, thereby closing the file. When the `input_to`-thread now does the final `read()` on the file handle, it will trigger a `ValueError`. That in turn leads to closing `stdin` of the shell and renders the shell unusable.

The error was difficult to find because it depends on the outcome of a race condition and because `iterable_subprocess.<locals>.input_to` fails silently. The latter should be fixed.

datalad_next/utils/shell.py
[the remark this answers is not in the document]
christian-monch: I renamed it to `get_final_command`, which is a better description of its function. I also changed the signature, it now returns a single byte-string.

mih:
Notes on a conversation with @christian-monch in another channel. This PR is considered done. The issue of client code having to take care of the features of the shell that is being connected to is not to be addressed within this PR.

A review needs to pay attention to properly labeling of implementations (e.g., check that the `posix` operations actually work in any POSIX environment, and not just a subset of shells found in a POSIX environment).

This PR is self-contained, i.e. has the feature implementation and tests for it. However, the feature is added only, and no existing alternative implementations are consolidated on it. A reviewer needs to go through these existing implementations and determine whether they can be replaced with this feature, and if not, why not (candidates to look at: uncurl, (ssh) test helpers (and potentially others), url_operations).

datalad_next/shell/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+)
+from .response_generators import (
+    FixedLengthResponseGenerator,
+    FixedLengthResponseGeneratorPosix,
+    FixedLengthResponseGeneratorPowerShell,
+    ShellCommandResponseGenerator,
+    VariableLengthResponseGenerator,
+    VariableLengthResponseGeneratorPosix,
+    VariableLengthResponseGeneratorPowerShell,
+)
```
mih, line 171: Should there be a common naming/selection strategy?

Client code must choose the current implementation for there use case, but it seems a simple label for a shell-setup is not sufficient. We have a `posix` operations module, but no one for response generators.

Why not have one `posix` module and it has all pieces needed?

Along the same lines, I think we should also avoid that client code has to switch class names per use case. Now thye need to select a module to import from  and what to import also. This means that client code also needs to maintain a mapping of the API. If we would provide modules with the exact same API, client code could be simplified to just get the label of the "platform/shell-type" module to work with.

I also want to point out that the promise is that everything imported from `datalad_next.shell` is stable (or needs a minor/major release to change). We import the entire `posix` module, which not only has the pieces that aim to be a "public" API. I think this calls for a dedicated module that exposes the API.

mih (commented):
I looked into replacing the implementation of `assert_ssh_access` with this functionality. Here is some code that I came up with, based on the existing documentation

```py
    ssh_bash_call = [
        ssh_bin,
        '-i', seckey,
        '-p', port,
        f'{login}@{host}',
        'bash',
    ]
    from more_itertools import consume
    with shell(ssh_bash_call) as ssh:
        consume(ssh(f'mkdir -p {path}'))
        consume(ssh(f'touch {path}/datalad-tests-probe'))
```

Two main issues confused me. The top-level documentation makes the impression that all commands need to be byte-strings (so no f-string formatting is possible). But fortunately this is not true.

It took me a moment to understand the implications of each line of the doc example `shell()`.

`print(b''.join(result))`

Is not just a demo. But the command associated with `result` does not actually run without it.

I think that `ShellCommandExecutor.__call__` should not have this behavior. `myshell(something)` appears to execute a command, but it really gives a handler that needs more boilerplate code to actually do things. At minimum something like

`consume(myshell(something))`

I believe `__call__()` should better do that. Maybe have a `check=` like `subprocess.run()`.

Strangely, when I run a command that fails, I get a `CommandError` immediately. The docs suggest that there should be no exception, but I have to inspect a `returncode` manually.

In some sense I am glad that I do not have to implement a custom error handler for each execution, but I am also not sure if I am doing things correctly.

Overall, I feel we need more examples in the top-level docs. Use cases are similar to the `call_git` helpers:

- run a command that is expected to work, raise exception if not
- check if a command fails or not, return bool
- run a command that is expected to fail, act on returncode
- run a command that produces output, variants
  - expect and produce one line, fail if more
  - expect and produce any number of lines

christian-monch:
> I looked into replacing the implementation of `assert_ssh_access` with this functionality. Here is some code that I came up with, based on the existing documentation
> 
> ```python
>     ssh_bash_call = [
>         ssh_bin,
>         '-i', seckey,
>         '-p', port,
>         f'{login}@{host}',
>         'bash',
>     ]
>     from more_itertools import consume
>     with shell(ssh_bash_call) as ssh:
>         consume(ssh(f'mkdir -p {path}'))
>         consume(ssh(f'touch {path}/datalad-tests-probe'))
> ```
> 
> Two main issues confused me. The top-level documentation makes the impression that all commands need to be byte-strings (so no f-string formatting is possible). But fortunately this is not true.

A thanks for catching that, I modified the code according to a reviewer-comment and forgot to change the docs.

> 
> It took me a moment to understand the implications of each line of the doc example `shell()`.
> 
> `print(b''.join(result))`
> 
> Is not just a demo. But the command associated with `result` does not actually run without it.
> 
> I think that `ShellCommandExecutor.__call__` should not have this behavior. `myshell(something)` appears to execute a command, but it really gives a handler that needs more boilerplate code to actually do things. At minimum something like
> 
> `consume(myshell(something))`
> 
> I believe `__call__()` should better do that. Maybe have a `check=` like `subprocess.run()`.

We can do that, it would mean all output is assembled before it is returned to the user. I assume that we don't expect long slow outputs, e.g. `find /`, so that would be OK. What do you think about collecting all output in `__call__()` and adding a second method that would return a generator?

>
> Strangely, when I run a command that fails, I get a `CommandError` immediately. The docs suggest that there should be no exception, but I have to inspect a `returncode` manually.
> 

Thanks, will fix the docs (also an oversight after addressing a reviewer comment).


> In some sense I am glad that I do not have to implement a custom error handler for each execution, but I am also not sure if I am doing things correctly.
> 
> Overall, I feel we need more examples in the top-level docs. Use cases are similar to the `call_git` helpers:
> 
> * run a command that is expected to work, raise exception if not
> * check if a command fails or not, return bool
> * run a command that is expected to fail, act on returncode
> * run a command that produces output, variants
>   
>   * expect and produce one line, fail if more
>   * expect and produce any number of lines

These are very useful suggestions, I suggest doing the following:

1. Fix the documentation regarding input types and exceptions
2. modify `__call__` to:
   2.1. accept a `check`-argument (default: `False`) that determines whether an exception is raised when the remote command fails.
   2.2. collect all output and return a result dict or dataclass that contains stdout and the return code (unless `check` is `True` and the remote command exits with non-zero.
3. add another execution call that returns a generator (useful for timely output processing of long running remote commands)
4. add more examples to top level documentation according to the list above.

mih:
I will merge this now. There are some unresolved things and tests should still run. But there are too many things pending and we will work on them in parallel for better throughput.

PR #610, opened 2024-01-29
Fix `datalad_next.utils.ParamDictator`

### LINKED ISSUE #609: Fix `datalad_next.utils.ParamDictator`
The code of `datalad_next.utils.ParamDictator` is faulty. `ParamDictator` tries to invoke `__getattr__` on its superclass. The superclass is `object`, which does not possess a `__getattr__`-method. It only has a `__getattribute__`-method.

This should be fixed.

AFAICS the class is currently neither used (except in an unused import) nor tested, which might explain why the error has not been triggered yet. Unless there is an intention to use it in the future, we might also consider removing it.

christian-monch:
Fixes #609

Adds a regression test for `datalad_next.utils.ParamDictator`.

### DIFF
--- a/datalad_next/utils/__init__.py
+++ b/datalad_next/utils/__init__.py
@@ -216,7 +216,7 @@ class ParamDictator:
 
     def __getattr__(self, attr: str):
         if attr.startswith('__'):
-            return super().__getattr__(attr)
+            return super().__getattribute__(attr)
         return self.__params[attr]
 
     def __setattr__(self, attr: str, value: Any):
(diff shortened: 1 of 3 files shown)


mih (commented):
I see that this class is not used at all. I'd be in favor of removing it. WDYT?

We would apply this fix regardless for now.

christian-monch:
I don't mind removing it. But I would also do that at a later stage and merge this branch containing the fix for now. Especially because it also fixes the "pytest-deprecation" error in `datalad_next.utils.tests.test_deprecated` (the fix was done in this PR because it is required for a green appveyor result).

christian-monch:
To get this quickly done, let's agree to do one of the following:

1. merge the branch in its current state
2. add a commit that removes `ParamDictator` and merge the branch then

Issue #613, opened 2024-02-02
Formalize public vs private API rule


mih:
This has been a frequent discussion topic. We need a simple rule for people to tell if something is meant to be public or part of an internal API.

Right now the contributing guide says:

> Code users should be able to import the most relevant functionality from the sub-package's __init__.py.

I would propose to establish a rule that *anything* that cannot be imported from a subpackages `__init__.py` is not considered part of the public API.

This rule is simple.

The one counter argument that I can think of it: There might be pieces that are more special (within in particular subpackage), for example, heavy-ish runtime deps that should not force unconditional imports. If that is the case, it might be worth to:

a) consider moving that code into its own subpackage (if it already has a different dependency set)
b) consider importing within "runtime" code (ie. within a function body)

while otherwise holding on to the general pattern.

adswa:
This rule is indeed straightforward - it makes complete sense to me.

christian-monch:
I like the rule. Same reasoning as @adswa: "straightforward".

mih:
Thanks for the feedback. I looked into what this would mean (using something like `git grep 'datalad_next\..*\.' datalad_next/ | grep -v __init__.py | grep -v '/patches/'` for gathering initial evidence for subsequent curation).

This reveal a bunch of questions. Should we apply this rule uniformly to all sub-packages (or even sub-directories)?

I'd be in favor of that, for simplicity. I am inclined to make an exception for `datalad_next/patches/` (it is not really a place to import anything from. I would try to not make an exception for `datalad_next/tests/` and also not for `datalad_next/utils/`.

With that in mind, here are to chunks of rule-violations we would have in the current code base:

### Test tooling imports not from `datalad_next.tests` directly
```
# utils
datalad_next/annexbackends/tests/test_base.py:from datalad_next.tests.utils import
datalad_next/annexremotes/tests/test_archivist.py:from datalad_next.tests.utils import
datalad_next/annexremotes/tests/test_uncurl.py:from datalad_next.tests.utils import
datalad_next/commands/tests/test_create_sibling_webdav.py:from datalad_next.tests.utils import
datalad_next/commands/tests/test_credentials.py:from datalad_next.tests.utils import
datalad_next/commands/tests/test_download.py:from datalad_next.tests.utils import
datalad_next/commands/tests/test_status.py:from datalad_next.tests.utils import
datalad_next/commands/tests/test_tree.py:from datalad_next.tests.utils import
[...]

PR #615, opened 2024-02-05
Define public API for `url_operations`

mih:
This implements the plan declared at
https://github.com/datalad/datalad-next/issues/613#issuecomment-1923489248

A dedicated list of components is exposed at the package's top-level, and only that. All other functionality and imports have been move to package-internal files.

All other, now undesired, top-level imports have been annotated and scheduled for removal with the v2.0 release.


mih:
Test failure is unrelated.

PR #622, opened 2024-02-05
Define public API of `datalad_next.archive_operations`

mih:
See https://github.com/datalad/datalad-next/issues/613


mih:
Test failure is unrelated

Issue #624, opened 2024-02-05
Changes of the public API scheduled for v2.0


mih:
- https://github.com/datalad/datalad-next/issues/482
- https://github.com/datalad/datalad-next/issues/611

Issue #616, opened 2024-02-05
Remove undesired top-level imports before/with v2.0


mih:
Here is an example PR https://github.com/datalad/datalad-next/pull/615

Look for markers in top-level `__init__.py`, such as `# TODO REMOVE EVERTHING BELOW FOR V2.0` and act on them.

This will resolve https://github.com/datalad/datalad-next/issues/613

PR #627, opened 2024-02-06
Uplift Python API docs

mih:
This continues the work done for #613. Now all relevant symbols are directly documented, for every top-level package.

In doing so, I added a first set of markers to shape the API streamlining that will come with v2.0.
Ping #624
Ping #616

Some implementations that continued to be located in `__init__.py` files have been move into new, dedicated source files.


mih:
Test failure is unrelated.

PR #628, opened 2024-02-06
Free `tree` command from `subdatasets` dependency

mih:
The previous implementation could now be made simpler, with fewer dependencies, using the new iterators. Here `iter_submodules()`.

This use case also helped to identify suboptimal code organization. See individual commits.


mih:
Test failures are unrelated. QC complains about things not introduced here.

Issue #636, opened 2024-02-14
Improve error handling of `datalad-annex::` git remote helper


mih:
A good demo is this https://github.com/datalad/datalad-osf/issues/194

But the behavior is not necessarily specific to the OSF remote.

The basic problem is that we "clone an empty repository", whenever there is actually nothing to clone. Whether that is due to no repository being present, or insufficient authorization, or any other problem: the clone should fail and nor result in an empty repository.

mslw:
Also seen in: https://github.com/datalad/datalad-next/issues/373

mih:
Exploring the issue, using the `directory` special remote as an example:

Cloning from a path that does not exist, wordy, but fails and tells why

```
❯ datalad clone 'datalad-annex::file:///tmp/invalidpath?type=directory&encryption=none&directory={path}' t1
install(error): /tmp/t1 (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were:
- datalad-annex::file:///tmp/invalidpath?type=directory&encryption=none&directory={path}
  CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false clone --progress 'datalad-annex::file:///tmp/invalidpath?type=directory&encryption=none&directory={path}' /tmp/t1' failed with exitcode 128 [err: 'Cloning into '/tmp/t1'...
fatal: CommandError(CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false annex initremote origin type=directory encryption=none directory=/tmp/invalidpath -c annex.dotfiles=true' failed with exitcode 1 under /tmp/t1/.git/dl-repoannex/origin/repoannex [out: 'initremote origin 
failed'] [err: 'git-annex: Directory does not exist: /tmp/invalidpath
initremote: 1 failed'])
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.']]
```

Cloning from an empty directory, succeeds, but should not.

```
❯ datalad clone 'datalad-annex::file:///tmp/validpath?type=directory&encryption=none&directory={path}' t1
[WARNING] You appear to have cloned an empty repository.                                     
[WARNING] Cloned /tmp/t1 but could not find a branch with commits                            
install(ok): /tmp/t1 (dataset)
```

So we are good already, whenever the special remote fails to work entirely. The issue seems limited to the handling of the remote refs detection.

Issue #373, opened 2023-05-24
Clone from a datalad-annex WebDAV URL pointing to a directory that doesn't exist will create that directory


mslw:
I am starting with a dataset that was published to my Sciebo account (using `create-sibling-webdav`) under `$BASEURL/test-20230524/CamelCaseDirectory"` (where base URL is the webdav url shown in Sciebo UI).

First, I am cloning the dataset to a new location, without any issues:

``` console
$ datalad clone "datalad-annex::?type=webdav&encryption=none&exporttree=no&url=$BASEURL/test-20230524/CamelCaseDirectory" clone_camel_case
[INFO   ] Remote origin uses a protocol not supported by git-annex; setting annex-ignore                                                      
[INFO   ] access to 1 dataset sibling sciebo-storage not auto-enabled, enable with:
    (...)
install(ok): /home/mszczepanik/Documents/test/lettercase/clone_camel_case (dataset)
```

Unexpected behavior occurs when using a bogus url (pointing to a directory that doesn't exist in Sciebo). Clone ends with OK and a warning, "You appear to have cloned an empty repository". The cloned dataset is a git repo with no commits.

``` console
$ datalad clone "datalad-annex::?type=webdav&encryption=none&exporttree=no&url=$BASEURL/test-20230524/nonexistentdirectory" clone_non_existent
[WARNING] You appear to have cloned an empty repository.                                                                                      
[WARNING] Cloned /home/mszczepanik/Documents/test-taylan/lettercase/clone_non_existent but could not find a branch with commits               
install(ok): /home/mszczepanik/Documents/test/lettercase/clone_non_existent (dataset)
```

The URLs are case-sensitive, the same happens when typing the url with (incorrect) lowercase:

``` console
$ datalad clone "datalad-annex::?type=webdav&encryption=none&exporttree=no&url=$BASEURL/test-20230524/camelcasedirectory" clone_wrong_case
[WARNING] You appear to have cloned an empty repository.                                                                                      
[WARNING] Cloned /home/mszczepanik/Documents/test/lettercase/clone_wrong_case but could not find a branch with commits                 
install(ok): /home/mszczepanik/Documents/test-taylan/lettercase/clone_wrong_case (dataset)
```

And in the end my Sciebo folder contains `CamelCaseDirectory` (correctly created with `create-sibling-webdav`), but also `nonexistentdirectory` and `camelcasedirectory` which came into being during clone attempts.

Is this an issue with server behavior or with DataLad behavior?

[...]

PR #638, opened 2024-02-15
`datalad-annex::` Test for a repository deposit before establishing mirrorrepo 

mih:
The `mirrorrepo` property of the datalad-annex Git remote helper implementation creates an empty repository, if the remote has no refs (or does not exist). This is sensible behavior for enabling initial pushes.

However, it leads to successful but empty clones/fetches for any misspecified URL or connectivity failure. This is undesirable.

This change adds an explicit test for an available/accessible repository deposit, whenever a clone/fetch is performed. Wrong URL specifications or connectivity issues will now lead to an explicit error exit. This fact is communicated via a non-verbose message, and the full error is available at verbosity level 2, i.e. with a `git clone -v`.

- Closes: https://github.com/datalad/datalad-next/issues/636

TODO

- [x] Add test
- [x] Look into test failure with create-sibling-webdav

### DIFF
--- a/datalad_next/gitremotes/datalad_annex.py
+++ b/datalad_next/gitremotes/datalad_annex.py
@@ -220,17 +220,13 @@ from datalad_next.runners import (
 )
 from datalad_next.uis import ui_switcher as ui
 from datalad_next.utils import (
-    external_versions,
-    patched_env,
-    rmtree,
-)
-
-from datalad_next.utils import (
     CredentialManager,
+    external_versions,
     get_specialremote_credential_envpatch,
     get_specialremote_credential_properties,
     needs_specialremote_credential_envpatch,
     patched_env,
+    rmtree,
     specialremote_credential_envmap,
     update_specialremote_credential,
 )
--- a/datalad_next/gitremotes/datalad_annex.py
+++ b/datalad_next/gitremotes/datalad_annex.py
@@ -691,6 +691,9 @@ class RepoAnnexGitRemote(object):
                     '\n'
                 )
             elif line == 'connect git-receive-pack\n':
+                # "receive", because we are receiving at the local mirror repo
+                # from a `send-pack` process that is connected to the main local
+                # repo
                 self.log('Connecting git-receive-pack\n')
                 self.send('\n')
                 # we assume the mirror repo is in-sync with the remote at
@@ -755,6 +758,16 @@ class RepoAnnexGitRemote(object):
                 self._store_credential()
                 return
             elif line == 'connect git-upload-pack\n':
+                # "upload", because we are uploading from the local mirror repo
+                # to a `fetch-pack` process that is connected to the main local
+                # repo
+                try:
+                    self.get_remote_refs(raise_on_error=True)
+                except Exception as e:
+                    self.log("fatal: couldn't find remote refs (repository deposit does not exist, or is inaccessible", level=1)
+                    self.log(f"query error: {e!r}", level=2)
+                    return
+
                 self.log('Connecting git-upload-pack\n')
                 self.send('\n')
                 # must not capture -- git is talking to it directly from here.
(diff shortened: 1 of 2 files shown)


mih:
The current implementation is not yet optimal:

Initial push to a now yet existing repo deposit

```
❯ git push 127.0.0.1
[DATALAD-ANNEX] Cannot query remote's refs (repository deposit does not exist, or is inaccessible
Enumerating objects: 9, done.
Counting objects: 100% (9/9), done.
...
```

Initial datalad push

```
❯ datalad push --to 127.0.0.1
copy(ok): testfile.dat (file) [to 127.0.0.1-storage...]                                      
Update availability for '127.0.0.1':  75%|████████▎  | 3.00/4.00 [00:00<00:00, 29.9k Steps/s]CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false fetch 127.0.0.1 git-annex' failed with exitcode 128                                                              
[DATALAD-ANNEX] Cannot query remote's refs (repository deposit does not exist, or is inaccessible
-> exit(128)
```

mslw:
> Verify the fix for https://github.com/datalad/datalad-next/issues/373 (@mslw can you give that a try please?)

As of now, the local side seems OK (DataLad produces an understandable error message):

```
❱ datalad clone "datalad-annex::?type=webdav&encryption=none&exporttree=no&url=https://fz-juelich.sciebo.de/remote.php/dav/files/m.szczepanik%40fz-juelich.de/test-2024-02-15/bogus-directory" clone-bogus-directory
install(error): /tmp/clone-bogus-directory (dataset) [Failed to clone from any candidate source URL. Encountered errors per each url were:
- datalad-annex::?type=webdav&encryption=none&exporttree=no&url=https://fz-juelich.sciebo.de/remote.php/dav/files/m.szczepanik%40fz-juelich.de/test-2024-02-15/bogus-directory
  CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false clone --progress 'datalad-annex::?type=webdav&encryption=none&exporttree=no&url=https://fz-juelich.sciebo.de/remote.php/dav/files/m.szczepanik%40fz-juelich.de/test-2024-02-15/bogus-directory' /tmp/clone-bogus-directory' failed with exitcode 128 [err: 'Cloning into '/tmp/clone-bogus-directory'...
[DATALAD-ANNEX] Cannot query remote's refs (repository deposit does not exist, or is inaccessible']]
```

But on the remote side (in Sciebo), `bogus-directory` (which did not exist) gets created by the clone operation.

So I'd say it is not fixed, or partially fixed.

mih:
Thanks for the feedback @mslw -- with that info, it looks like there is something specific to the remote type. I will detach #373 from the set of ambitions here. Needs to have a dedicated investigation.

mih:
I found one of the issues that are a problem here. In `_sync_remote_annex_branch()` (in the push patch `optimize`), it does error handling on fetch failure:

```py
    except mod_push.CommandError as e:
        # it is OK if the remote doesn't have a git-annex branch yet
        # (e.g. fresh repo)
        # TODO is this possible? we just copied? Maybe check if anything
        # was actually copied?
        if "fatal: couldn't find remote ref git-annex" not in e.stderr.lower():
            raise
        lgr.debug('Remote does not have a git-annex branch: %s', e)
```

The messaging from git that it expects is not happening, because the actual message is coming from `datalad-annex` in this case. This same would be true for other Git remote helpers.

I believe that this check needs to be replaced. Not sure how to handle that yet. It would be useful to stop there in cases where something other than a missing ref causes an error exit.

Maybe the test is shortened to `fatal: couldn't find remote ref`

mslw (approved):
Thank you for investigating this. I tried `clone`, and I like the improved error message and the fact that an empty repo is not created.

The commit message for e415687 says that it would also close #373 but as far as I understand the webdav case still warrants investigation (which I cannot commit to at the moment), as `clone` would still create an empty directory on the webdav end.

PR #644, opened 2024-02-20
Add benchmark and codspeed integration

mih:
This changeset adds two (arbitrary) benchmarks via `pytest-benchmark`. This works fine.

It disables the execution of benchmarks (but not the benchmarked implementations) on Appveyor.

It adds a codspeed GitHub action to run and submit benchmarks to codspeed.io.

The latter implementation follows the docs, but it not working. As one can see in `.github/workflows/codspeed.yml` and comparing to the actions logs:

- the installation (with git-annex) works
- the debug-step runs the very same test configuration as the codspeed runner, and passes
- the codspeed run fails, because it cannot find git-annex installed.

This behavior is independent of specific approach to set up git-annex. The exact same behavior is observed when git-annex is installed as a system package (via NeuroDebian). So the failure is not easily attributable to some kind of environment sanitization.

Any ideas what could happen here?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -303,7 +303,7 @@ for:
       # run tests on installed module, not source tree files
       - |
         cd __testhome__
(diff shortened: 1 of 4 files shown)


mih:
I think we are out of luck here. There has been no response on the upstream issue in a month. Closing.

Issue #454, opened 2023-09-28
Implement "force" for `UrlOperations.delete|upload()`


mih:
This is the companion issue for https://github.com/datalad/datalad-ria/issues/103

We need to figure out if and how to deal with upload/delete requests where the associated user has all necessary permissions to successfully complete the operation -- but it would require additional steps (besides the actual upload), e.g. (temporarily) setting a parent directory.

In general, and depending on the protocol, uploads can already be more complex. For example, in some cases it might be desirable for uploads to happen in an atomic fashion (see https://github.com/datalad/datalad-ria/issues/102), which would typically amount to an upload to a tmp location, plus a final rename/move to the actual upload target.

mih:
We might need to put a user of `UrlOperations` handlers in the position to know what they can actually do.

Either we make all uniform, or we make the behavior configurable, or at least queriable.

Example: I am doing an `upload`

Will it overwrite an existing resource silently, or fail.
If it fails, how?
Is the mitigation to run an explicit `delete` before, or is there a flag that can enable forcing an overwrite?

If the container is not-writable by default (but could be made writable), will it fail?
If it fails, how?
Is the mitigation to run an explicit `<yet-to-be-invented>` before, or is there a flag that can enable forcing an write?

christian-monch:
> We might need to put a user of `UrlOperations` handlers in the position to know what they can actually do.
> 
> Either we make all uniform, or we make the behavior configurable, or at least queriable.
> [...]

I would generally opt for a uniform behavior with errors that allow a user to distinguish between causes, i.e. `ResourceExists`-errors, `InsufficientPermission`-errors, and other errors.

> 
> Example: I am doing an `upload`
> 
> Will it overwrite an existing resource silently, or fail. If it fails, how? Is the mitigation to run an explicit `delete` before, or is there a flag that can enable forcing an overwrite?
>  

[...]

PR #649, opened 2024-03-27
Atomic UrlOperations.upload for ssh- and file-URLs 

christian-monch:
This PR addresses issue [datalad-ria #102](https://github.com/datalad/datalad-ria/issues/102) and touches on issue #454 where atomic upload operations are discussed as well.

The PR also adds atomic upload to files. This can prevent data loss in the case where multiple threads write to the same file concurrently.

All ssh- and file-uploads should be atomic, which means an `atomic`-parameter that controls atomicity on a per-call basis is not necessary.

### DIFF
--- a/datalad_next/url_operations/base.py
+++ b/datalad_next/url_operations/base.py
@@ -179,6 +179,11 @@ class UrlOperations:
                timeout: float | None = None) -> Dict:
         """Upload from a local file or stream to a URL
 
+        Whenever possible, uploads are performed atomically This means that the
+        destination will never see a partially uploaded file. It will either
+        see the previous content (or nothing) or the newly uploaded content.
+        Note: this is not supported by all implementations of URL-operations.
+
         Parameters
         ----------
         from_path: Path or None
--- a/datalad_next/url_operations/file.py
+++ b/datalad_next/url_operations/file.py
@@ -4,9 +4,16 @@
 from __future__ import annotations
 
 import logging
-from pathlib import Path
+import random
 import sys
-from typing import Dict
+import time
+from io import IOBase
+from math import floor
+from pathlib import Path
+from typing import (
+    BinaryIO,
+    Dict,
+)
 from urllib import (
     request,
     parse,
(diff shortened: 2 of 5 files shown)


mih (changes requested):
Thanks for putting this PR together. I hope we can get this done quickly now. Given the age of this PR, I'd like to ask for the [first](https://github.com/datalad/datalad-next/pull/649/commits/35cea1de843db906f08338e426f623d26c2dde9b) and the [last](https://github.com/datalad/datalad-next/pull/649/commits/0cbdbbcdd2d3c48cfb2f04a8d7215bd35f94a9eb) commit of this series to be posted as two individual PRs. They are largely unrelated to the actual topic, valid and complete fixes that should not be hanging here. Thanks!

The main diff of the PR changes the URL operations upload from a straight copy to a copy-to-tmp-and-final-mv implementation. This is good IMHO.

However, the diff introduces two random number reinitializations. We cannot do this, and I do not understand why it would be needed either. Please add rationale on the necessity, and add the protective scaffolding (getstate(), setstate()) to limit the impact of the reinitialization to only this very function and without impairing the ability of any Python script importing DataLad to run reproducibly.

Thanks!

christian-monch:
> Thanks for putting this PR together. I hope we can get this done quickly now. Given the age of this PR, I'd like to ask for the [first](https://github.com/datalad/datalad-next/pull/649/commits/35cea1de843db906f08338e426f623d26c2dde9b) and the [last](https://github.com/datalad/datalad-next/pull/649/commits/0cbdbbcdd2d3c48cfb2f04a8d7215bd35f94a9eb) commit of this series to be posted as two individual PRs. They are largely unrelated to the actual topic, valid and complete fixes that should not be hanging here. Thanks!
> 

Done in PR #736 and #737

> [...] 
> However, the diff introduces two random number reinitializations. We cannot do this, and I do not understand why it would be needed either. Please add rationale on the necessity, and add the protective scaffolding (getstate(), setstate()) to limit the impact of the reinitialization to only this very function and without impairing the ability of any Python script importing DataLad to run reproducibly.
> [...]

I removed the reinitialization of random numbers and added a time-stamp modifier to the nonce-values. That should prevent name collisions, even if the random state is identically on calls to `upload` (which is already unlikely).

mih (approved):
The new implementation adds a longer suffix (38chars) (like `.transfer-8267388836_17682337760925293` that include a random number and a timestamp to make collisions unlikely. This is not cheap on filesystems with path length limits, but also an implementation detail that can be addressed, if and when there is a real need.

mih:
Thanks @christian-monch !

PR #650, opened 2024-04-07
Make `iter_diff()` robust to lstat-only changes

### LINKED ISSUE #639: Spurious modification reports from `next-status`
Here is an example log of a bunch of command I ran within a few seconds:

(`git st` is actually an alias to `datalad next-status -r mono`)

```
❯ git st
 modified: src/linkml/ontology/datasets.yaml (file)
untracked: src/examples/datalad-dataset.jsonld
❯ git diff
❯ git diff
❯ git status
On branch bf-35
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        src/examples/datalad-dataset.jsonld

nothing added to commit but untracked files present (use "git add" to track)
❯ git diff
❯ git st
untracked: src/examples/datalad-dataset.jsonld
```

Something like this has happened more than once. The repository here is small, no annexed files (not even an annex at all). So this is not about smudge filters or anything in that direction. But clearly, the implementation in `next-status` suffers from something that is either fixed by time, or by calling `git diff/status` directly.

mih:
With this change, a dedicate `update-index` refresh run is performed before running `diff-index`. This fixes undesired modification reports that are triggered by mtime-change only.

This comes at a performance cost. For a repo with 36k files (many-ish), this is about 5% of the total runtime:

```
%timeit next_status('.', result_renderer='disabled', untracked='all', recursive='datasets')
229 ms ± 537 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit next_status('.', result_renderer='disabled', untracked='all', recursive='datasets')
240 ms ± 2.59 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Closes #639

### DIFF
--- a/changelog.d/20240407_154322_michael.hanke_bf_639.md
+++ b/changelog.d/20240407_154322_michael.hanke_bf_639.md
@@ -1,0 +1,6 @@
+### 🐛 Bug Fixes
+
+- `next-status` no longer issues undesirable modification reports
+  that are based on mtime changes alone (i.e., no content change).
+  Fixes https://github.com/datalad/datalad-next/issues/639 via
+  https://github.com/datalad/datalad-next/pull/650 (by @mih)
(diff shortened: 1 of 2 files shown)


mih:
Failures are unrelated to the changes.

PR #653, opened 2024-04-17
Fix RIA store operations from non-Linux clients

mih:
Aiming to resolve https://github.com/datalad/datalad/issues/7536


mih:
I will merge this now. The `SSHRemoteIO` related patches will like receive a further consolidation based on the work coming in from https://github.com/datalad/datalad-next/pull/596 and https://github.com/datalad/datalad-next/pull/655

PR #660, opened 2024-04-18
Add progress-callback to `datalad_next.shell.operations.posix.download`

christian-monch:
This PR adds a progress callback to the method `datalad_next.shell.operations.posix.download`. It also extends the tests to verify that progress downloads are called for upload- and download-operations.

The PR also corrects a faulty description in the doc-string of `datalad_next.shell.operations.posix.download`.


mih (approved):
LGTM! Thanks for doing that so swiftly.

I agree with comments made elsewhere that we already have `with_progress`, but it it also makes sense to offer comparable features for upload and download. Glad we have it now.

Issue #662, opened 2024-04-24
Failing powershell tests


mih:
The windows tests have been failing consistently since https://github.com/datalad/datalad-next/pull/596 was merged.

The hypothesis is that it fails due to an incompatibility with powershell beyond 5.x (we test on 7.x).

We either need to take the underlying code out (we don't test it at all), or fix the test.

```
FAILED ..\shell\tests\test_shell.py::test_fixed_length_response_generator_powershell - AssertionError: assert b'\nOops, som...---------\r\n' == b'0123456789'
  
  At index 0 diff: b'\n' != b'0'
  
  Full diff:
  - (b'0123456789')
  + (b'\nOops, something went wrong.  Please report this bug with the details be'
  +  b'low.\nReport on GitHub: https://github.com/lzybkr/PSReadLine/issues/new\r\n'
  +  b'-----------------------------------------------------------------------\n'
  +  b'Last 0 Keys:\n\n\nException:\nSystem.IO.IOException: The handle is inval'
  +  b'id.\r\n\r\n   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeF'
  +  b'ullPath)\r\n   at System.Console.GetBufferInfo(Boolean throwOnNoConsole, B'
  +  b'oolean& succeeded)\r\n   at Microsoft.PowerShell.Internal.VirtualTerminal.'
  +  b'get_BufferWidth()\r\n   at Microsoft.PowerShell.PSConsoleReadLine.Initiali'
  +  b'ze(Runspace runspace, EngineIntrinsics engineIntrinsics)\r\n   at Microsof'
  +  b't.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics '
  +  b'engineIntrinsics)\n------------------------------------------------------'
  +  b'-----------------\r\n0123456789\nOops, something went wrong.  Please report'
  +  b' this bug with the details below.\nReport on GitHub: https://github.com/l'
  +  b'zybkr/PSReadLine/issues/new\r\n-------------------------------------------'
  +  b'----------------------------\nLast 0 Keys:\n\n\nException:\nSystem.IO.IOE'
  +  b'xception: The handle is invalid.\r\n\r\n   at System.IO.__Error.WinIOError(I'
  +  b'nt32 errorCode, String maybeFullPath)\r\n   at System.Console.GetBufferInf'
[...]

PR #663, opened 2024-04-24
Disable (power)`shell` test on windows, known incompatibility

mih:
The issues is tracked at https://github.com/datalad/datalad-next/issues/662 We cannot have a permanently failing test.


mih:
This was a mistake. A fix was already found.

PR #664, opened 2024-04-24
Fix a test that fails due to a PowerShell bug

mih:
This commit fixes the test
`test_fixed_length_response_generator_powershell`, which fails on certain PowerShell-versions. The test is skipped on those versions.

Closes https://github.com/datalad/datalad-next/issues/662


mih:
Solved. Yeah!

PR #669, opened 2024-04-25
Patch RIA and ORA code to work on multiple client platforms

christian-monch:
This PR addresses the issues with Windows-clients that use RIA-stores. It replaces the classes `datalad.distributed.ora_remote.ORARemote` and `datalad.distributed.create_sibling_ria.CreateSiblingRia` with updated versions that should work properly on Linux, OSX, and Windows.

##### What does it do?

The main difference to the original code is that all path-operations, e.g. to determine the location of `ria-layout-version`, are performed on URL-paths. Those are represented by instances of `PurePosixPath`. All subclasses of `BaseIO`, i.e. `LocalIO`, `SSHRemoteIO`, and `HTTPRemoteIO`, are extended to contain the method `url2transport_path`. This method converts an URL-path into the correct path for the respective IO abstraction.

Before methods on a subclass of `BaseIO` that require a path are called, the generic URL-path is converted into the correct path for the IO-class by calling `url2transport_path` on the respective IO-class.

A number of tests had to be adapted because the patched methods now assert the correct type for the abstract paths, i.e. `PurePosixPath`. This assertion made several tests fail because they provided OS-paths.

The PR keeps changes to the necessary minimum. That means the source is mostly identical to the original, except for the changes described above.

##### TODO

- [ ] Improve test coverage

### DIFF
--- a/datalad_next/patches/fix_ria_ora_tests.py
+++ b/datalad_next/patches/fix_ria_ora_tests.py
@@ -1,0 +1,1028 @@
+"""Patch ria-, ora-, ria_utils-, and clone-tests to work with modified ria_utils
+
+The ria-utils-patches use an abstract path representation for RIA-store elements.
+This patch adapts the tests that use `ria_utils.create_store` and
+`ria_utils.create_ds_in_store` to these modifications.
+"""
+from __future__ import annotations
+
+import logging
+import shutil
+import stat
+from pathlib import (
+    Path,
+    PurePosixPath,
+)
+from urllib.request import pathname2url
+
+from datalad.api import (
+    Dataset,
+    clone,
+    create_sibling_ria,
+)
+from datalad.cmd import (
+    WitlessRunner as Runner,
+    NoCapture,
+)
+from datalad.customremotes.ria_utils import (
+    UnknownLayoutVersion,
+    create_ds_in_store,
+    create_store,
+    get_layout_locations,
+)
+from datalad.distributed.ora_remote import (
+    LocalIO,
+    SSHRemoteIO,
+)
+from datalad.distributed.tests.ria_utils import (
+    common_init_opts,
+    get_all_files,
+    populate_dataset,
+)
+from datalad.support.exceptions import (
+    CommandError,
+    IncompleteResultsError,
+)
+from datalad.support.network import get_local_file_url
+from datalad.tests.utils_pytest import (
+    SkipTest,
+    assert_equal,
+    assert_false,
+    assert_in,
+    assert_not_in,
+    assert_raises,
+    assert_repo_status,
+    assert_result_count,
+    assert_status,
+    assert_true,
+    create_tree,
+    has_symlink_capability,
+    known_failure_githubci_win,
+    known_failure_windows,
+    rmtree,
+    serve_path_via_http,
+    skip_if_adjusted_branch,
+    swallow_logs,
+    with_tempfile,
+)
+
+from . import apply_patch
+
+
+def local_path2pure_posix_path(path: Path | str):
+    return PurePosixPath(pathname2url(str(path)))
+
+
+# taken from datalad-core@864dc4ae24c8aac0ec4003604543b86de4735732
+@with_tempfile
+def patched__postclonetest_prepare(lcl, storepath, storepath2, link):
+
+    from datalad.customremotes.ria_utils import (
+        create_ds_in_store,
+        create_store,
+        get_layout_locations,
+    )
+    from datalad.distributed.ora_remote import LocalIO
+
+    create_tree(lcl,
+                tree={
+                        'ds': {
+                            'test.txt': 'some',
+                            'subdir': {
+                                'subds': {'testsub.txt': 'somemore'},
+                                'subgit': {'testgit.txt': 'even more'}
+                            },
+                        },
+                      })
+
+    lcl = Path(lcl)
+    storepath = Path(storepath)
+    storepath2 = Path(storepath2)
+    # PATCH: introduce `ppp_storepath` and `ppp_storepath2` and use them instead
+    # of `storepath` and `storepath2`.
+    ppp_storepath = local_path2pure_posix_path(storepath)
+    ppp_storepath2 = local_path2pure_posix_path(storepath2)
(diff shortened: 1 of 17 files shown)


mih (commented):
Few first, very high-level comments:

The PR adds a number of patches that seem to replace all(?) of the ORARemote implementation -- or it is at least not easy to see which parts are left untouched. No all individual patches are documented properly (top-level docstring explaining the purpose).

This makes it hard to understand why these are individual patches, and not just one -- especially given the strong patch order requirements.

From what I can see, all of the following could be one patch (may still be composed of several files):

- `add_method_url2transport_path`
- `replace_sshremoteio`
- `ria_utils`
- `replace_ora_remote`
- `fix_ria_ora_tests`
- `replace_create_sibling_ria`

Please also at some indication where the, then modified, code was taken from. Further changes might be made in datalad-core, and someone would need to figure out which version to compare to. Grep for `taken from` in `datalad_next/patches`.

datalad_next/patches/fix_ria_ora_tests.py
```diff
@@ -0,0 +1,1010 @@
...
+        assert_equal(len(ds.repo.whereis('one.txt')), len(whereis) + 1)
+        # test creating an archive with filters on files
+        ds.export_archive_ora(archive_dir / 'archive2.7z', annex_wanted='(include=*.txt)')
+        # test with wanted expression of a specific remote
+        ds.repo.set_preferred_content("wanted", "include=subdir/*", remote="store")
+        ds.export_archive_ora(archive_dir / 'archive3.7z', remote="store")
+        # test with the current sha
+        ds.export_archive_ora(
+            archive_dir / 'archive4.7z',
+            froms=ds.repo.get_revisions()[1],
```
mih, line 616: This parameter name has a typo, yet the tests pass. This appears to be an indication that this particular test implementation never runs.

christian-monch: I think the archive is purposefully named `archive4.7z`. The test uses `archive2.7z`, `archive3.7z`, and `archive4.7z`. The original test seems incomplete. It just seems to check whether `export_archive_ora` succeeds and never checks the content of the created archives.

The test was executed and triggered errors before it was patched.

christian-monch:
> Few first, very high-level comments:
> 
> The PR adds a number of patches that seem to replace all(?) of the ORARemote implementation -- or it is at least not easy to see which parts are left untouched. No[t] all individual patches are documented properly (top-level docstring explaining the purpose).
> [...]
> Please also at some indication where the, then modified, code was taken from. Further changes might be made in datalad-core, and someone would need to figure out which version to compare to. Grep for `taken from` in `datalad_next/patches`.
> [...]

Thanks for the feedback. You are right, the source of the code and the amount of modifications are not obvious.  I extended the patch to contain the following:

- "taken from"-indicator that contains the commit from which code parts were taken
- "# PATCH:"-comments that indicate where changes were made.
- file doc-strings that explain the purpose of the patch.

> [...]
> This makes it hard to understand why these are individual patches, and not just one -- especially given the strong patch order requirements.
> 
> From what I can see, all of the following could be one patch (may still be composed of several files):
> 
> * `add_method_url2transport_path`
> * `replace_sshremoteio`
> * `ria_utils`
> * `replace_ora_remote`
> * `fix_ria_ora_tests`
> * `replace_create_sibling_ria`
> [...]

I see your point. I will create a single `py`-file that will be referenced in `enabled.py` and apply all patches.

datalad_next/patches/fix_ria_ora_tests.py
mih: Sorry, I was not precise enough. I was talking about `froms=`. AFAICS there is no such argument.

jsheunis:
I have tested this on my mac locally, with the following setup (miniconda virtual environment):
- `datalad` version 1.0.2
- `datalad-next` version installed from this PR branch: 1.3.0+96.g3737a51
- `git-annex` version: 10.20230227
- Python: 3.12.3
- System: darwin/20.6.0 10.16/x86_64

I ran the associated tests locally with no errors. I created a local datalad dataset with annexed content, and created a ria sibling in a ria store on the local file system as well as another sibling in a store accessible via SSH (juseless). Both commands ran without errors. `datalad push` to the siblings worked correctly. And then cloning the siblings to my local filesystem and getting content succeeded in both cases.

Is there anything else I should try out?

mih:
I extracted the config `None` handling to https://github.com/datalad/datalad-next/pull/682 -- it has been failing the tests for too long, and there is no reason to block this fix.

datalad_next/patches/fix_ria_ora_tests.py
christian-monch: I was surprised as well, but there is a `froms`-argument in `ExportArchiveORA`. It holds the content of the `--from` command line argument. Obviously the name was chosen to prevent using the keyword `from`. It should probably have been called `from_` though.

christian-monch:
> I have tested this on my mac locally, with the following setup (miniconda virtual environment):
> 
> * `datalad` version 1.0.2
> * `datalad-next` version installed from this PR branch: 1.3.0+96.g3737a51
> * `git-annex` version: 10.20230227
> * Python: 3.12.3
> * System: darwin/20.6.0 10.16/x86_64
> 
> I ran the associated tests locally with no errors. I created a local datalad dataset with annexed content, and created a ria sibling in a ria store on the local file system as well as another sibling in a store accessible via SSH (juseless). Both commands ran without errors. `datalad push` to the siblings worked correctly. And then cloning the siblings to my local filesystem and getting content succeeded in both cases.
> 
> Is there anything else I should try out?

Perfect, thanks a lot @jsheunis

datalad_next/patches/fix_ria_ora_tests.py
mih: Ah! :man_facepalming: Thanks for clarifying that!

mih (approved):
Thanks a lot for tackling this issue. The patch is a monster, but there was no other way. I am glad that its development also led to a generally more correct test behavior.

I tuned the docs a bit and added a changelog.

Thanks!

Issue #668, opened 2024-04-25
Docs on getting annex info in Python


mih:
Chat room question

> I was looking for some methos in python api to get annex key (MD5) of a file with datalad.. 

Was not simple to answer:

- ls_file_collection docs do not mention all collections
- call_git runners are not accessible in docs and lack json-based example

PR #671, opened 2024-04-25
Update docs

mih:
Closes https://github.com/datalad/datalad-next/issues/668


mslw (approved):
I think everything is good, thanks!

PR #672, opened 2024-04-25
Update versions of GitHub actions

mslw:
This updates the following GitHub actions:

- `actions/checkout` v4 → v5 and `actions/setup-python` v4 → v5, to address warnings (Node.js 16 actions are deprecated) in GitHub which show up in workflow runs
- `tj-actions/changed-files` v23 → v44, to address other warnings which show up in GitHub workflow runs
- `vsoch/pull-request-action` v1.0.24 → v1.1.1 to address workflow ending with an error

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -14,6 +14,6 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


mih (approved):
Thanks! I am not aware of any reason not to upgrade.

PR #686, opened 2024-05-15
Add shell-examples

### LINKED ISSUE #656: Add examples to `shell` documentation
https://github.com/datalad/datalad-next/pull/596 already has extensive documentation. However, there is a significant gap between this documentation and the ability to get something going.

I believe we need to add concrete examples on:

- run a command with stdin input (key advice is presently "hidden" in the docs of one argument)
- run a command with an expected failure
- perform a down/upload (fixed length)
- use the tool when it is impossible/impractical to go with a standard `with` block

And there need to be guidelines/examples on how to deal with different/unknown shell capabilities/platform detection.

Quite a few examples snippets already exist. It could be sufficient(?), but certainly useful to point them out explicitly in the top-level module docs.

christian-monch:
This PR adds a number of additional shell examples and tips. It also links the examples in the top-level package document.

Fixes #656

### DIFF
--- a/datalad_next/shell/__init__.py
+++ b/datalad_next/shell/__init__.py
@@ -86,9 +86,6 @@ it does not rely on the uniqueness of the end marker. It is most useful for
 operation like ``download``, where the length of the output can be known in
 advance.
 
-See the documentation of :func:`datalad_next.shell.shell` for examples of how to
-use the shell-function and different response generator classes.
-
 As mentioned above, the classes :class:`VariableLengthResponseGenerator` and
 :class:`FixedLengthResponseGenerator` are abstract. The module
 :mod:`datalad_next.shell.response_generators` provides the following concrete
@@ -125,6 +122,12 @@ class identified by ``zero_command_rg_class`` will be used by default to create
 the final command list and to parse the result. Users can override this on a
 per-call basis by providing a different response generator class in the
 ``response_generator``-parameter of :meth:`ShellCommandExecutor.__call__`.
+
+Examples
+--------
+
+See the documentation of :func:`datalad_next.shell.shell` for examples of how to
+use the shell-function and different response generator classes.
 
 .. currentmodule:: datalad_next.shell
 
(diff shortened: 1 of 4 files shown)


datalad_next/shell/__init__.py
```diff
@@ -123,6 +123,12 @@ class identified by ``zero_command_rg_class`` will be used by default to create
 per-call basis by providing a different response generator class in the
 ``response_generator``-parameter of :meth:`ShellCommandExecutor.__call__`.
 
+Examples
+--------
+
+See the documentation of :func:`datalad_next.shell.shell` for examples of how to
+use the shell-function and different response generator classes.
+
```
mih, line 131: Adding a subheading here. Otherwise the API table below makes the impression to be a list of examples.

```suggestion

API overview
------------
```

mih (commented):
Awesome examples!!!! I think they add enormous value! Thanks!

I made a small suggestion re clarity.

I am also leaving an observation here that is largely unrelated to this PR (except for being reflected in the API overview):

The rules for the "public API" are "whatever can be imported from the subpackage `__init__.py`. Here we import `.operations.posix`, ie., the whole module. Taken verbatim that would mean that every detail of this module is part of the stable API. I am not sure how to deal with this case optimally. It makes sense to import this at the top-level. Maybe we should at least add an `__all__` declaration to `operations.posix` and amend the rules to state that `__all__` sets the boundary in care of whole-module imports. WDYT?

christian-monch:
>  [...]
> I am also leaving an observation here that is largely unrelated to this PR (except for being reflected in the API overview):
> 
> The rules for the "public API" are "whatever can be imported from the subpackage `__init__.py`. Here we import `.operations.posix`, ie., the whole module. Taken verbatim that would mean that every detail of this module is part of the stable API. I am not sure how to deal with this case optimally. It makes sense to import this at the top-level. Maybe we should at least add an `__all__` declaration to `operations.posix` and amend the rules to state that `__all__` sets the boundary in care of whole-module imports. WDYT?

That makes sense to me. I added an `__all__` to `shell.operations.posix` and amended the contribution guidelines.

mih (approved):
Thanks!

Issue #661, opened 2024-04-18
Consistent handling of shell command quoting 


mih:
As found in https://github.com/datalad/datalad-next/pull/655, we need to deal with command arg quoting in a more consistent manner. Sometimes things are not quoted (leading to failure with, e.g., filenames with spaces), sometimes they are manually quotes (putting arg in `"{args}"`), sometime `sh_quote()` is used.

We need to find *and document* the standard way(s) and implement that throughout.

mih:
Closed via #690

PR #690, opened 2024-05-17
Consistent shell command quoting

christian-monch:
Fixes issue #661 

This PR automatically quotes all path parameters provided as `Path` or `PurePosixPath`-instances to `shell.posix`-functions via `shlex.quote`.

Byte-parameter like the `command`-parameter of `ShellCommandExecutor.__call__()` are never quoted. The caller has to ensure that file names are properly quoted. The are a number of reasons for this policy:

1. It is not clear which syntax is used by the shell-program, for example, it could be: `bash`, `PowerShell`, `Python`, etc.
2. Even if the shell-program was known, e.g. `bash`, there is no way to generally determine whether a path is contained in the command-argument.

### DIFF
--- a/datalad_next/shell/operations/posix.py
+++ b/datalad_next/shell/operations/posix.py
@@ -6,6 +6,7 @@ from pathlib import (
     PurePosixPath,
 )
 from queue import Queue
+from shlex import quote as posix_quote
 from typing import (
     BinaryIO,
     Callable,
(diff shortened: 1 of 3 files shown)


christian-monch:
> The failure on windows seems to be real.

Yes, I overestimated the capabilities of FAT and NTFS. Everything should be fine now

mih (approved):
Thanks!

PR #695, opened 2024-05-20
Run `update-index` in target path, not PWD

mih:
This addresses a problem that rendered the recent lstat-related fix for `iter_gitdiff()` in 87f20e1076e70128594f68a8c520fd6311b7e19b ineffective when not running in a repository directly (PWD).


mih:
Type annotation test failure is unrelated.

PR #696, opened 2024-05-20
Relax type assertion from PurePosixPath to PosixPath in RIA patches

mih:
The patched code is new patch-code, introduced in
https://github.com/datalad/datalad-next/pull/669

The assertions were added to make sure that no platform paths are passed around. The Debian maintainer observed non-pure paths coming in. This may need an investigation, but this changeset maintains the spirit of the assertions while relaxing them enough to let the tests pass in a Debian build environment.

Ping @christian-monch WDYT?

This patch is part of Debian's 1.4.0-1

### DIFF
--- a/datalad_next/patches/add_method_url2transport_path.py
+++ b/datalad_next/patches/add_method_url2transport_path.py
@@ -46,7 +46,7 @@ def local_io_url2transport_path(
         self,
         url_path: PurePosixPath
 ) -> Path | PurePosixPath:
-    assert url_path.__class__ is PurePosixPath
+    assert isinstance(url_path, PurePosixPath)
     if on_windows:
         return str2windows_path(url_path)
     else:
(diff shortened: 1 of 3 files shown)


mih:
Likely also the cause for the breakage observed in https://github.com/datalad-handbook/book/pull/1223

mih:
@christian-monch is not immediately available for a review at this point. Given the relatively large impact, I am going ahead and merge this change. We may need to revisit the decisions made here at a later point in time.

christian-monch:
Thanks @mih. It's good to have the tests pass. I think we have to revisit the approach though.

I looked into issue #705. While this patch turns the tests green, it masks an error in the datalad-next extension tests. The error was due to the fact that extensions were not loaded early enough (see this [comment](https://github.com/datalad/datalad-next/issues/705#issuecomment-2133187476) for details). I opened PR [datalad #7605](https://github.com/datalad/datalad/pull/7605), that basically sets the environment variable `DATALAD_EXTENSION_TESTS=next` when running the datalad-next extension tests. Locally this turns the datalad-next extension tests in datalad green with datalad-next version 1.4.0

I will look now into the debian case.

Issue #702, opened 2024-05-21
Add note that `rev-parse --path-format=relative` requires Git > 2.30


mih:
Reported in office hours

PR #704, opened 2024-05-21
docs: add note of Git >= v2.31 requirement for next-status

mih:
It relies on `rev-parse --path-format=relative`, which has been added in 2.31.

Thanks to @FeHoff for reporting!

Closes: https://github.com/datalad/datalad-next/issues/702


mslw (approved):
It has indeed been added in Git 2.31

https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/RelNotes/2.31.0.txt (line 38):

>  * "git rev-parse" can be explicitly told to give output as absolute
   or relative path with the `--path-format=(absolute|relative)` option.

mih:
Thanks @mslw for the review!

Issue #598, opened 2024-01-17
Add pathspec support to `next_status()`


mih:
The command implementation is coming via #592. The initial pathspec handling support (mostly needed for translating pathspec into the scope of a submodule) is in #588).

The need to be completed and put together.

mih:
I somewhat completed a first implementation that left me wanting more. Thoughts:

It would be good to have pathspecs be honored consistently. For example, saying `:(exclude)*.txt` should remove any `.txt` file from a report and thereby also remove their impact re determining a container item (directory/submodule) to be modified. Git does that for within-repository queries, but not for submodules. So a submodule is modified when a commit has changed -- regardless of whether that change only involved `.txt` files. Same for untracked files. Any untracked file is sufficient to say `untracked content`, even if it is only a `.txt` file.

Thinking how this can be achieved:

`iter_gitdiff` needs updates

- `eval_submodule_state` should be generalized to an `eval_container_state` --  to be uniformly applicable to containers of type directory (with recursion=no) too.
- when pathspecs are given, the reporting of `GitContainerModificationType.new_commits` should be conditional on a pathspec-parametrized change report, no just the mere existence of commits
- we may want to move the reporting of a container item such that it follows its content items, rather than preceding them. This allows us to observe and summarize the reports, rather than relying on Git to provide a report (possibly based on a duplicate investigation
- pathspecs must be passed on to recursive `iter_gitdiff()` calls.

mih:
### Desired alignment with `git status` vs explicit deviation from its behavior

Here is an example:

I am looking for a status report that is concerned with all changes to files that and with `file_a`. I am using the pathspec `*file_a` (`*` matches a directory separator too).

`git status` will report on all files in the top-level repo, but it will completely ignore any submodules that do not itself match the pathspec directly.

A different implementation could look for `*file_a` in any submodule too where the full resulting content paths match the pathspec. This would be more aligned with the behavior of a `git status` without any pathspec constraints (it does full recursion into submodules until it can classify the nature of the change (commits, modifications, untracked content). If this (implicit) recursion is undesired, `--ignore-submodules` has to be used.

We should think about a coherent behavior concept for `next-status`. ATM we have a options that influence reporting, and we have options that influence change detection, and we have an intersection between these two.

[...]

PR #714, opened 2024-05-23
Support for Git's pathspecs

mih:
Eventually replaces #588

- [x] Move `GitPathSpec` out of `__init__.py` and only import that one symbol
- [ ] Add documentation to `GitPathSpec`
- [x] Fix up type annotation of `GitPathSpec`
- [ ] Test subdir propagation of `:attr` type pathspecs (they are like regular globs, but the prefix needs to be maintained)
- [x] Add `EnsurePathSpec` constraint
- [ ] Make pathspecs an orthogonal constraint to any other behavior switch, as described in https://github.com/datalad/datalad-next/issues/598#issuecomment-2134416441
- Closes: #598 
- Closes: #587

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -53,7 +53,7 @@ build: off
 
 environment:
   # unless indicated otherwise, we test datalad_next
-  DTS: datalad_next
+  DTS: datalad_next.gitpathspec datalad_next.iter_collections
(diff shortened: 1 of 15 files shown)


mih:
Replaced by the more advanced #719. Missing pieces will be recovered and put into a future PR.

PR #717, opened 2024-05-29
Add test for the patched RIA/ORA code

christian-monch:
This PR blacklists all datalad-core tests that would fail due to the patched RIA/ORA code in datalad-next. It adds patched versions of the blacklisted tests to datalad-next.

This should fix issues with circular imports when testing datalad extensions while datalad-next is loaded.


christian-monch:
> This works for me! Thx!
> 
> Please remove/fixup the `TMP` commits prior to a merge.

Will do.

PR #718, opened 2024-06-05
rf(EnsureRemoteName): satisfy type-annotation checker

mih:
There should be no functional change, just `assert` statements in strategic places that help `mypy` infer that things are correct the way they are.


christian-monch (dismissed):
LGTM

PR #728, opened 2024-06-17
rf(`CommandError`): start using datasalad's `CommandError`

mih:
This makes our patch obsolete -- once datalad-core has been updated.

DO NOT MERGE UNTIL DataLad 1.2 (or a comparable matching version is released).


mih:
This PR has been made obsolete by #729 -- which achieves a similar goal with a slightly different approach.

PR #731, opened 2024-06-18
Remove non-asserting test from align_pattern tests

christian-monch:
This PR deletes the test `datalad_next.itertools.tests.test_align_pattern.test_performance`, which did not assert any properties, but just printed some performance measurements.


mih:
This code will be removed here. An analog change has already been applied at https://github.com/datalad/datasalad/commit/5d23a439e39af62a827248474e90c81c8ba1f953

Merging nevertheless. Thanks.

PR #732, opened 2024-06-19
Fix type annotations in `alig_pattern`

christian-monch:
This is a backport from `datasalad`  [PR datalad/datasalad#33](https://github.com/datalad/datasalad/pull/33)


mih:
This code will be removed here. An analog change has been applied at https://github.com/datalad/datasalad/pull/33

Merging nevertheless. Thanks!

PR #733, opened 2024-06-28
add `delete`-method to `UrlOperations`

### LINKED ISSUE #533: Implement `DELETE` for `SSHUrlOperations`
Needed ASAP for INM1

Implement a simple delete, i.e. no mode-changes of containing interfaces and no force.

christian-monch:
Fixes #533 

Remark: this PR sits on top of the commits from PR #649, because #649 uses `shell`-code to implement `SshUrlOperations`. If this PR is approved, it will be rebased and merged after #649 has been merged. The last three commits are specific to this PR (the previous commits implement #649).


christian-monch:
The remaining error in `datalad-core 4` is due to a typo in datalad. I openened this [PR](https://github.com/datalad/datalad/pull/7628) to fix it in datalad.

mih (changes requested):
Thanks for this PR!

A delete operation here is conceptualized as "the resource is gone afterwards" -- and the test implementations confirm that.

This makes sense to me. When we send a DELETE via HTTP to some API it will perform whatever makes sense, without necessarily making a client aware of related complexities. Likewise, here, for a deletion of a directory via SSH, the deletion is automatically performed recursively, without something like a "directory-must-be-empty" safety net. It feels rough, but it makes sense.

I feel like it should be made clearer in the docs that this is what is implemented. Right now it is subject to speculation.

christian-monch:
> Thanks for this PR!
> 
> A delete operation here is conceptualized as "the resource is gone afterwards" -- and the test implementations confirm that.
> 
> This makes sense to me. When we send a DELETE via HTTP to some API it will perform whatever makes sense, without necessarily making a client aware of related complexities. Likewise, here, for a deletion of a directory via SSH, the deletion is automatically performed recursively, without something like a "directory-must-be-empty" safety net. It feels rough, but it makes sense.
> 
> I feel like it should be made clearer in the docs that this is what is implemented. Right now it is subject to speculation.

Thanks for the comment. It got me thinking about the automated recursive delete. I opted for not deleting a non-empty directory in `SshUrlOperations.delete`. The reasons were:

1. consistency with `FileUrlOperations.delete`
2. reduce the chance of unintended data removal.

The doc-string of `SshUrlOperations.delete` has been amended by a description of the behavior of `SshUrlOperations.delete` on non-empty directories.

mih:
> I opted for not deleting a non-empty directory in SshUrlOperations.delete. 

Hmm, ok. I wonder how a client implementation could manage to delete a non-empty directory now?

AFAIK, there is no operation that would allow a client to discover the content of a directory. This seems to rule out a (expensive) client-side recursion. How would it be done?

christian-monch:
> [...]
> Hmm, ok. I wonder how a client implementation could manage to delete a non-empty directory now?
> 
> AFAIK, there is no operation that would allow a client to discover the content of a directory. This seems to rule out a (expensive) client-side recursion. How would it be done?

It could not be done in the context of `SshUrlOperations` (I think the same holds for `FileUrlOperations`). So, removing a non-empty directory would require a different communication channel. I can see that this is easier in the `FileUrlOperations` context. 

I think this is a good argument to keep the removal of non-empty directories in `ssh`. It might even be an argument to modify write permissions to ensure the deletion of a remote directory (if we own it). I brought back the non-empty directory delete and added deletion of write-protected targets.

The only thing that irritates me slightly is the different semantics of `FileUrlOperations.delete` and `SshUrlOperations.delete`.

mih:
I think it makes sense to keep the semantics of all implementations aligned and also to document the intent behind that.

Even if it is possible to work around limitations in a filesystem context, it would still defeat the purpose of the handlers. The basic idea was to be able to code around/with defined operations and achieve some protocol abstraction through that. If individual implementations come with individual semantics, this feature is lost (or never achieved) -- which makes them just arbitrary standalone implementations, with little advantage of adopting them over any other solution.

I'd vote to align the semantics of all implementations, incl. the filesystem one.

christian-monch:
`FileUrlOperations.delete` and `SshUrlOperations.delete` now have matching semantics. Both delete a file or directory (including its content). Both can delete write-protected targets if the effective user can modify the permissions of the targets.

They do not try to set write permissions on the containing directory. That means, they cannot delete targets from write-protected directories.

PR #736, opened 2024-07-01
Fix type annotation of exception arguments

christian-monch:
This PR fixes the type annotations of the arguments to `UrlOperationsRessourceUnknown` and `UrlOperationsRemoteError` in `FileUrlOperations.upload()`. and add A regression test for `UrlOperationsRemoteError`.


christian-monch:
The failing `test_wtf`-test in datalad-core is addressed in [PR datalad #7628](https://github.com/datalad/datalad/pull/7628). 

The type errors will be corrected by PR #649

mih (approved):
Thanks!

PR #737, opened 2024-07-01
Add a `close`-handler to the uncurl-annex remote

christian-monch:
This PR adds a close-handler to the uncurl annex remote. The handler closes all open `url_operations-instances`. This allows, for example, to terminate all `shell`-related threads and to exit the uncurl-process.


mih (approved):
Thanks!

Issue #740, opened 2024-07-10
`decode_bytes()` improper error handling


mih:
I have a line from `git ls-files` that comes out of `itemize()` like so:

```b'120000 39d6af4c8d20bfdd8effecf5babaaacc971873e6 0\tBob Marley & The Wailers/Live 9-23-80/08 War \xaf No More Trouble.shn.mp3'
```

`decode_bytes()` trips on the `\xaf` and yields

```
'120000 39d6af4c8d20bfdd8effecf5babaaacc971873e6 0\tBob Marley & The Wailers/Live 9-23-80/08 War \\xaf'
```

i.e. it swallows the end of the file name.

christian-monch:
The iterator returns the result of `decode_bytes([b'08 War \xaf No More Trouble.shn.mp3'])` in two strings.

```python
>>> tuple(decode_bytes([b'08 War \xaf No More Trouble.shn.mp3']))`
('08 War \\xaf', ' No More Trouble.shn.mp3')
```

This behavior allows for relatively simple error handling. But there is no guarantee that `decode_bytes` yields the same number of chunks as its input iterable. It might split input bytes strings at error-locations and it might join input byte strings, if a multi-byte encoding is spread over multiple chunks.

If one wants to receive individual decoded lines, one can decode first and itemize the decoded stream.

mih:
Thanks! This makes sense and it should be documented, I agree.

I addressed the problem for `iter_gitworktree` in #741 

This revealed another issue #742

mih:
I am moving this issue to `datasalad`. The respective code (still affected by #742) has already been migrated there.

PR #741, opened 2024-07-11
fix(`gitworktree`): decode bytes before itemization

mih:
`decode_bytes()` can yield multiple output chunks for one input chunk (in the case of decoding errors). This implies that it cannot be meaningfully used after `itemize()` without threatening the semantics of the items (i.e., with split-by-line items are no longer unique lines).

For this reason, this commit changes the order of these helpers in the `gitworktree()` implementation.

Refs: https://github.com/datalad/datalad-next/issues/740


christian-monch (approved):
Thx @mih, LGTM

PR #742, opened 2024-07-11
fix(`decode_bytes`): always backslashreplace when asked to

mih:
The previous implementation enabled error handling in decoding only for the segment of a bytestring that an exception was raised for.

However, it may well be that more decoding errors exist in other parts of the bytestring. I have a complicated real-world case where this happens, i.e. raising `UnicodeDecodeError` again, even though `decode_bytes` was called with `backslash_replace=True`.  Unfortunately the data is so large that I did not manage to catch the condition exactly.

It seems to be a needless sophistication to decode some part of the bytestring with error handling, but not another.

There is a good chance that this patch is badly interacting with the logic to obtain the next chunk before attempting a decoding again.

### DIFF
--- a/datalad_next/itertools/decode_bytes.py
+++ b/datalad_next/itertools/decode_bytes.py
@@ -103,11 +103,8 @@ def decode_bytes(
         else:
             return (
                 position + exc.end,
-                joined_data[:position + exc.start].decode(encoding)
-                + joined_data[position + exc.start:position + exc.end].decode(
-                    encoding,
-                    errors='backslashreplace'
-                ),
+                joined_data[:position + exc.end].decode(
+                    encoding, errors='backslashreplace')
             )
 
     joined_data = b''
--- a/datalad_next/itertools/decode_bytes.py
+++ b/datalad_next/itertools/decode_bytes.py
@@ -103,8 +103,11 @@ def decode_bytes(
         else:
             return (
                 position + exc.end,
-                joined_data[:position + exc.end].decode(
-                    encoding, errors='backslashreplace')
+                joined_data[position:position + exc.start].decode(encoding)
+                + joined_data[position + exc.start:position + exc.end].decode(
+                    encoding,
+                    errors='backslashreplace'
+                ),
             )
 
     joined_data = b''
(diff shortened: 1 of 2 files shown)


datalad_next/itertools/decode_bytes.py
```diff
@@ -103,11 +103,8 @@ def handle_decoding_error(position: int,
         else:
             return (
                 position + exc.end,
-                joined_data[:position + exc.start].decode(encoding)
-                + joined_data[position + exc.start:position + exc.end].decode(
-                    encoding,
-                    errors='backslashreplace'
-                ),
+                joined_data[:position + exc.end].decode(
+                    encoding, errors='backslashreplace')
```
christian-monch, line 107: I think this does not work. The code duplicates input data, as in the following example:

```python
>>> tuple(decode_bytes([b'08 War \xaf No \xaf More \xaf Trouble.shn.mp3'])
('08 War \\xaf', '08 War \\xaf No \\xaf', '08 War \\xaf No \\xaf More \\xaf', ' Trouble.shn.mp3')
```

I think the return statement should be (a missing `position` index is added):
```python
            return (
                position + exc.end,
                joined_data[position:position + exc.start].decode(encoding)
                + joined_data[position + exc.start:position + exc.end].decode(
                    encoding,
                    errors='backslashreplace'
                ),
            )
```

christian-monch (changes requested):
Thx @mih for catching that. The fix needs a change IMHO.

christian-monch:
We should also add a test for proper handling of multiple errors in a single input chunk. For example

```python
def test_multiple_errors():
    r = ''.join(decode_bytes([b'08 War \xaf No \xaf More \xaf Trouble.shn.mp3']))
    assert r == '08 War \\xaf No \\xaf More \\xaf Trouble.shn.mp3'
```

mih:
Thanks @christian-monch ! Can you make the changes directly, and push into this PR. You have a better understanding of the logic and a fear that more iterations are needed, if I do it. Thanks!

mih:
An alternative to fixing it here, is to fix it in `datasalad`. This code has already been migrated and is pending a release.

mih (commented):
Thanks! Also merged into datasalad!

PR #748, opened 2024-07-16
Fix `CreateSiblingRia` code in `replace_create_sibling_ria.py`

christian-monch:
This PR fixes a bug where the canonified ria-store URL was assigned to `push_url` instead of the
canonified ria-store **push-URL**.


mih (approved):
Makes sense! Thank you!

PR #751, opened 2024-07-18
Fix tests: bring back datalad core wtf-test

christian-monch:
With datalad release 1.1.1 `test_wtf` will pass and no longer let CI fail.


adswa (approved):

mih (approved):
Thanks!

Issue #723, opened 2024-06-15
Move to `hatch`?


mih:
The integration is coming along nicely for `datasalad`. it would make sense to move here too.

adswa:
This is a nice read in this context: https://effigies.gitlab.io/posts/python-packaging-2023/

mih:
Second hatch project is https://hub.datalad.org/ui/dlcmd

PR #752, opened 2024-07-25
Transition to hatch

adswa:
I toyed with hatch following #723. Its pretty neat, and I wanted to leave a work-in-progress draft PR. 

- [x] remove the versioneer setup - this is hindered by the doc/man-page generation, that relies on it for versioning still
- [x] adjust documentation
- [x] adjust ci to use hatch
- Closes: #756
- Closes: #723 

A bunch of tests now fail. These failures smell like an issue with the test assumptions that are no longer valid. The failures are captured here: https://github.com/datalad/datalad-next/issues/759


mih:
I have pushed a seemingly complete setup now. It would make sense to me, if you rebase and consolidate the commits in the PR and we merge, after addressing the two last TODO items.

Re documentation I think we can take most, if not all, from https://github.com/datalad/datalad-core/pull/10

`datalad-core` also uses hatch in the various CI workflows.

I solved the manpage problem by disabling them. This is not nice, but also a minor issue. I filed https://github.com/datalad/datalad-next/issues/758 with thoughts on how to deal with this better.

mih:
https://github.com/pypa/hatch/issues/1677 is an issue that prevents using hatch test --cover on windows.

PR #753, opened 2024-07-30
docs: remove unused intersphinx extension

adswa:
I saw a deprecation warning in a recent doc build (https://github.com/datalad/datalad-next/actions/runs/10028357984/job/27715183098), but did not find the intersphinx mapping that was complained about anymore, so I assume it was removed. Thus, as far as I can see, this extension isn't used anymore


mih (approved):
Thanks!

PR #755, opened 2024-09-09
`eval_results` decorator replacement

mih:


mih:
From my POV this has reached a sensible milestone. I will now evaluate the new possibilities in the context of https://hub.datalad.org/datalad/dlcmd

Once proven, this can be merged.

mih:
I think this can be closed now. The next iteration along these lines is proposed for inclusion into datalad-core: https://github.com/datalad/datalad-core/pull/25

It is part of a more radical change that replaces all of the legacy (and also next's) approach to command implementations with a single decorator.

Issue #756, opened 2024-09-10
Macos codecov uploader fails on Appveyor


adswa:
Appveyors macos builds recently started failing with

```
/usr/local/Cellar/appveyor-build-agent/7.0.3293/bash-shell.sh: line 51: ./codecov: Bad CPU type in executable
Command exited with code 126
Running "on_finish" scripts
```

The codecov docs point to using the "codecov-cli", released in 2023, instead of the uploader: https://docs.codecov.com/docs/codecov-uploader

adswa:
I read up some more. Here is a [blog post](https://about.codecov.io/blog/january-product-update-updating-the-codecov-ci-uploaders-to-the-codecov-cli/) announcing that the codecov bash uploaders will not be maintained anymore outside of critical security releases. 

Instead, everything migrates to use [codecov-cli](https://docs.codecov.com/docs/codecov-uploader). I tried replacing the uploader with codecov-cli (and, later, stumbled over https://github.com/datalad/datalad/pull/7649, in which @yarikoptic tried to fix the same issue, and found an uploader that supports the CPU architecture (though isn't invoked correctly at the moment in datalad-core - [it lacks the "do-upload" subcommand](https://github.com/datalad/datalad/pull/7649#issuecomment-2341759479))). 

**SADLY**, while this uploader does not have the "Bad CPU Type" error anymore, it also doesn't upload reports anymore (on any OS), ultimately I believe because we don't seem to have a codecov token (the tokenless upload would be for PRs from forks of repositories, but it relies on a token in the origin project nevertheless I think: https://docs.codecov.com/docs/tokenless-bundle-analysis). 
```
curl -Os $CODECOV_BINARY
chmod +x codecov
./codecov do-upload
info - 2024-09-10 14:05:06,639 -- ci service found: appveyor
warning - 2024-09-10 14:05:06,675 -- No config file could be found. Ignoring config.
warning - 2024-09-10 14:05:06,700 -- No swift data found.
warning - 2024-09-10 14:05:06,701 -- No gcov data found.
info - 2024-09-10 14:05:06,701 -- Running coverage combine -a in /Users/appveyor/projects/datalad-next/__testhome__
Combined data file .coverage.worker-1248-001.shared.3571.XlaJFYKx
info - 2024-09-10 14:05:06,811 -- Generating coverage.xml report in /Users/appveyor/projects/datalad-next/__testhome__
info - 2024-09-10 14:05:18,353 -- Wrote XML report to coverage.xml
[...]

PR #757, opened 2024-09-10
ci: switch to codecov cli on appveyor

adswa:
This PR replaces the macos codecov uploader that failed with a CPU mismatch with the codecov-cli, which is supposed to be a replacement. See #756 for the background exploration


mih:
Thanks for starting this effort. I am closing this PR now, because I have extended the approach here to all platforms in #752 and it seems to be working properly.

A token for codecov is now needed unconditionally.

PR #760, opened 2024-09-25
New config manager

mih:
TODO:

Remaining issues for replacing the legacy `ConfigManager` with the actual new `LegacyConfigManager`

FAILED datalad_next/patches/tests/test_cli_configoverrides.py::test_cli_configoverrides - AttributeError: 'mappingproxy' object has no attribute 'update'
FAILED datalad_next/annexremotes/tests/test_archivist.py::test_archivist_retrieval - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmo...

### DIFF
--- a/datalad_next/annexremotes/__init__.py
+++ b/datalad_next/annexremotes/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 # in a single place
 from annexremote import UnsupportedRequest
 from typing import Any
+from os import environ
 
 from datalad.customremotes import (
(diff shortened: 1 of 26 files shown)


mih:
This PR has served its purpose and I will close it now.

A config manager like it was proposed here is now part of datalad-core: https://github.com/datalad/datalad-core/pull/19

If there is a desired to include the configuration manager adapter as a semi-long-lived transition helper, it could be taken from here and rebased on datalad-core. However, it is unclear to me if that is even useful.

In any case, this PR has shown that it is possible.

PR #772, opened 2025-02-27
fix: handle URLs more gracefully

### LINKED ISSUE #770: uncurl special remote fails when a key has a URL from a different special remote that it does not understand
I have a file in my dataset that is coming from the [CDS](https://cds.climate.copernicus.eu/) using my [datalad-cds](https://github.com/matrss/datalad-cds) extension. This means it has a `cds:` type URL associated with it that the special remote of that extension uses.

The same file is available from JUDAC and I would like to add that as an additional source via SSH. So I tried out using uncurl for that:
```
git annex initremote uncurl type=external externaltype=uncurl encryption=none autoenable=true
git annex addurl --file=01/2023010100_sf.grb ssh://judac.fz-juelich.de/p/data1/slmet/met_data/ecmwf/era5/grib/2023/01/2023010100_sf.grb
````

The issue now is that when I `datalad get` the file, this happens:
```
$ pixi run datalad get -s uncurl grib/2023/01/2023010100_sf.grb
get(error): grib/2023/01/2023010100_sf.grb (file) [external special remote error: unsupported URL 'cds:v1-eyJkYXRhc2V0IjoicmVhbmFseXNpcy1lcmE1LWNvbXBsZXRlIiwic3ViLXNlbGVjdGlvbiI6eyJjbGFzcyI6ImVhIiwiZGF0ZSI6IjIwMjMtMDEtMDEiLCJleHB2ZXIiOiIxIiwiZm9ybWF0IjoiZ3JpYiIsImdyaWQiOiIuMy8uMyIsImxldnR5cGUiOiJzZmMiLCJwYXJhbSI6IjMxLjEyOC8zMy4xMjgvMzQuMTI4LzM1LjEyOC8zOS4xMjgvNDAuMTI4LzQxLjEyOC80Mi4xMjgvNTkuMTI4LzY2LjEyOC82Ny4xMjgvMTI5LjEyOC8xMzQuMTI4LzEzOS4xMjgvMTQxLjEyOC8xNTEuMTI4LzE1OS4xMjgvMTY0LjEyOC8xNjUuMTI4LzE2Ni4xMjgvMTY3LjEyOC8xNjguMTI4LzE3MC4xMjgvMTcyLjEyOC8xODMuMTI4LzE4Ni4xMjgvMTg3LjEyOC8xODguMTI4LzE5OC4xMjgvMjMyLjEyOC8yMzUuMTI4LzIzNi4xMjgvMjM4LjEyOCIsInN0cmVhbSI6Im9wZXIiLCJ0aW1lIjoiMDAiLCJ0eXBlIjoiYW4ifX0%3D'
external special remote error: unsupported URL 'cds:v1-eyJkYXRhc2V0IjoicmVhbmFseXNpcy1lcmE1LWNvbXBsZXRlIiwic3ViLXNlbGVjdGlvbiI6eyJjbGFzcyI6ImVhIiwiZGF0ZSI6IjIwMjMtMDEtMDEiLCJleHB2ZXIiOiIxIiwiZm9ybWF0IjoiZ3JpYiIsImdyaWQiOiIuMy8uMyIsImxldnR5cGUiOiJzZmMiLCJwYXJhbSI6IjMxLjEyOC8zMy4xMjgvMzQuMTI4LzM1LjEyOC8zOS4xMjgvNDAuMTI4LzQxLjEyOC80Mi4xMjgvNTkuMTI4LzY2LjEyOC82Ny4xMjgvMTI5LjEyOC8xMzQuMTI4LzEzOS4xMjgvMTQxLjEyOC8xNTEuMTI4LzE1OS4xMjgvMTY0LjEyOC8xNjUuMTI4LzE2Ni4xMjgvMTY3LjEyOC8xNjguMTI4LzE3MC4xMjgvMTcyLjEyOC8xODMuMTI4LzE4Ni4xMjgvMTg3LjEyOC8xODguMTI4LzE5OC4xMjgvMjMyLjEyOC8yMzUuMTI4LzIzNi4xMjgvMjM4LjEyOCIsInN0cmVhbSI6Im9wZXIiLCJ0aW1lIjoiMDAiLCJ0eXBlIjoiYW4ifX0%3D'
external special remote error: unsupported URL 'cds:v1-eyJkYXRhc2V0IjoicmVhbmFseXNpcy1lcmE1LWNvbXBsZXRlIiwic3ViLXNlbGVjdGlvbiI6eyJjbGFzcyI6ImVhIiwiZGF0ZSI6IjIwMjMtMDEtMDEiLCJleHB2ZXIiOiIxIiwiZm9ybWF0IjoiZ3JpYiIsImdyaWQiOiIuMy8uMyIsImxldnR5cGUiOiJzZmMiLCJwYXJhbSI6IjMxLjEyOC8zMy4xMjgvMzQuMTI4LzM1LjEyOC8zOS4xMjgvNDAuMTI4LzQxLjEyOC80Mi4xMjgvNTkuMTI4LzY2LjEyOC82Ny4xMjgvMTI5LjEyOC8xMzQuMTI4LzEzOS4xMjgvMTQxLjEyOC8xNTEuMTI4LzE1OS4xMjgvMTY0LjEyOC8xNjUuMTI4LzE2Ni4xMjgvMTY3LjEyOC8xNjguMTI4LzE3MC4xMjgvMTcyLjEyOC8xODMuMTI4LzE4Ni4xMjgvMTg3LjEyOC8xODguMTI4LzE5OC4xMjgvMjMyLjEyOC8yMzUuMTI4LzIzNi4xMjgvMjM4LjEyOCIsInN0cmVhbSI6Im9wZXIiLCJ0aW1lIjoiMDAiLCJ0eXBlIjoiYW4ifX0%3D']
action summary:
  get (error: 1, notneeded: 1)
```

It seems like the uncurl special remote is not properly skipping URLs it cannot handle or trying the remaining URLs when it encounters one it can't handle.

From skimming the code, it seems like the error originates here: https://github.com/datalad/datalad-next/blob/25ac6ef0fcd7b4b1e479d94fa5a8d8c001fc904c/datalad_next/url_operations/any.py#L137

And the special remote fails to continue with the remaining URLs here: https://github.com/datalad/datalad-next/blob/25ac6ef0fcd7b4b1e479d94fa5a8d8c001fc904c/datalad_next/annexremotes/uncurl.py#L526-L541

I think it should additionally catch this exception and continue on with the next URL, until it either finds a working one or has none left to try.

matrss:
Previously an unknown URL (e.g. from a different special remote) would crash the uncurl remote. Additionally, if a key had more than one URL registered only one was tried in some cases.

Now, if the data is not found at one location or the URLs scheme is unsupported, the next URL is tried instead.

Fixes #770.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -32,6 +32,11 @@
             "orcid": "0000-0001-7628-0801"
         },
         {
+            "affiliation": "Institute of Climate and Energy Systems - Stratosphere (ICE-4), Forschungszentrum Jülich, Jülich, Germany",
+            "name": "Riße, Matthias",
+            "orcid": "0009-0006-6081-9327"
+        },
+        {
             "affiliation": "Institute of Neuroscience and Medicine, Brain & Behaviour (INM-7), Research Centre Jülich, Jülich, Germany",
             "name": "Szczepanik, Michał",
             "orcid": "0000-0002-4028-2087"
--- a/datalad_next/annexremotes/uncurl.py
+++ b/datalad_next/annexremotes/uncurl.py
@@ -528,10 +528,13 @@ class UncurlRemote(SpecialRemote):
                 handler(url)
                 # we succeeded, no need to try again
                 return True
-            except UrlOperationsResourceUnknown:
-                # general system access worked, but at the key location is nothing
-                # to be found
-                return False
+            except (UrlOperationsResourceUnknown, ValueError):
+                # UrlOperationsResourceUnknown: general system access worked, but
+                # at the key location is nothing to be found
+                # ValueError: something else went wrong trying to handle the URL
+                # (e.g. unknown scheme)
+                # -> try the next URL
+                pass
             except UrlOperationsRemoteError as e:
                 # return False only if we could be sure that the remote
                 # system works properly and just the key is not around


mih:
Thanks for the PR!

If I understand correctly, the two changes are:

(1) catch a handler crash with `ValueError`
(2) not return `False` to cycle to a next URL, in case of a crash like (1), but also any "unknown" report.

Can you provide a bit more insight into the situation that triggered the undesirable behavior for you?

What handler tripped? On which kind of URL?

Maybe some configuration is too broad in the defaults.

With this change, the return value semantics of the function are a bit crippled. I want to understand, whether you point to a conceptual issue with the imolemented logic. Thanks!

PR #777, opened 2025-07-02
enable tests on top of @mslw's patch

adswa:
I could not push to #776 - this PR includes its changesets, and enables tests in addition.


mih (approved):
LGTM, thanks!

PR #782, opened 2025-07-04
Homogenize

mih:


datalad_next/runners/iter_subproc.py
```diff
@@ -1,122 +0,0 @@
...
-
-
-def iter_subproc(
-    args: List[str],
-    *,
-    input: Iterable[bytes] | None = None,
-    chunk_size: int = COPY_BUFSIZE,
-    cwd: Path | None = None,
-    bufsize: int = -1,
-):
```
adswa, line 24: There are a few test failures in this PR:

```
E       TypeError: iter_subproc() got an unexpected keyword argument 'input'
```

I tried to dig into them a bit, and found https://github.com/datalad/datasalad/commit/4b448912ded25e91f9a6c763620d921cb894a8c0, where the "input" parameter of the now used iter_subproc() implementation in datasalad was renamed to "inputs". 

I will try to push a change that adjusts to the different parameter name, but I suspect that the problem is that different runners either expect "input" or "inputs"? Let's see...

adswa: sadly, yes, if I adjust the parameter name, while previously failing tests pass, I get tons of 

```
FAILED datalad_next/gitremotes/tests/test_datalad_annex.py::test_submodule_url - datalad.runner.exception.CommandError: CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false push -u dla dl-test-branch' failed with exitcode 128 [err: 'fatal: TypeError(call_git_oneline() got an unexpected keyword argument 'input')
```

So this would mean we either homogenize the parameter name across runners (but thus also across several different datalad extensions), revert the renaming in datasalad, or keep the separate runner implementation.

mih:
Closing. Most of this is part of v1.6.0 now

Issue #785, opened 2025-07-16
Updating a datalad-annex Git sibling does not work in adjusted mode


mslw:
On Windows, running `datalad update --how merge` in a dataset previously cloned from WebDAV reports (ok) but the update is not applied (there are no new commits). I suppose, but can't confirm, that this is related to being on an  adjusted branch.

Steps to reproduce (this assumes that the dataset is pushed from LINUX and cloned / updated from WINDOWS, not sure if the pushing OS matters):

LINUX:

(skipping creation steps)

create webdav sibling (git-only because that is the relevant part) and push:
```
datalad create-sibling-webdav -s sciebo --mode git-only https://fz-juelich.sciebo.de/remote.php/dav/files/m.szczepanik%40fz-juelich.de/2025-07-16/test_win_update
datalad push --to sciebo
```

WINDOWS:

clone the dataset:

```
datalad clone "datalad-annex::?type=webdav&encryption=none&exporttree=no&url=https%3A//fz-juelich.sciebo.de/remote.php/dav/files/m.szczepanik%2540fz-juelich.de/2025-07-16/test_win_update"
```

this retrieves the latest state (plus an adjusted branch), which is currently:
```
(test-update) C:\Users\mslw9\Documents\test-update\test_win_update>git log --pretty=oneline
fe3517e2608d6a14c67845a05f4703d227971042 (HEAD -> adjusted/main(unlocked)) git-annex adjusted branch
9da02e8e4ea922e9e301dce7af2fff943fd7e64e (origin/main, origin/HEAD, main) Add a readme
f71f97772b2380c00f90ca9fd7d1d1b7228ce9ae [DATALAD] new dataset
```

LINUX:

(change content, datalad save, datalad push)

this is the state afterwards (f3f... commit is added):

```
❱ git log --pretty=oneline
f3f19e6fbbdba0af7768090a6430e4a64664d3cd (HEAD -> main, sciebo/main) Add content to README
9da02e8e4ea922e9e301dce7af2fff943fd7e64e Add a readme
f71f97772b2380c00f90ca9fd7d1d1b7228ce9ae [DATALAD] new dataset
```

WINDOWS:

Update, this seemingly succeeds:
```
(test-update) C:\Users\mslw9\Documents\test-update\test_win_update>datalad update -s origin --how merge
[INFO] Fetching updates for Dataset("C:\Users\mslw9\Documents\test-update\test_win_update")
[...]

PR #786, opened 2025-07-23
Remove patch to ``update`` that is already included in datalad

adswa:
While looking into #785, I found a patch that I deem superfluous.
The patch to datalad updates '_choose_update_target' has been proposed as a PR in https://github.com/datalad/datalad/pull/7522/ and was merged Oct. 30th 2023.
As the code is part of datalad already, I see no reason that this patch needs to be applied or stay in the code base


mslw (approved):
I think you are right. That PR in datalad was merged and the function in datalad is identical to the one in this patch (except for the comment saying "THIS IS THE PATCH").

adswa:
Thanks for the review! FWIW I can't merge because "Merging is blocked Waiting on code owner review from mih."

yarikoptic:
let's just go and bypass

PR #788, opened 2025-12-07
fix: create WebDAV collection when reconfiguring sibling to new URL

yarikoptic:
When using `existing='reconfigure'` with create_sibling_webdav, the code uses `git annex enableremote` instead of `initremote`. Unlike initremote, enableremote doesn't create the directory structure on the WebDAV server.

This fix adds explicit WebDAV collection creation using MKCOL when reconfiguring to ensure the target directory exists before git-annex tries to use it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

I tasked claude to fixup testing locally, here is the result. Yet to review myself etc

### DIFF
--- a/datalad_next/commands/create_sibling_webdav.py
+++ b/datalad_next/commands/create_sibling_webdav.py
@@ -14,6 +14,8 @@ from urllib.parse import (
     urlunparse,
 )
 
+import requests
+
 from datalad_next.commands import (
     EnsureCommandParameterization,
     ValidatedInterface,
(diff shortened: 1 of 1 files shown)


yarikoptic:
appveyor setup is unhappy as wants to be ahead for minimal python version if I got it right:

```
$TEST_SCRIPT -m "not (turtle)" -k "$KEYWORDS" --pyargs ${DTS}
─────────────────────────────────────────────────────────────────────────────────────────────────────────────── hatch-test.py3.9 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────
Creating environment: hatch-test.py3.9
Installing project in development mode
Checking dependencies
Syncing dependencies
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of coverage[toml] are available:
          coverage[toml]<=7.11.3
          coverage[toml]==7.12.0
      and the current Python version (3.9.21) does not satisfy Python>=3.10, we can conclude that coverage[toml]>=7.11.0 cannot be used.
      And because you require coverage[toml]>=7.11, we can conclude that your requirements are unsatisfiable.
      hint: Pre-releases are available for `coverage` in the requested range (e.g., 7.12.1b1), but pre-releases weren't enabled (try: `--prerelease=allow`)
Command exited with code 1
Running "on_finish" scripts
while [ -f ~/BLOCK ]; do sleep 5; done
Build failed
```

could someone look into updating this repo to their desires? otherwise I will just propose a PR to drop 3.9 (done: #789)

PR #792, opened 2026-05-15
fix: use Bearer for token credentials with Basic-only challenge

hrhee:
## Problem

When a server responds to an unauthenticated request with
`WWW-Authenticate: Basic realm=""` and the stored credential is of type
`token` (Bearer), `handle_401` crashes with `IndexError: list index out of range`.

The root cause is in the `if ascheme is None` block:

```python
ascheme = cred.get('http_auth_scheme')   # set, but immediately overwritten
ascheme = [
    c for c in auth_schemes
    if c in DataladAuth._supported_auth_schemes
    and cred.get('type') == DataladAuth._supported_auth_schemes[c]
][0]   # IndexError when list is empty (Basic != token)
```

Two bugs:
1. The `http_auth_scheme` value from the credential is overwritten
   unconditionally by the list comprehension below it.
2. The list comprehension result is indexed with `[0]` without guarding
   against an empty list.

## Error

```
$ datalad --dbg download --credential mytoken https://example.com/file

download(error): file [download failure] [list index out of range]
...
'exception': list index out of range
  [requests_auth.py:handle_401:290],
```

## Fix

- Split into two separate `if ascheme is None` checks so `http_auth_scheme`
  from the credential is honoured before the list comprehension runs.
- Guard the list comprehension with `if matching:` instead of bare `[0]`.
- Add fallback: if the credential type is `token` and no matching scheme
  was found, try Bearer anyway.

## Steps to reproduce

1. Store a credential of type `token` for an HTTP endpoint:
   ```
   datalad credentials set mytoken type=token \
     realm=https://example.com \
     secret=<some-token>
   ```
2. Download from a server that responds to unauthenticated requests with
   `WWW-Authenticate: Basic realm=""` but accepts Bearer tokens:
   ```
   datalad download --credential mytoken https://example.com/file
   ```
3. → `IndexError: list index out of range` in `handle_401`

This combination occurs with dCache WebDAV endpoints using OIDC
authentication: dCache advertises only Basic in its 401 challenge but
requires a Bearer token.

## TODO

- [x] Add a unit test for the case where a `type=token` credential meets a `Basic`-only `WWW-Authenticate` header
    Tested with
    ```
    uv run --with pytest pytest datalad_next/utils/tests/test_requests_auth.py  datalad_next/utils/tests/test_parse_www_authenticate.py
    ```

### DIFF
--- a/datalad_next/utils/requests_auth.py
+++ b/datalad_next/utils/requests_auth.py
@@ -6,10 +6,12 @@ from __future__ import annotations
 import logging
 from typing import Dict
 from urllib.parse import urlparse
+
 import requests
 
 from datalad_next.config import ConfigManager
 from datalad_next.credman import CredentialManager
+
 from .http_helpers import get_auth_realm
 
 lgr = logging.getLogger('datalad.ext.next.utils.requests_auth')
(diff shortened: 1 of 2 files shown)


mih:
Thanks! I cherry-picked these commits into the main branch. I am planning to release a new version soon.

